PR #4, opened 2018-07-09
Code refactor - first pass

sachaaaaa:
This review is a first step towards moving loki-network forward with modern C++.
Although it's a work in progress, this code currently builds and runs (not functionally tested tho) on my Ubuntu vm.

This patch:
- moves the structs from `iwp_link.cpp` into their own header/source files.
- removes most of the unnecessary `extern "C"`
- removes the unnecessary separation between interfaces and implementations (especially the pointers to functions) since I assume there won't be multiple implementations.

TODO:

- change structs to classes (encapsulation: hide private objects/methods)

- remove the remaining pointers to functions and replace with appropriate c++ mechanism, e.g. lambdas if required

- merge the .h into the .hpp's

- personal preference: change clang formatting, especially the "AlwaysBreakAfter*ReturnType"

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -13,6 +13,7 @@
             },
             "includePath": [
                 "${workspaceFolder}/include",
+                "${workspaceFolder}/llarp",
                 "${workspaceFolder}/vendor/cppbackport-master/lib"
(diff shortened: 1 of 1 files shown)


majestrate:
please add include guards

PR #18, opened 2018-10-01
timebase correction

despair86:
- finally making bencode ILP32/LLP64 clean (long is 32 bits even in 64-bit windows!)
- need to arrange more testing on the avx2 stuff (it looks promising)
- set up CI for windows builds
- clang-format


daemon/main.cpp
```diff
@@ -127,5 +155,8 @@ main(int argc, char *argv[])
     llarp_main_free(ctx);
   }
   exit(code);
+#ifdef _WIN32
+  ::WSACleanup();
```
majestrate: called after `exit()` ?

llarp/ev_win32.hpp
```diff
@@ -85,6 +85,81 @@ namespace llarp
...
+      {
+        t->before_write(t);
+      }
+      ev_io::flush_write();
+    }
+
+    int
+    read(void* buf, size_t sz)
+    {
+      ssize_t ret = tuntap_read(tunif, buf, sz);
```
majestrate: qusetion, does `tuntap_read` include a packet info header?

daemon/main.cpp
despair86: fixed

majestrate:
merged :^)

PR #19, opened 2018-10-04
Autoconfig improvements, DNS integration improvements

neuroscr:
auto-detect config improvement
-c to configure as client
llarp::addr improvements
no log level
bind DNSd to 127.0.0.1:DNS_PORT
NAT fixes when no inbound
DNSd mapAddress on all tun interfaces
Make DNS use unique_ptrs in vectors
Improve internal DNS data structure to play with tun integration better


llarp/dnsd.cpp
```diff
@@ -15,11 +15,7 @@ raw_sendto_dns_hook_func(void *sock, const struct sockaddr *from,
   int *fd = (int *)sock;
   // how do we get to these??
   socklen_t addrLen = sizeof(struct sockaddr_in);
-#ifdef _WIN32
-  return sendto(*fd, (const char *)buffer, length, 0, from, addrLen);
-#else
   return sendto(*fd, (const char *)buffer, length, 0, from, addrLen);
```
despair86: nice.
this was back when i tried supporting ms c++

include/llarp/net.hpp
```diff
@@ -408,6 +446,12 @@ namespace llarp
       return addr4()->s_addr;
     }
 
+    sockaddr*
+    getSockAddr()
```
majestrate: I currently use an operator overload for accessing `llarp::Addr` as a `sockaddr *`

llarp/config.cpp
```diff
@@ -35,13 +35,300 @@ namespace llarp
...
+  f << std::endl;
+  f << "# uncomment following line to set router nickname to 'lokinet'"
+    << std::endl;
+  f << "#nickname=lokinet" << std::endl;
+  f << std::endl << std::endl;
+
+  f << "[bind]" << std::endl;
+  std::string ifname;
+
+  if(llarp::GetBestNetIF(ifname, AF_INET))
```
majestrate: clients shouldn't even need a `bind` section

llarp/dnsd.cpp
```diff
@@ -432,38 +438,58 @@ raw_handle_recvfrom(int *sockfd, const struct sockaddr *saddr, const void *buf,
...
   dns_udp_tracker.dnsd = dnsd;
 
-  dnsd->tracker = &dns_udp_tracker;  // register global tracker with context
-  // dnsd->logic     = logic;             // set logic worker for timers
-  dnsd->intercept = nullptr;  // set default intercepter
+  dnsd->tracker   = &dns_udp_tracker;  // register global tracker with context
+  dnsd->intercept = nullptr;           // set default intercepter
+
+  // set up fresh socket
+  dnsd->client.udp           = new llarp_udp_io;
```
majestrate: `llarp_udp_io` doesn't need to be a pointer on a client, it can and probably should be a member as not a pointer. you can always do `&dnsd->client.udp` instead and it'll work. better that way so you don't have to remember to `delete` it

llarp/ev_kqueue.hpp
```diff
@@ -42,7 +42,15 @@ namespace llarp
       sockaddr* addr = (sockaddr*)&src;
       ssize_t ret    = ::recvfrom(fd, buf, sz, 0, addr, &slen);
       if(ret == -1)
+      {
+        llarp::LogWarn("recvfrom failed");
```
majestrate: this should probably not return -1 here, i remember changing this in epoll version.

llarp/net.cpp
```diff
@@ -887,6 +887,61 @@ namespace llarp
     return found;
   }
 
+  std::string
+  findFreePrivateRange()
+  {
+    // pick ip
+    struct privatesInUse ifsInUse = llarp_getPrivateIfs();
```
majestrate: does `llarp_getPrivateIfs` iterate over all of our network interfaces? i don't understand the `findFreePrivateRange` function.

include/llarp/net.hpp
neuroscr: oh ok cool, I think I had problems in the past, so I didn't bother to try.

llarp/config.cpp
neuroscr: well I left it commented out, so if they chose to change their client into a router, they could without having to look up the reference.

llarp/dnsd.cpp
neuroscr: I understand that but I would still chose a pointer here. The client may needs to reuse the daemon socket (as the previous design needed), or multiple clients may need to run on the same socket. I'm favoring the flexibility here, as I envision it as more of a library than something hard coded to our specific use (trying to keep the lokinet specifics to inside the hooks). And we're talking leaks no more than 10 pointers here that likely only need to be cleaned up on exit but I understand your general principle.

llarp/ev_kqueue.hpp
neuroscr: yea, kqueue may need to be updated, I did see you took a pass at syncing in the last week, not sure if it's still needed.

llarp/net.cpp
neuroscr: Yes, it looks at all interfaces for any private IPs bound to them. findFreePrivateRange finds a range that you're not currently using. I think we may have talked a bit more about this in chat. Let me know if you have any more questions about it.

include/llarp/net.hpp
neuroscr: tried pulling it out, and got:
llarp/dns_dotlokilookup.cpp:113:42: warning: invalid user-defined conversion from ‘llarp::Addr’ to ‘sockaddr*’ [-fpermissive]

PR #21, opened 2018-10-10
tun+ip: lots of NAT and checksum (re)calculation improvements

cathugger:
use incremental checksum calculation where we can.
support UDP checksum recalculation.
also fixes some bugs and cleans up bogus code.

### DIFF
--- a/llarp/ip.cpp
+++ b/llarp/ip.cpp
@@ -60,15 +60,21 @@ namespace llarp
     deltachksum(uint16_t old_sum, uint32_t old_src_ip_n, uint32_t old_dst_ip_n,
                 uint32_t new_src_ip_n, uint32_t new_dst_ip_n)
     {
-#define ADDIPCS(x) ((uint32_t)(x & 0xFFFF) + (uint32_t)(x >> 16))
-#define SUBIPCS(x) ((uint32_t)((~x) & 0xFFFF) + (uint32_t)((~x) >> 16))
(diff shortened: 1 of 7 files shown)


neuroscr:
Wow, good stuff 👍

llarp/ip.cpp
```diff
@@ -16,11 +16,7 @@ namespace llarp
     bool
     IPv4Packet::Load(llarp_buffer_t pkt)
     {
-#ifndef MIN
-#define MIN(a, b) (a < b ? a : b)
-      sz = MIN(pkt.sz, sizeof(buf));
-#undef MIN
-#endif
+      sz = std::min(pkt.sz, sizeof(buf));
```
neuroscr: I wonder if this is going to cause some cross platform problems (I think we tried this before)

despair86: the windows compiler usually requires `#include <algorithm>` *shrugs*]

cathugger: Initial code was stupid because it wouldn't assign anything at to `sz` all if `MIN` was defined.
I'll add <algorithm>, thanks for mentioning that.

majestrate:
good work, merged

llarp/ip.cpp
neuroscr: Ah I see, yea good catch. Just thought it was a simple swap.

PR #26, opened 2018-10-12
some tweaks and optimizations, {h,n}uint16_t types

cathugger:
shouldn't change current logic of anything

### DIFF
--- a/include/llarp/ip.hpp
+++ b/include/llarp/ip.hpp
@@ -155,7 +155,7 @@ namespace llarp
 
       // update ip packet (after packet gets out of network)
       void
-      UpdatePacketOnDst(huint32_t nsrcIP, huint32_t ndstIP);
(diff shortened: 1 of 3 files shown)


majestrate:
This patch doesn't seem to recalculate ICMP checksums.

cathugger:
it never needed to recalculate ICMP checksums, I already had commented that out in my earlier PRs and they work fine.
this patch does remove comment about ICMP, though, maybe I should re-add it as it causes uncertainty for you.

majestrate:
On Fri, Oct 12, 2018 at 11:56:21AM +0000, cathugger wrote:
> it never needed to recalculate ICMP checksums, I already had commented that out in my earlier PRs and they work fine.
> this patch does remove comment about ICMP, though, maybe I should re-add it as it causes uncertainty for you.

I mean if it worked without the recomputation then it's probably fine.

Merging.

PR #29, opened 2018-10-12
fix router configuration bug

despair86:
`this->ip4addr` was no longer written to after the switch to `llarp::Addr`, which left it pointing to zeros or a system-specific test pattern (on Windows: `BA AD F0 0D`) which was later written verbatim to the RC file

### DIFF
--- a/include/llarp/net.hpp
+++ b/include/llarp/net.hpp
@@ -383,7 +383,7 @@ namespace llarp
           out << "]";
       }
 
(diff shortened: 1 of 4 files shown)


cathugger:
fyi new `{h,n}uint16_t` types in https://github.com/loki-project/loki-network/pull/26 can help with keeping track of byte ordering for ports

despair86:
nice, thx

PR #36, opened 2018-10-30
the mother of all pull requests

despair86:
:DDDDD

all of the platform-specific code is now in the platform headers

also includes packaging fluff


neuroscr:
and where was DNSd doing that? I don't see any changes or fixes to DNSd here.

despair86:
i just found out about it last night, before this, the stack would end up completely blasted to bits

test/test_dns_unit.cpp
```diff
@@ -68,7 +68,7 @@ TEST_F(DNSTest, TestDecodeDNSstring)
 TEST_F(DNSTest, TestCodeDomain)
 {
   char buffer[16];
-  llarp::Zero(&buffer, 15);
+  llarp::Zero(&buffer, 16);
```
majestrate: should probably be `buffer` not `&buffer`

despair86: fixed

majestrate:
this ready now?

PR #40, opened 2018-11-05
MX/TXT support, llarp::Addr implementation from header to cpp

neuroscr:
Make DNS string handling more correct. Implement MX/TXT packet types (only returns one of multiple atm). Change most of DNS* sockaddr to *uint32_t structs.

llarp::Addr move from header to cpp to make editing functional compile faster. More net unit tests.

A lot of things I didn't mean to commit (like various guards). Bring fuzz stubs back in, I plan to make them work eventually.


include/llarp/net.hpp
```diff
@@ -134,6 +156,17 @@ namespace llarp
     constexpr bool operator <(huint16_t x) const { return h < x.h; }
     constexpr bool operator ==(huint16_t x) const { return h == x.h; }
 
+    friend std::ostream&
+    operator<<(std::ostream& out, const huint16_t& a)
+    {
+      char tmp[INET_ADDRSTRLEN]   = {0};
+      if(inet_ntop(AF_INET, (void*)&a.h, tmp, INET_ADDRSTRLEN))
```
majestrate: this needs to be converted to network byte order before calling `inet_ntop`

PR #42, opened 2018-11-06
Vendor libuv - latest release from github

michael-loki:
PR for discussion.

This is tied to https://github.com/libuv/libuv/releases/tag/v1.23.2

Add to cmake build, although nothing consumes it. Sorry for your build times!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "vendor/libuv"]
+	path = vendor/libuv
+	url = https://github.com/libuv/libuv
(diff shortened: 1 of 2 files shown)


majestrate:
I have had a change of heart: i think our current event loop is fine given enough unit tests, ripping it out for libuv would be meh because we are already zero allocation event loop, the way libuv does allocations for read is meh and i remembered why i moved off it. 
also, last night in this dream about `uv_buffer_t` memory allocation stealing my face.

majestrate:
On Tue, Nov 06, 2018 at 03:58:56AM -0800, Jeff wrote:
> I have had a change of heart: i think our current event loop is fine given enough unit tests, ripping it out for libuv would be meh because we are already zero allocation event loop, the way libuv does allocations for read is meh and i remembered why i moved off it. 
> also, last night in this dream about `uv_buffer_t` memory allocation stealing my face.

Also sorry for the flip flop on this.

PR #50, opened 2018-11-13
remove windows 2003 stuff

despair86:
rip D:


despair86:
oh heck i accidentally removed the cross-compile rules from cmake

PR #52, opened 2018-11-15
Correct warning in exit.hpp

michael-loki:


majestrate:
that part is due for a revamp soon, it won't be a subtype of that anymore.

PR #58, opened 2018-11-20
Major DNS refactor

neuroscr:
Introduces structure record types, allowing for compound replies
relies on vector<byte_t> more
changes some apis to be less redundant
reduce number of times dns header is parsed
clean up, formatting
includes packet fixes for cname, ptr and mx
fix bug with getDNSstring partial string references
DNSc use actual answer type don't assume question type will be the same
ultimately allows resolution of github.com and actually returns multiple record types now.


include/llarp/dns_rectypes.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <llarp/buffer.h>  // for byte_t
+#include <llarp/net.hpp>   // for llarp::Addr , llarp::huint32_t
+
+namespace llarp
+{
+  namespace dns
+  {
+    struct record
+    {
+      virtual ~record(){};
```
michael-loki: If this class shouldn't be instantiated, tag the destructor as `virtual ~record() = 0;`, and define out-of-line.

include/llarp/dns_rectypes.hpp
```diff
@@ -0,0 +1,114 @@
...
+      record()
+      {
+      }
+
+      virtual bool
+      parse(std::vector< byte_t > bytes);
+
+      virtual std::vector< byte_t >
+      to_bytes();
```
michael-loki: `= 0;` here too.

include/llarp/dns_rectypes.hpp
```diff
@@ -0,0 +1,114 @@
...
+  {
+    struct record
+    {
+      virtual ~record(){};
+      record()
+      {
+      }
+
+      virtual bool
+      parse(std::vector< byte_t > bytes);
```
michael-loki: `= 0;` here too.

include/llarp/dns_rectypes.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    struct type_1a : public record
+    {
+      huint32_t ipaddr;
+
+      virtual ~type_1a(){};
+      type_1a();
+
+      bool
+      parse(std::vector< byte_t > bytes);
```
michael-loki: `override`?

include/llarp/dnsc.hpp
```diff
@@ -32,17 +32,21 @@ typedef void (*dnsc_answer_hook_func)(dnsc_answer_request *request);
...
+  dns_packet packet;
   /// hook
   dnsc_answer_hook_func resolved;
   /// result
   bool found;
-  llarp::huint32_t result;
-  std::string revDNS;
+
+  // llarp::huint32_t result;
+  // std::string revDNS;
```
michael-loki: commented out code?

llarp/dns.cpp
```diff
@@ -124,10 +124,180 @@ code_domain(char *&buffer, const std::string &domain) throw()
...
+
+void
+dns_writeType(std::vector< byte_t > &bytes, llarp::dns::record *record)
+{
+  llarp::dns::type_1a *type1a = dynamic_cast< llarp::dns::type_1a * >(record);
+  if(type1a)
+  {
+    std::vector< byte_t > more_bytes = type1a->to_bytes();
+    llarp::LogDebug("[1]Adding ", more_bytes.size());
+    bytes.insert(bytes.end(), more_bytes.begin(), more_bytes.end());
```
michael-loki: Should a call to `bytes.reserve(bytes.size() + more_bytes.size())` be made before this insert?

llarp/dns.cpp
```diff
@@ -124,10 +124,180 @@ code_domain(char *&buffer, const std::string &domain) throw()
...
+  }
+}
+
+std::vector< byte_t >
+packet2bytes(dns_packet &in)
+{
+  std::vector< byte_t > write_buffer;
+  vput16bits(write_buffer, in.header->id);
+
+  int fields = (in.header->qr << 15);   // QR => message type, 1 = response
```
michael-loki: hoist these magic numbers to constants somewhere?

llarp/dns.cpp
```diff
@@ -323,22 +493,45 @@ extern "C"
                       */
       moveable += answer->rdLen;
       (*pos) += answer->rdLen;  // advance the length
+
+      answer->record.reset(new llarp::dns::type_1a());
```
michael-loki: would this work here?

```c++
answer->record = std::make_unique< llarp::dns::type_1a >();
```

llarp/dns.cpp
```diff
@@ -323,22 +493,45 @@ extern "C"
...
+      answer->record->parse(answer->rData);
     }
     else
     {
       llarp::LogDebug("Got type ", answer->type);
       // FIXME: move this out of here, this shouldn't be responsible for decode
       switch(answer->type)
       {
-        case 2: // NS
+        case 2:  // NS
```
michael-loki: more magic?

llarp/dns_rectypes.cpp
```diff
@@ -0,0 +1,175 @@
...
+    {
+      std::vector< byte_t > retval;
+      vput16bits(retval, 4);               // rdLength
+      vput32bits(retval, this->ipaddr.h);  // write IP
+      return retval;
+    };
+
+    type_2ns::type_2ns() : record()
+    {
+      this->ns = "";
```
michael-loki:
```c++
this->ns.clear();
```

llarp/dns_rectypes.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+    bool
+    type_2ns::parse(std::vector< byte_t > bytes)
+    {
+      // trim last 2 bytes... probably the size
+      this->ns = std::string((char *)bytes.data(), bytes.size() - 2);
```
michael-loki: `static_cast` / `reinterpret_cast` where applicable.

llarp/dnsc.cpp
```diff
@@ -155,21 +157,22 @@ generic_handle_dnsc_recvfrom(dnsc_answer_request *request,
...
   if(!request)
   {
     llarp::LogError(
         "User data to DNS Client response not a dnsc_answer_request");
     // we can't call back the hook
     return;
   }
+  // llarp::LogInfo("got a response, udp user is ", udp->user);
+
+  unsigned char *castBuf     = (unsigned char *)buf;
```
michael-loki: use the right kind of cast here?

llarp/dnsc.cpp
```diff
@@ -224,14 +227,19 @@ generic_handle_dnsc_recvfrom(dnsc_answer_request *request,
   for(uint32_t i = 0; i < hdr->qdCount; i++)
   {
     question = decode_question(castBufc, &pos);
-    //llarp::LogDebug("Read a question, now at ", std::to_string(pos));
+    std::unique_ptr< dns_msg_question > unique(question);
+    request->packet.questions.push_back(std::move(unique));
```
michael-loki: instead just do:

```c++
request->packet.questions.emplace_back(question);
```

llarp/dns.cpp
neuroscr: clang-format

include/llarp/dns_rectypes.hpp
neuroscr: ah yea, tried copying the inheritance pattern from bencode.hpp, message.hpp and findintro.hpp but because of the 3 layers was a bit harder to follow.

Changed the destructor. Not sure what you mean by out-of-line.

include/llarp/dns_rectypes.hpp
neuroscr: what does = 0 mean?

llarp/dns_rectypes.cpp
neuroscr: what's the difference?

llarp/dnsc.cpp
neuroscr: Ah thank you! I knew this could be done better.

llarp/dns_rectypes.cpp
neuroscr: had to use reinterpret_cast, it didn't like static_cast

llarp/dns.cpp
neuroscr: awesome, wanted to remove these news

include/llarp/dnsc.hpp
neuroscr: Never heard a good argument for not including it (beyond the muh OCD), I find older comment snippets invaluable, especially when debugging.

llarp/dns.cpp
neuroscr: I don't see how that would be better, insert() has all the information to do it itself. If I was looping by char, I think it would make sense.

llarp/dns.cpp
neuroscr: Geez uhm, I'm not even sure I understand what these are. Basically just some position adjustment, not really anything labelled.

llarp/dns_rectypes.cpp
majestrate: what you have above is a copy assignment (i think). doing `clear` does it with no allocations (i think)

include/llarp/dns_rectypes.hpp
majestrate: it means it's a pure virtual function with no implementation. meaning the subtype HAS to implement the function in order for it to be constructed.

llarp/dns_rectypes.cpp
neuroscr: it's a cstr, should I even bother initializing it then?

include/llarp/dns.hpp
```diff
@@ -84,12 +86,24 @@ struct dns_packet
...
+packet2bytes(dns_packet &in);
+
 std::string
 getDNSstring(const char *const buffer, uint32_t *pos);
 
 void
 code_domain(char *&buffer, const std::string &domain) throw();
 
+void
+vcode_domain(std::vector< byte_t > &bytes, const std::string &domain) throw();
```
michael-loki: `throw()` specifiers are deprecated in later versions of the standard, since the introduction of `noexcept`.

llarp/dns_rectypes.cpp
michael-loki: yeah, this is the constructor, missed that part. just this is fine:

```c++
type_2ns::type_2ns() : record()
{}
```

include/llarp/dns.hpp
neuroscr: so should I remove throw()? or replace it with noexcept? Again either Jeff code or was just copying patterns from Jeff.

PR #70, opened 2018-11-21
Require C++17 support

lesderid:


majestrate:
debian stable does not have any c++17 capable toolchains as of writing. i run my exit on debian stable.

majestrate:
can you make a new issue for that? thanks.

PR #102, opened 2018-12-04
Minor build fixes

michael-loki:
Add the overrides, add a header.

### DIFF
--- a/llarp/ev.hpp
+++ b/llarp/ev.hpp
@@ -25,6 +25,12 @@ struct asio_evt_pkt
   size_t sz;  // if this doesn't match what is in the packet, note the error
 };
 #else
+
+#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) \
+    || (__APPLE__ && __MACH__)
+#include <sys/event.h>
+#endif
+
 #include <sys/un.h>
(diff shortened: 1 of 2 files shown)


llarp/ev.hpp
```diff
@@ -25,6 +25,7 @@ struct asio_evt_pkt
   size_t sz;  // if this doesn't match what is in the packet, note the error
 };
 #else
+#include <sys/event.h>
```
despair86: ah nice
had that in my freshly-cut dev branch

despair86: oh you forgot the CPP includes
```c
#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) \
    || (__APPLE__ && __MACH__)
#include <sys/event.h>
#endif
```
since `sys/event.h` does not exist on linux (nor do we expect linux users to have `libkqueue` installed, hence the `epoll(2)` implementation of event loop)

michael-loki: done!

majestrate:
On Tue, Dec 04, 2018 at 05:12:22PM -0800, michael-loki wrote:
> michael-loki commented on this pull request.
> 
is this ready for merge now?

PR #103, opened 2018-12-05
merge in dnslib and fix setcap

majestrate:
i need to roll back changes to dnsd and friends tho


majestrate:
changes to dnsd rolled back

PR #116, opened 2018-12-10
Remove -j8 and use -j

michael-loki:
make works this out for you, no need to under-use powerful machine or over commit underpowered machines.


majestrate:
`make -j` runs with unlimited jobs, this is a bad idea.

michael-loki:
👍

majestrate:
On Mon, Dec 10, 2018 at 04:37:12AM -0800, michael-loki wrote:
> 👍 

we can however document how to use ninja.

PR #123, opened 2018-12-10
Convert llarp_router to be a true C++ class

michael-loki:
Also tidy up a few includes around the place. this removes the `router.h` header.


majestrate:
merge now?

PR #152, opened 2018-12-14
Compiler macros to disbale iOS build breaking things

liilac:
**Don't merge me to the master branch yet, as this will break existing macOS support**

Some compiler macro conditionals that don't build iOS-incompatible headers when `Darwin` is defined.  This covers macOS as well, which will be using the same APIs and will not need the excluded headers.

`vendor/libtuntap-master/tuntap-unix-darwin.c` isn't really needed for Darwin support; I am writing Darwin support directly touching the llarp job queue.

### DIFF
--- a/include/llarp.h
+++ b/include/llarp.h
@@ -1,6 +1,8 @@
 #ifndef LLARP_H_
 #define LLARP_H_
 
+#include <stdbool.h>
+
 #ifdef __cplusplus
 #include <ev.h>
 #include <mem.h>
--- a/include/tuntap.h
+++ b/include/tuntap.h
@@ -43,7 +43,12 @@
 #if defined Linux
 #include <netinet/in.h>
 #else
+
+// iOS and other mobile Darwin platforms don't  have this header
+#ifndef Darwin
 #include <netinet/if_ether.h>
+#endif
+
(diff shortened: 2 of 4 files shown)


neuroscr:
https://stackoverflow.com/a/6802945/7697705

has the proper defines to tell the difference between iOS and MacOS

Note: TARGET_OS_MAC is always true on both
so probably want #if (TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR)

despair86:
>iphone sim

or, the platform for which the bizarre `i586-macos-ios` target exists 🤔

despair86:
pr is stale, should we close it?

liilac:
Requesting PR be re-opened; will update with suggestion of using the iOS target detection.

Spent an bit over an hour helping onboard a new dedicated iOS developer, @nielsandriesse , who was experiencing this as a blocking issue, and struggled to work out how to resolve it.

Niels was commenting that as an iOS developer, he doesn't have much familiarity with C/C++ codebases, or more traditional cross-compiling, and found the build infrastructure for this (the C/C++ codebase) to be quite difficult to wrap his head around.

Interest in merging this still?

PR #172, opened 2018-12-20
Convert llarp::AlignedBuffer to be backed by std::array

michael-loki:


llarp/aligned.hpp
```diff
@@ -57,75 +73,72 @@ namespace llarp
...
+      return as_array() >= other.as_array();
     }
 
     AlignedBuffer
     operator^(const AlignedBuffer& other) const
     {
       AlignedBuffer< sz > ret;
-      for(size_t idx = 0; idx < sz / sizeof(Long_t); ++idx)
-        ret.l[idx] = l[idx] ^ other.l[idx];
+      std::transform(as_array().begin(), as_array().end(),
```
majestrate: the point of doing it with longs is so that it's optimized by the compiler better

michael-loki: https://godbolt.org/z/FNhPK7 <- wildly easy to optimise.

michael-loki: doing it with `-mach=native` just uses AVX instructions - https://godbolt.org/z/Uei2HP

PR #191, opened 2019-01-02
let's get this show on the road

despair86:
windows port should be working now
still need to do some testing on my other devices
features such as setting lokinet as default route will be the next pull
but hidden service and service node access is working
```
i'd think the event loop queues are a thing because
of the way unix handles non-blocking i/o
[got to wait for files/sockets to become writable, 
so once you get the data off the network card, 
you still have to wait for the device to become ready, 
so you put the data on the queue until then]

but on windows, async i/o to sockets, files, or devices is easy
no need to wait for anything
```

### DIFF
--- a/llarp/path.cpp
+++ b/llarp/path.cpp
@@ -586,6 +586,10 @@ namespace llarp
     {
       byte_t tmp[MAX_LINK_MSG_SIZE / 2];
       auto buf = llarp::StackBuffer< decltype(tmp) >(tmp);
+      // should help prevent bad paths with uninitialised members
+      // FIXME: Why would we get uninitialised IMessages?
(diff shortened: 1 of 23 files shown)


llarp/path.cpp
```diff
@@ -586,6 +586,10 @@ namespace llarp
     {
       byte_t tmp[MAX_LINK_MSG_SIZE / 2];
       auto buf = llarp::StackBuffer< decltype(tmp) >(tmp);
+      // should help prevent bad paths with uninitialised members
+      // FIXME: Why would we get uninitialised IMessages?
+      if(msg->version != LLARP_PROTO_VERSION)
```
majestrate: why? because we reuse messages between parsing, we don't reset every member but we probably should.

PR #207, opened 2019-01-09
Consolidate multiple CONTRIBUTING files

cheezecat:
1. The .github copy is what currently appears when creating pull requests
2. The .github copy is older and sparser
3. I'm pretty sure the presence of the .github copy is what prevents the root copy from being respected <https://blog.github.com/2016-02-17-issue-and-pull-request-templates/>

All that being said, I don't have an easy way to test that this does what I expect before it is merged.  And is the `lint` pass still requested?

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,4 +1,0 @@
-
-* RUN `make format && make lint -j8` BEFORE COMMITING ALWAYS.
-
-* no tabs
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 * Act like a responsible adult.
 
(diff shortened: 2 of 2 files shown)


majestrate:
On Tue, Jan 08, 2019 at 10:50:35PM -0800, Raven Kopelman wrote:
> 1. The .github copy is what currently appears when creating pull requests
> 2. The .github copy is older and sparser
> 3. I'm pretty sure the presence of the .github copy is what prevents the root copy from being respected <https://blog.github.com/2016-02-17-issue-and-pull-request-templates/>
> 
> All that being said, I don't have an easy way to test that this does what I expect before it is merged.  And is the `lint` pass still requested?

the lint target is way out of date and probably unused at the moment, it can be omitted.

PR #210, opened 2019-01-10
Move lokinet unspecific components to a util/ library

michael-loki:


majestrate:
@neuroscr how hard will this screw over your unmerged fixes?

neuroscr:
actually doesn't look like anything I can't handle. It's the compiler breaker ones I have more of a problem with. This just looks like moving around files/includes.

PR #213, opened 2019-01-11
Reorganise tests to mirror source layout

michael-loki:
Also tidied some dead code up.


neuroscr:
I don't think this should have been merged. Any refactor could introduce new bugs. The name/description seem harmless but review of it's contents, It should have waited until Tuesday meeting acceptance for a merge.

PR #221, opened 2019-01-13
Enable C++14 for all builds

michael-loki:
Bump to C++14. 

Adding to staging to start discussion.


michael-loki:
gcc 6.3.0 supports a subset of c++17 as well 👍 

Will update this PR with a change to `debian.Dockerfile` to use `debian:stable` as verification.

PR #223, opened 2019-01-13
Produce hierarchy of 'libraries' inside libllarp

michael-loki:
This is built on top of #222, but has a .png of the current relationships between directories in `llarp/` (where the `llarp` node means the top level files).

Dependencies were calculated using `grep -hR '#include' "$@" | sed 's/>.*$/>/g' | sort | uniq` for each directory.


neuroscr:
this includes the commits from #222 , maybe separate branches are needed

michael-loki:
thats what happens when you work at weekends :P

PR #230, opened 2019-01-16
UX improvements

neuroscr:
Logging changes design to help during remotely supporting/debugging lokinet problems

- kqueue fix: fix read stomping write data
- bring back ShouldCreateDefaultHiddenService(), allows you to turn off default so you can be pure HS
- Jeff approved workaround to read Keyfile early so Name() can be correct during set up/configuration

- be more verbose about failures
- always show what config file it's using
- declare it's attempting to create config file
- make llarp_ev_add_tun smarter ("" ifnames/ifaddrs are treated as auto)
- make kqueue more verbose about tun failures (esp helpful on macos)
- Handle identity-privkey better
- a fair amount of clang-formatting


michael-loki:
LGTM!

llarp/ev/ev.cpp
```diff
@@ -143,13 +143,54 @@ llarp_ev_udp_sendto(struct llarp_udp_io *udp, const sockaddr *to,
...
+    num = std::stoi(part);
+#endif
+    if(num <= 0)
+    {
+      llarp::LogError("bad ifaddr netmask value: ", ifaddr);
+      return false;
+    }
+    tun->netmask = num;
+    addr         = ifaddr.substr(0, pos);
+    strcpy(tun->ifaddr, addr.c_str());
```
majestrate: `strncpy`

llarp/ev/ev.cpp
```diff
@@ -143,13 +143,54 @@ llarp_ev_udp_sendto(struct llarp_udp_io *udp, const sockaddr *to,
...
+    llarp::LogInfo("IfAddr autodetect: ", tun->ifaddr, "/", tun->netmask);
+  }
+  if(strcmp(tun->ifname, "") == 0 || strcmp(tun->ifname, "auto"))
+  {
+    strcpy(tun->ifname, llarp::findFreeLokiTunIfName().c_str());
```
majestrate:
```
std::string ifname = llarp::findFreeLokiTunIfName();
strncpy(tun->ifname, ifname.c_str(), std::min(sizeof(tun->ifname), ifname.size()));
```

llarp/ev/ev.cpp
michael-loki: I would write this as:

```c++
const std::string addr = ifaddr.substr(0, pos);
std::copy_n(addr.begin(), std::min(sizeof(tun->ifaddr), addr.size(), tun->ifaddr);
```

llarp/ev/ev.cpp
michael-loki: same as above. `const std::string` if you care enough.

llarp/ev/ev_kqueue.hpp
```diff
@@ -286,17 +286,33 @@ namespace llarp
     setup()
     {
       llarp::LogDebug("set up tunif");
-      if(tuntap_start(tunif, TUNTAP_MODE_TUNNEL, TUNTAP_ID_ANY) == -1)
+
+      if(tuntap_start(tunif, TUNTAP_MODE_TUNNEL, 0) == -1)
```
majestrate: why did this change from `TUNTAP_ID_ANY` to `0`?

neuroscr: bad conflict merge? It's nothing I want.

despair86: no idea how that got there honestly

PR #241, opened 2019-01-24
add testnet bootstrap for lokinet on windows

despair86:
static linked, no dependency on mono or .NET. also reverted the "public" bootstrap loader to be more generic

(I suppose it'd be useful for messenger testnet integration?)


contrib/lokinet-bootstrap-winnt/README.md
```diff
@@ -28,8 +28,8 @@ cross-compile build:
 
 # Usage
 
-   C:\>lokinet-bootstrap [uri] [local download path]
+   C:\>lokinet-bootstrap
```
neuroscr: This is the opposite of what we want: a custom URL should be allowed.

majestrate:
ready for merge?

despair86:
👌🏻 ye

PR #249, opened 2019-01-28
enable exit info serialization

majestrate:


neuroscr:
Code looks sparse, is this really everything?

majestrate:
yes, you just need to add the xi to the rc when you're an exit node and the rest is correctness stuff.

PR #258, opened 2019-01-29
Fix MacOS Build, but also disables onion key rotation

neuroscr:


neuroscr:
Michael got it without the disablements

PR #278, opened 2019-02-05
lower utp fragment size by 512 bytes

majestrate:
*  lower utp fragment size by 512 bytes
* backport crash fix from master

### DIFF
--- a/docs/old-wire-protocol.txt
+++ b/docs/old-wire-protocol.txt
@@ -30,13 +30,13 @@ B.rx_K = B_K
 
 the initial value of a_K is HS(A.k) and b_K is HS(B.k)
 
-1120 byte fragments are sent over UTP in an ordered fashion.
+608 byte fragments are sent over UTP in an ordered fashion.
 
 The each fragment F has the following structure:
 
-[ 32 bytes blake2 keyed hash of the following 1088 bytes (h)]
+[ 32 bytes blake2 keyed hash of the following 576 bytes (h)]
 [ 32 bytes random nonce (n)]
-[ 1056 bytes encrypted payload (p)]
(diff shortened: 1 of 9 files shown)


majestrate:
additionally forwardport today's fixes from master

docs/old-wire-protocol.txt
```diff
@@ -30,13 +30,13 @@ B.rx_K = B_K
 
 the initial value of a_K is HS(A.k) and b_K is HS(B.k)
 
-1120 byte fragments are sent over UTP in an ordered fashion.
+608 byte fragments are sent over UTP in an ordered fashion.
```
michael-loki: 👍 for updating the protocol.

llarp/service/endpoint.cpp
```diff
@@ -254,6 +260,17 @@ namespace llarp
...
+      // deregister dead sessions
+      {
+        auto itr = m_DeadSessions.begin();
+        while(itr != m_DeadSessions.end())
+        {
+          if(itr->second->IsDone(now) && itr->second->ShouldRemove())
+            itr = m_DeadSessions.erase(itr);
+          else
+            ++itr;
+        }
```
michael-loki: you could replace this block with:

```c++
std::erase(std::remove_if(m_DeadSessions.begin(), m_DeadSessions.end(), [now](const auto& val) {
  return (val.second->IsDone(now) && val.second->ShouldRemove());
}), m_DeadSessions.end());
```

majestrate: yea i could, should i?

michael-loki: up to you. I prefer the STL version, but I'm a C++ weenie.

PR #283, opened 2019-02-06
more windows fixes

despair86:
more windows patches
- abseil gets built, but not used. removed from win32 build config.
- near (possibly feature-complete) libabyss
- fix that annoying crash/assert on exit
- make sure upoll cleans up properly (as part of the above)
- current unit test suite passes now
- remove text translation (all text files open in raw/binary mode to appease linux)
- move the epoll impl out of header (finally)
- other minor fixes

up next: libcurl integration for lokid automation


llarp/service/Info.hpp
```diff
@@ -5,7 +5,11 @@
 #include <service/types.hpp>
 #include <util/bencode.hpp>
 
+#if __cplusplus >= 201703L
+#include <optional>
```
michael-loki: The entire point of doing this was to have a consistent api across all platforms, so the lokinet codebase is "clean" and we push all that jank into the 3rd party libraries where possible.

despair86: >macintosh only
```cpp
// ABSL_HAVE_STD_OPTIONAL
//
// Checks whether C++17 std::optional is available.
#ifdef ABSL_HAVE_STD_OPTIONAL
#error "ABSL_HAVE_STD_OPTIONAL cannot be directly set."
#endif

#ifdef __has_include
#if __has_include(<optional>) && __cplusplus >= 201703L && \
    ABSL_INTERNAL_MACOS_HAS_CXX_17_TYPES
#define ABSL_HAVE_STD_OPTIONAL 1
#endif
#endif
```
No.

michael-loki: That gets set later in:

https://github.com/abseil/abseil-cpp/blob/d78310fe5a82f2e0e6e16509ef8079c8d7e4674e/absl/base/config.h#L435-L448

despair86: this project does not support Microsoft C++...had to abandon it early on because `getopt(3)` 

```c
#if defined(_MSC_VER) && _MSC_VER >= 1910 && ((defined(_MSVC_LANG) && _MSVC_LANG > 201402) || __cplusplus > 201402)
```

majestrate:
ready for merge?

despair86:
it is now

PR #284, opened 2019-02-06
Convert llarp::Router into an abstract base class

michael-loki:
This doesn't convert it everywhere, but in "most" places. Further iterations on this would be:

- Move `AbstractRouter` into its own header, which would hopefully break some of the dependency cycles and improve compile times
- Replace the rest of the usage of `Router` in the majority of places with the interface.


majestrate:
i like this

majestrate:
ready for merge?

michael-loki:
yup

PR #286, opened 2019-02-07
Refactor and add tests for bits.hpp

michael-loki:
Expand to handle any container (not just C-style arrays).

clang manages to vectorise this which is cool - https://godbolt.org/z/lMD1eD for vector, https://godbolt.org/z/7ySPBe for array. for smaller array the assembler looks ~equivalent.


michael-loki:
Also I moved the stream out operator for `ExitInfo` into the `.cpp` and reduced the number of components which need to be compiled when `bits.hpp` changes from ~90 to less than 10.

despair86:
https://godbolt.org/z/fhiBB2 vector (32-bit)
https://godbolt.org/z/wnWFEi array (32-bit)

PR #294, opened 2019-02-08
Staging

majestrate:
* fix makefile jobserver issue
* don't send more than 1 introset in reply to a lookup
* fix deadlock

### DIFF
--- a/llarp/service/context.cpp
+++ b/llarp/service/context.cpp
@@ -82,11 +82,18 @@ namespace llarp
           ++itr;
         }
       }
+
+      auto ep = getFirstEndpoint();
+      if(!ep)
+        return;
+      std::vector< RouterID > expired;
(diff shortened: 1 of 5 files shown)


llarp/service/context.cpp
```diff
@@ -92,6 +92,8 @@ namespace llarp
           expired.emplace_back(rc.pubkey);
         return true;
       });
+      // TODO: we need to stop looking up service nodes that are gone forever
+      // how do?
```
neuroscr: well you need to know they're gone forever and we can never know that. Need to adjust your where clause there. I imagine some type of timestamp threshold comes into play here

majestrate: correct, this is a problem of propagating consensous state to lokinet clients

PR #300, opened 2019-02-09
add rpc endpoint for dumping internal state

majestrate:


llarp/dht/bucket.hpp
```diff
@@ -13,13 +14,25 @@ namespace llarp
...
     {
       using BucketStorage_t = std::map< Key_t, Val_t, XorMetric >;
       using Random_t        = std::function< uint64_t() >;
 
       Bucket(const Key_t& us, Random_t r) : nodes(XorMetric(us)), random(r){};
 
+      void
+      ExtractStatus(util::StatusObject& obj) const override
+      {
+        for(const auto& item : nodes)
```
michael-loki: DYC:

```c++
std::for_each(nodes.begin(), nodes.end(), [&obj](const auto& item) {
  std::string keyname = item.first.ToHex();
  util::StatusObject itemObj;
  item.second.ExtractStatus(itemObj);
  obj.PutObject(keyname.c_str(), itemObj);
});
```

llarp/dht/node.hpp
```diff
@@ -46,6 +52,24 @@ namespace llarp
...
+      void
+      ExtractStatus(util::StatusObject& obj) const override
+      {
+        obj.PutInt("timestamp", introset.T);
+
+        std::vector< util::StatusObject > introsObjs(introset.I.size());
+        size_t idx = 0;
+        for(const auto intro : introset.I)
+        {
+          auto& introObj = introsObjs[idx++];
```
michael-loki: this is a little suspect just imo - I'm not a fan of putting `i++` inside longer sequences, because it makes it easy to introduce UB. transform gives something cleaner:

```c++
std::transform(introset.I.begin(), introset.I.end(), std::back_inserter(introsObjs), [](const auto& intro) -> util::StatusObject {
  util::StatusObject introObj;
  introObj.PutString("router", intro.router.ToHex());
  introObj.PutInt("expiresAt", intro.expiresAt);
  introObj.PutInt("latency", intro.latency);
  introObj.PutInt("version", intro.version);

  return introObj;
});
```

test/test_libabyss.cpp
```diff
@@ -26,7 +26,7 @@ struct AbyssTestBase : public ::testing::Test
   void
   AssertMethod(const std::string& meth) const
   {
-    ASSERT_TRUE(meth == method);
+    ASSERT_EQ(meth, method);
```
michael-loki: 👍

llarp/util/status.hpp
```diff
@@ -0,0 +1,63 @@
...
+      void
+      PutObject(const char* name, StatusObject& obj);
+
+      void
+      PutStringArray(const char* name, std::vector< std::string >& arr);
+
+      void
+      PutObjectArray(const char* name, std::vector< StatusObject >& arr);
+
+      StatusObject_Impl Impl;
```
michael-loki: DYC:

```c++
struct StatusObject
{
  using Variant = std::variant<uint64_t, std::string, bool, std::unique_ptr< StatusObject >, std::vector< std::string >, std::vector< StatusObject > >;
  using value_type = std::tuple<std::string_view, Variant>;

  void put(std::string_view name, const Variant& value);
 
  void put (const value_type& value);
  
  StatusObject();
  StatusObject(std::initializer_list<value_type> init) { /* magic */ }
};
```

then you should be able to initialize like:

```c++
StatusObject obj{{"val1", "my string"}, {"val2", false}, {"val3", {"one", "two", "three"}}};
```

llarp/dht/node.hpp
michael-loki: Note combined with https://github.com/loki-project/loki-network/pull/300#discussion_r255365608, you might end up with:

```c++
util::StatusObject introObj{
  {"router", intro.router.ToHex()}, 
  {"expiresAt", intro.expiresAt},
  {"latency", intro.latency},
  {"version", intro.version}
};

return introObj;
```

llarp/util/status.hpp
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    /// an entity that has a status that can be extracted
+    struct IStateful
+    {
+      virtual ~IStateful(){};
+
+      virtual void
+      ExtractStatus(StatusObject& state) const = 0;
```
michael-loki: Any reason why this doesn't just return a state object? That makes it easier to chain across multiple objects.

majestrate: such a pattern seems a bit messy to me if you end up chaining calls it's less clear what is going on maybe.

llarp/util/status.hpp
majestrate: the internal implementation is optional on compile time, i.e. if we don't want to compile with jsonrpc, would this pattern allow for that?

michael-loki: It should do.

majestrate:
On Mon, Feb 11, 2019 at 05:12:01AM -0800, michael-loki wrote:
> michael-loki commented on this pull request.
> 
> 
> 
> > +      void
> +      PutString(const char* name, const std::string& val);
> +
> +      void
> +      PutBool(const char* name, bool val);
> +
> +      void
> +      PutObject(const char* name, StatusObject& obj);
> +
> +      void
> +      PutStringArray(const char* name, std::vector< std::string >& arr);
> +
> +      void
> +      PutObjectArray(const char* name, std::vector< StatusObject >& arr);
> +
> +      StatusObject_Impl Impl;
> 
> It should do.

I'll get on it now.

llarp/util/status.hpp
majestrate: you're probably right tho

michael-loki:
This is a bit awkward, its introduced a cyclic dependency between libabyss and the rest of lokinet

majestrate:
On Tue, Feb 12, 2019 at 02:18:32AM -0800, michael-loki wrote:
> This is a bit awkward, its introduced a cyclic dependency between libabyss and the rest of lokinet
ooooooofff

PR #317, opened 2019-02-18
 Fix undefined behaviors and misaligned member access / macos compile fix

neuroscr:
I think the Michael commits zero'd out because they were already applied.
Also llu (MacOS) vs lu (linux gcc 7.3) issue


crypto/libntrup/src/ref/random32.c
```diff
@@ -22,6 +22,7 @@ small_random32(void)
 #else
   unsigned char x[4];
   randombytes(x, 4);
-  return x[0] + (x[1] << 8) + (x[2] << 16) + (x[3] << 24);
+  uint32_t x4 = x[3] << 4;
```
majestrate: this should be  `uint32_t x4 = x[3] << 24;`

llarp/util/encode.hpp
```diff
@@ -80,7 +80,8 @@ namespace llarp
   Base32Encode(const V& value, Stack& stack)
   {
     size_t ret = 0, pos = 1;
-    int bits = 8, tmp = value[0];
+    int bits = 8;
+    uint32_t tmp = value[0];
```
majestrate: why `uint32_t` ?

crypto/libntrup/src/ref/random32.c
neuroscr: Holy fuck, what a typo

llarp/util/encode.hpp
neuroscr: encode.hpp:92:15: runtime error: left shift of 230014456 by 8 places cannot be represented in type 'int32_t' (aka 'int')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/admin/Sites/llarp/llarp/util/encode.hpp:92:15 in 

something about shifted a signed 32 bit integer.

PR #320, opened 2019-02-18
[WIP] new wire protocol

majestrate:
don't merge me yet


docs/wire-protocol.txt
```diff
@@ -257,7 +284,7 @@ request protocol upgrade
 <1 byte protocol max version to upgrade to>
 <remaining bytes discarded>
 
-command: 'V' (version upgrade)
+command: 'v' (version upgrade)
```
michael-loki: Is there a reason to change all the casings here? Is it to help migration?

llarp/link/iwp.cpp
```diff
@@ -1,25 +1,165 @@
...
-      return nullptr;
+      using namespace std::placeholders;
+      return NewServer(
+          r->crypto(), r->encryption(), std::bind(&AbstractRouter::rc, r),
+          std::bind(&AbstractRouter::HandleRecvLinkMessageBuffer, r, _1, _2),
+          std::bind(&AbstractRouter::OnSessionEstablished, r, _1),
+          std::bind(&AbstractRouter::CheckRenegotiateValid, r, _1, _2),
+          std::bind(&AbstractRouter::Sign, r, _1, _2),
+          std::bind(&AbstractRouter::OnConnectTimeout, r, _1),
+          std::bind(&AbstractRouter::SessionClosed, r, _1));
```
michael-loki: this is making me feel like `AbstractRouter` is a bit too long. `IRouter` might have been better?

llarp/link/iwp.cpp
```diff
@@ -1,25 +1,165 @@
...
-              llarp::SessionRenegotiateHandler, llarp::SignBufferFunc,
-              llarp::TimeoutHandler, llarp::SessionClosedHandler)
+    NewServer(Crypto* c, const SecretKey& enckey, GetRCFunc getrc,
+              LinkMessageHandler h, SessionEstablishedHandler est,
+              SessionRenegotiateHandler reneg, SignBufferFunc sign,
+              TimeoutHandler t, SessionClosedHandler closed)
     {
-      // TODO: implement me
-      return nullptr;
+      return std::unique_ptr< ILinkLayer >(
```
michael-loki: FYI `return std::make_unique<LinkLayer>(c, .....);` is a safer option in the general case (doesn't make a difference here)

llarp/link/iwp_internal.hpp
```diff
@@ -17,6 +18,117 @@ namespace llarp
   {
     struct LinkLayer;
 
+    using FlowID_t = llarp::AlignedBuffer< 32 >;
```
michael-loki: I would do `struct FlowID_t : public AlignedBuffer< 32 > {};` here, it helps prevent implicit conversions between other typedefs to the same size `AlignedBuffer`.

docs/wire-protocol.txt
majestrate: disctinction from inner and outer message commands, stylistic shtick.

llarp/link/iwp.cpp
majestrate: probably better `IRouter` but it's too late :DDDDDDDDD

llarp/link/iwp.cpp
majestrate: this code is not done yet.

llarp/link/iwp_internal.hpp
majestrate: it's actually supposed to be a 16 byte thing according to spec so this change is out dated.

majestrate:
migration path is we enable a separate transport of different rank and then switch the rank when it's time to use it as primary.

neuroscr:
We should not run two protocols in parallel. Just hard cut over.

PR #349, opened 2019-02-25
Create Printer - A general-purpose, stateful printer class

michael-loki:
This hopefully makes a lot of the logging/debugging cleaner, as well as gives us a uniform print format (which matches what we have right now)


neuroscr:
cleaner? how?

PR #356, opened 2019-02-27
output pubkey in base32z

neuroscr:
thinking

`lokinet: [NFO] (706) Wed Feb 27 01:15:02 2019 UTC router/router.cpp:1234	0db5bdf8862ce02c356e67f7c3dbbc03ebf0e283b096a414ac6dec996d51f0ac is NOT a valid service node, rejecting`

should be in base32z to have clearer meaning


majestrate:
the way it's done in the rest of the code is to use `RouterID` instead of `PubKey` when printing.

neuroscr:
Weird that it conflicted, not sure where it got that close from... Hopefully Jeff will know how to resolve

Issue #324, opened 2019-02-19
Got recursion not available from 127.0.0.1, trying next server


neuroscr:
Mojave 10.14

```
Admins-MacBook-Pro:~ admin$ nslookup 4et5rg7edgzimjt34ehro4gj9uetddf4qpq6zftxsnsqhkdci4zy.loki 
;; Got recursion not available from 127.0.0.1, trying next server
Server:		8.8.8.8
Address:	8.8.8.8#53

** server can't find 4et5rg7edgzimjt34ehro4gj9uetddf4qpq6zftxsnsqhkdci4zy.loki: NXDOMAIN


Admins-MacBook-Pro:~ admin$ nslookup localhost.loki
;; Got recursion not available from 127.0.0.1, trying next server
Server:		8.8.8.8
Address:	8.8.8.8#53

** server can't find localhost.loki: NXDOMAIN

Admins-MacBook-Pro:~ admin$ nslookup random.snode
;; Got recursion not available from 127.0.0.1, trying next server
Server:		8.8.8.8
Address:	8.8.8.8#53

** server can't find random.snode: NXDOMAIN

Admins-MacBook-Pro:~ admin$ nslookup 10.200.0.1
;; Got recursion not available from 127.0.0.1, trying next server
Server:		8.8.8.8
Address:	8.8.8.8#53

** server can't find 1.0.200.10.in-addr.arpa: NXDOMAIN
```

HOWEVER

```
Admins-MacBook-Pro:~ admin$ nslookup localhost.loki 127.0.0.1
Server:		127.0.0.1
Address:	127.0.0.1#53

Name:	localhost.loki
Address: 10.200.0.1

Admins-MacBook-Pro:~ admin$ nslookup random.snode 127.0.0.1
Server:		127.0.0.1
Address:	127.0.0.1#53

random.snode	canonical name = wpfcwhxfnjbmok6koqcim6p6x7agc54cwfctikhsyn1riz5jj6zy.snode.

Admins-MacBook-Pro:~ admin$ nslookup 4et5rg7edgzimjt34ehro4gj9uetddf4qpq6zftxsnsqhkdci4zy.loki 127.0.0.1
Server:		127.0.0.1
Address:	127.0.0.1#53

Name:	4et5rg7edgzimjt34ehro4gj9uetddf4qpq6zftxsnsqhkdci4zy.loki
Address: 10.200.0.1

Admins-MacBook-Pro:~ admin$ nslookup 10.200.0.1 127.0.0.1
[...]

PR #362, opened 2019-02-28
Fix cname lookups on mojave: issue #324

neuroscr:


neuroscr:
Actually undoes the base32z change... whoops. Not sure how to fix

PR #371, opened 2019-03-03
 Convert to use abseil synchronisation primitives

michael-loki:
also add annotations, and fix several potential data races/deadlocks.


llarp/link/server.cpp
```diff
@@ -175,16 +179,22 @@ namespace llarp
   {
     std::vector< util::StatusObject > pending, established;
 
-    std::transform(m_Pending.begin(), m_Pending.end(),
```
michael-loki: this is have been a data race[1], which when fixed would have become a potential deadlock.

[1] - if this code wasn't using `NullLock`

CMakeLists.txt
```diff
@@ -72,6 +72,10 @@ add_compile_options(-Wvla)
 add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fpermissive>)
 add_compile_options(-Wno-unused-function -Wno-deprecated-declarations -Wno-unknown-pragmas)
 
+if (USING_CLANG)
+  add_compile_options(-Wthread-safety -Wthread-safety-negative)
```
despair86: ah yes more warning on Windows 😟

despair86: let me guess: you're going to end up converting the _entire_ codebase over to abseil at some point ~~runs~~

michael-loki: are we using clang on windows? We can tweak this more if you want?

I'd prefer to keep with std where possible. abseil synchronisation primitives have some useful features. 

std > abseil >> writing it ourselves

despair86: >clang on windows

@michael-loki yes, in fact, i've spent the weekend cutting over to new solaris cross-compile box, only to find that clang v8 RC3 is still a bit broken: https://twitter.com/__rvx86/status/1101709699127349249

despair86: even worse, this is *before* the abseil integration, and in ***debug*** mode: https://spit.mixtape.moe/view/39b1a1db#WSOIKsFTA1vtBm2vmwiW0oGv7AtODkiC

michael-loki: yeah that's weird AF.

```c++
$ echo __ZNKSt9type_infoeqERKS_ | c++filt
std::type_info::operator==(std::type_info const&) const
```

`ccache` is magic.

PR #384, opened 2019-03-07
Staging

majestrate:


llarp/link/utp.cpp
```diff
@@ -475,13 +489,13 @@ namespace llarp
...
       } while(true);
+#endif
     }
 #endif
 
     void
     LinkLayer::Pump()
     {
-      utp_issue_deferred_acks(_utp_ctx);
 #ifdef __linux__
```
neuroscr: why only linux?

llarp/handlers/tun.cpp
```diff
@@ -343,12 +343,6 @@ namespace llarp
         }
         else if(addr.FromString(qname, ".snode"))
         {
-          if(HasPathToSNode(addr.as_array()))
-          {
-            msg.AddINReply(ObtainIPForAddr(addr, true));
-            reply(msg);
-            return true;
-          }
```
neuroscr: why did this get taken out?

llarp/ev/ev_kqueue.cpp
```diff
@@ -374,21 +374,14 @@ llarp_kqueue_loop::tick(int ms)
       llarp::ev_io* ev = static_cast< llarp::ev_io* >(events[idx].udata);
       if(ev)
       {
-        if(events[idx].filter & EV_ERROR)
```
neuroscr: why did we remove the EV_ERROR case?

llarp/link/utp.cpp
majestrate: icmp stuff is hard to do portably

llarp/handlers/tun.cpp
majestrate: the reply is deferred to when we have the ability to send to the service node now.

PR #409, opened 2019-03-20
this is it folks

despair86:
finally landing the cross platform .NET frontend for lokinet (and some relevant packaging patches)
- .NET CLR v2 or later
- now includes its own bootstrap (I should also add it to the command line processing)
- coming soon: configuration, still need to poke TOML-like syntaxes
- log rotation (Can someone get me the API spec for debuglogs.org?)


despair86:
oops wrong branch!

PR #418, opened 2019-03-21
do not conflict with any existing TAP ifs on windows

despair86:
also bring the interface up _before_ faking a DHCP offer

### DIFF
--- a/vendor/libtuntap-master/tuntap-windows.c
+++ b/vendor/libtuntap-master/tuntap-windows.c
@@ -140,7 +140,7 @@ reg_query(char *key_name)
       goto clean;
     }
     /* If its a tap adapter, its all good */
-    if(strncmp(data, "tap", 3) == 0)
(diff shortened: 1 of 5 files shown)


vendor/libtuntap-master/tuntap-windows.c
```diff
@@ -140,7 +140,7 @@ reg_query(char *key_name)
       goto clean;
     }
     /* If its a tap adapter, its all good */
-    if(strncmp(data, "tap", 3) == 0)
+    if(strncmp(data, "tap0901", 7) == 0)
```
neuroscr: seems specific?

despair86: 1. I haven't come across any working binaries for the old TAP v8 driver, probably doesn't work anyway
2. I used to have nord vpn installed, and the old function ended up activating *their* version of TAP v9, because it is also prefixed `tap` (`tapnordvpn`)

PR #421, opened 2019-03-21
track lookup fails and kill outbound context if too many

majestrate:


llarp/service/endpoint.cpp
```diff
@@ -1770,6 +1772,9 @@ namespace llarp
     bool
     Endpoint::OutboundContext::Tick(llarp_time_t now)
     {
+      // we are probably dead af
+      if(m_LookupFails > 16)
+        return true;
```
neuroscr: should we tell the operator? like log something? Otherwise we'll have HS just go offline and not know why?

PR #446, opened 2019-03-25
Metrics Collection System

michael-loki:
3 commits (developed in sync, just grouped some what logically).

1. Adds some useful utilities 
2. Adds metric collection mechanisms 
3. Sample integration with lokinet:
  - Instantiates a metric manager in main
  - Adds a publisher scheduler in main
  - Creates a streaming publisher to `stdout` 
  - Adds some initial collection work to `router/router.cpp`

Work to do (but this should still be mergeable in this state):
- empty metric GC (once a metric is posted once, we always keep track of it)
- json dumper / rotator so we don't fill everyone's disks up
- a remote metrics collector for debugging for lokinet devs (pushing to metric tank or something)
- adding them in more places in the code (the current integration is just sample)

### DIFF
--- a/llarp/util/metrics_types.hpp
+++ b/llarp/util/metrics_types.hpp
@@ -43,13 +43,13 @@ namespace llarp
       float m_scale;
       const char *m_format;
 
-      static const char *DEFAULT_FORMAT;
+      static constexpr char DEFAULT_FORMAT[] = "%f";
 
-      FormatSpec() : m_scale(1.0), m_format(DEFAULT_FORMAT)
+      constexpr FormatSpec() : m_scale(1.0), m_format(DEFAULT_FORMAT)
       {
       }
 
-      FormatSpec(float scale, const char *format)
+      constexpr FormatSpec(float scale, const char *format)
           : m_scale(scale), m_format(format)
       {
       }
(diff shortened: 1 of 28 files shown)


michael-loki:
edit: this fails debian builds

llarp/util/metrics_types.hpp
```diff
@@ -0,0 +1,532 @@
+#ifndef LLARP_METRICS_TYPES_HPP
+#define LLARP_METRICS_TYPES_HPP
+
+#include <util/string_view.hpp>
+#include <util/threading.hpp>
+#include <util/types.hpp>
+
+#include <absl/types/span.h>
+#include <iosfwd>
+
```
despair86: include what you use:
`absl::optional` (falls through to `std::optional` in modern c++)
`cstring`
`set`
`memory`
to fix non-macOS

michael-loki: yeah, just fixing up debian. looks like some of that is running into a fun gcc constexpr bug though.

include what you use is a good move, I'm bad at remembering.

michael-loki: Fixed.

michael-loki:
Debian is now happy.

majestrate:
merge?

michael-loki:
Go for it, this is into staging not master which makes sense to me.

PR #452, opened 2019-03-27
poll(2) event loop for svr4

despair86:
also ported over the epoll tick changes to all other epoll-like abstractions


majestrate:
ready for merge?

despair86:
yes go ahead

PR #473, opened 2019-03-30
use fixed path for lokinet-bootstrap

majestrate:
dont run unittest with CROSS=ON


majestrate:
this is for debian init scripts

PR #493, opened 2019-04-07
Add JSON publisher with configs

michael-loki:


majestrate:
neato, this ready?

michael-loki:
yup.

Technically this doesn't produce JSON, it produces JSONL (where each line is its own JSON object - http://jsonlines.org/), which means you can just append to the file easily without backtracking etc.

PR #497, opened 2019-04-09
Fixup build on macOS

michael-loki:


llarp/ev/pipe.cpp
```diff
@@ -12,8 +12,14 @@ bool
 llarp_ev_pkt_pipe::Start()
 {
   int _fds[2];
+#if defined(__APPLE__)
+  if(pipe(_fds) == -1)
+#else
   if(pipe2(_fds, O_DIRECT | O_NONBLOCK) == -1)
```
michael-loki: macOS doesn't support `pipe2` or `O_DIRECT`

majestrate: the fd should be set to non blocking though

michael-loki: done

PR #502, opened 2019-04-11
add syslog

majestrate:
* refactor logging to have module log streams
* fix up win32 parts so that `make windows` works again


llarp/service/endpoint.cpp
```diff
@@ -1206,25 +1206,6 @@ namespace llarp
...
-      {
-        path = PickRandomEstablishedPath();
-        if(!path)
-        {
-          llarp::LogError(Name(), "no working paths for lookup");
-          hook(remote, nullptr);
-          return false;
-        }
-      }
-
```
neuroscr: no documentation about this change? Is this intentional?

llarp/service/endpoint.cpp
```diff
@@ -1551,7 +1532,7 @@ namespace llarp
       else
       {
         llarp::LogInfo(Name(), " updating introset");
-        UpdateIntroSet(false);
+        UpdateIntroSet(true);
```
neuroscr: no documentation about this change? Is this intentional?

majestrate: i think my local changes got sucked into this pr

Issue #449, opened 2019-03-26
MetricTank metric publisher privacy issues


michael-loki:
We want to have an **explicitly** opt in metric publisher which would help lokinet developers collect performance metrics on testnets only by posting to a metrictank or similar instance running somewhere.

Making this a public issue to brainstorm ideas on how we make sure this is very very opt in, and we don't accidentally leak privacy.

My ideas:
1. command line parameter
2. environment variable
3. log in massive letters at start up
4. require a "yes" at `stdin` to enable

I think that a combination of 1, 2 and 3 is sufficient. 4 would probably make automation more painful than we want it to be.

KeeJef:
I'm assuming startup flags would work fine, its what is usually used across the Lokinet apps (Lokid, launcher, lokinet), so 
`./lokinet --expose-metrics `
or something similar

michael-loki:
We went with option in config file + environment variable + big warning.

PR #517, opened 2019-04-14
MetricTank metric collector

michael-loki:
- by default this isn't enabled, and requires two changes (per #449) to enable:
  - section in config file
  - environment variable 
- big warning once enabled
- collect some tags:
  - username 
  - system name
- tags can be added/overridden from the config (without needing to change code)

worker queue + thread keeps this encapsulated, without needing to pollute the event loop.

testing done:
- built on macOS and windows natively
- debian docker build
- verified metrics were collected in a local metrictank instance
- verified nothing bad happened when the instance wasn't available


majestrate:
is this ready for merge?

michael-loki:
yup.

PR #518, opened 2019-04-15
Allow builds on MSVC

michael-loki:
main bulk is importing cxxopts into `vendor/`

this is 100% a WIP, and I've not tested most of this (because it still doesn't gorram link!)


llarp/win32/abyss.rc
```diff
@@ -89,7 +89,7 @@ BEGIN
 #ifdef LLARP_RELEASE_MOTTO
             VALUE "ProductVersion", VERSION_STRING(0.4.0, RELEASE_MOTTO, GIT_REV)
 #else
-            VALUE "ProductVersion", VERSION_STRING(0.4.0-dev-, GIT_REV)
```
despair86: hmmm there should be a way to detect when /lib/cpp is being used as opposed to the K&R era RC preprocessor

CMakeLists.txt
```diff
@@ -320,7 +333,7 @@ if(UNIX)
 elseif(WIN32)
     get_filename_component(LIBTUNTAP_IMPL ${TT_ROOT}/tuntap-windows.c ABSOLUTE)
     get_filename_component(EV_SRC "llarp/ev/ev_win32.cpp" ABSOLUTE)
-    add_definitions(-DWIN32_LEAN_AND_MEAN -DWIN32 -DWINVER=0x500 -D_WIN32_WINNT=0x500)
+    add_definitions(-DWIN32_LEAN_AND_MEAN -DWIN32 -DWINVER=0x0602 -D_WIN32_WINNT=0x0602)
```
despair86: set Windows 8 requirement conditionally, the UNIX tools are slightly dated because they track reactos

despair86:
I love inline libs, we'll keep the new stuff then

llarp/win32/abyss.rc
michael-loki: probably. I'm not sure if this is even needed with MSVC, this is what I started digging into.

CMakeLists.txt
michael-loki: 👍 I'm just trying to get to a stage where this builds + tests pass first 😉

llarp/nodedb.cpp
```diff
@@ -173,15 +173,15 @@ llarp_nodedb::Insert(const llarp::RouterContact &rc)
...
 }
 
-ssize_t
+size_t
 llarp_nodedb::Load(const fs::path &path)
 {
   std::error_code ec;
   if(!fs::exists(path, ec))
   {
     return -1;
```
majestrate: this should be set to 0 if the return type is `size_t`

llarp/win32/abyss.rc
michael-loki: reverted.

llarp/nodedb.cpp
michael-loki: I'm gonna revert this.

michael-loki:
@despair86 / @majestrate this is ready for merge imo

CMakeLists.txt
despair86: I

michael-loki: ah I missed this.

despair86:
ok merging as is, i'll get a patch in shortly

PR #526, opened 2019-04-16
Make JSON metrics opt-in

michael-loki:


michael-loki:
this is a follow on from #521

PR #537, opened 2019-04-21
make config file used when provided in arguments

majestrate:


daemon/main.cpp
```diff
@@ -153,12 +154,12 @@ main(int argc, char *argv[])
       asRouter = true;
     }
 
-	if(result.count("config_file") > 0)
+	if(result.count("config") > 0)
     {
-      auto vec = result["config_file"].as< std::vector< std::string > >();
+      auto vec = result["config"].as< std::string >();
```
michael-loki: Rename vec to something else?

majestrate: updated

PR #539, opened 2019-04-22
Make IntroSet PoW be a unique_ptr

michael-loki:


llarp/service/IntroSet.hpp
```diff
@@ -96,9 +78,20 @@ namespace llarp
       bool
       operator==(const IntroSet& other) const
```
michael-loki: this bit is a ~behaviour~ change.

PR #542, opened 2019-04-23
Lokinet iOS

nielsandriesse:
To compile Lokinet for iOS, execute the following from the Lokinet root directory:

```
cmake -G Xcode -DCMAKE_TOOLCHAIN_FILE=ios/ios.cmake -DDEPLOYMENT_TARGET=10.0 -B ios/Core -S .
cmake --build ios/Core --config Debug --target lokinet
```

You can safely ignore the code signing warning that's emitted.


nielsandriesse:
We do need the project files, shared settings, assets and storyboards. Perhaps we could do without the project files for the dependencies but that's about it. An option would be to put all of this in a separate repo and use git submodules but that'd add a bunch of overhead.

despair86:
oh keep the assets of course
but the project may not even work properly if _any_ absolute paths are present anywhere 🤔

despair86:
i.e. this is the same reason I don't push a Visual Studio project into the tree

nielsandriesse:
Things should be way better now :)

CMakeLists.txt
```diff
@@ -324,6 +330,9 @@ if(UNIX)
   elseif (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
       set(LIBTUNTAP_IMPL ${TT_ROOT}/tuntap-unix-darwin.c ${TT_ROOT}/tuntap-unix-bsd.c)
       get_filename_component(EV_SRC "llarp/ev/ev_kqueue.cpp" ABSOLUTE)
+  elseif (IOS)
```
liilac: See the three lines above.  The implementation for Darwin platforms (which iOS is one of) is the same.

Prior to cmake 3.14, compiling for iOS has `CMAKE_SYSTEM_NAME` set to `Darwin`.  3.14+ have better support for mobile Darwin platforms, and set it to `iOS`, `watchOS`, etc.  See [here](https://github.com/leetal/ios-cmake/blob/3e5697cd87ca7882783ad3a22c86369256f69f6d/ios.toolchain.cmake#L420), lines 420 to line 444 inclusive.

Depending on maintainer preference, and the cmake versions we're using, probably should remove the lines 333-335 you've added, and amend line 330 to read
```cmake
  # this is a standard flag used in the iOS, macOS, etc, toolchains
  elseif (APPLE)
```
or
```cmake
  # prior to cmake 3.14, iOS/tvOS/watchOS set this to Darwin
  # cmake 3.14+ set it to iOS, tvOS, etc as appropriate
  elseif(${CMAKE_SYSTEM_NAME} MATCHES "(Darwin|(i|tv|watch)OS)")
```

CMakeLists.txt
```diff
@@ -419,17 +428,19 @@ else()
...
   endif(NOT WIN32)
 
   add_log_tag(${EXE})
 
-  install(TARGETS ${EXE} RUNTIME DESTINATION bin)
-  if(WIN32)
-    install(PROGRAMS ${CMAKE_SOURCE_DIR}/lokinet-bootstrap.exe DESTINATION bin)
-  else()
-    install(PROGRAMS ${CMAKE_SOURCE_DIR}/lokinet-bootstrap DESTINATION bin)
+  if(NOT IOS)
```
liilac: This is unnecessary and possibly mistaken?

The binaries installed here don't have to be included in an iOS extension (this can be controlled from Xcode, AppCode, or whatever you're using), and you may wish to run these binaries internally.

crypto/salsa20/stream_salsa20.c
```diff
@@ -66,6 +66,11 @@ _crypto_stream_salsa20_pick_best_implementation(void)
 {
   if(implementation)
     return 0;
+#if defined(iOS)
```
liilac: What's the point of this?

If AVX2 and SSE2 aren't available, which they aren't on iOS hardware, this implementation will be chosen anyway (see lines [64](https://github.com/loki-project/loki-network/blob/35cc9fcb52bc6cfe0cb3ec3ca058e9ac7a05f920/crypto/salsa20/stream_salsa20.c#L64)-[91](https://github.com/loki-project/loki-network/blob/35cc9fcb52bc6cfe0cb3ec3ca058e9ac7a05f920/crypto/salsa20/stream_salsa20.c#L91))

If this was to fix a compile time error, you were probably trying to build for the wrong platform.

crypto/salsa20/xmm6int/salsa20_xmm6int-sse2.c
```diff
@@ -1,4 +1,4 @@
-#if defined(ANDROID) || defined(RPI)
+#if defined(ANDROID) || defined(iOS) || defined(RPI)
```
liilac: This should read
```c
#if __SSE2__
```
instead, using inbuild compiler support to detect whether the SSE2 instruction is present, rather then relying on explicit build flags/defines.

crypto/salsa20/xmm6int/salsa20_xmm6int-sse2.h
```diff
@@ -3,7 +3,7 @@
 
 #include "../stream_salsa20.h"
 #include <sodium/crypto_stream_salsa20.h>
-#ifndef ANDROID
+#if !defined(ANDROID) && !defined(iOS)
```
liilac: This should read
```c
#if __SSE2__
```
instead, using inbuilt compiler support to detect whether the SSE2 instruction is present, rather then relying on explicit build flags/defines.

ios/Packet Tunnel Provider/Packet Tunnel Provider.entitlements
```diff
@@ -0,0 +1,18 @@
...
+	<array>
+		<string>packet-tunnel-provider</string>
+	</array>
+	<key>com.apple.developer.networking.vpn.api</key>
+	<array>
+		<string>allow-vpn</string>
+	</array>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.niels-andriesse.loki-network</string>
```
liilac: The group should use loki-controlled namespace.

Team Apple Developer accounts have also been discussed with Simon before, which he agreed to, if there are concerns around ease of signing and such.

ios/Loki Network/Loki Network.entitlements
```diff
@@ -0,0 +1,18 @@
...
+	<array>
+		<string>packet-tunnel-provider</string>
+	</array>
+	<key>com.apple.developer.networking.vpn.api</key>
+	<array>
+		<string>allow-vpn</string>
+	</array>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.niels-andriesse.loki-network</string>
```
liilac: This group should use loki-controlled namespace, if this in included in the repo at all.

Team Apple Developer accounts have been discussed with Simon before as well, and agreed to, if ease of signing and such is a concern.

ios/Loki Network/LKDaemon.swift
```diff
@@ -0,0 +1,35 @@
...
+    func configure(isDebuggingEnabled: Bool = false, completionHandler: @escaping (Error?) -> Void) {
+        // Prepare
+        let directoryPath = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!.path
+        // Enable debugging mode if needed
+        if isDebuggingEnabled { _llarp_enable_debug_mode() }
+        // Generate configuration file
+        let configurationFilePath = directoryPath + "/" + "lokinet-configuration.ini"
+        _llarp_ensure_config(configurationFilePath, directoryPath, true, false)
+        // Download bootstrap file
+        let remoteBootstrapFilePath = "https://i2p.rocks/i2procks.signed"
```
majestrate: ideally this should be configurable if it's not already

ios/Loki Network/lokinet-wrapper.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+extern "C"
+void
+_llarp_enable_debug_mode() {
+    llarp::SetLogLevel(llarp::eLogDebug);
+}
+
+extern "C"
+bool
+_llarp_ensure_config(const char *fname, const char *baseDir, bool overwrite, bool asRouter)
```
majestrate: these functions are already exposed via C ffi, this is redundant.

ios/Packet Tunnel Provider/Packet Tunnel Provider.entitlements
nielsandriesse: Of course, this is just a temporary fix to be able to use network extensions. We'll switch this over once the team developer account is available.

ios/Loki Network/Loki Network.entitlements
nielsandriesse: See comment above

ios/Loki Network/LKDaemon.swift
nielsandriesse: I thought about that, and there are other things that we might want to make configurable as well at some point, but I think it's a bit too early to be concerned with things like that right now.

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ option(AMD_RYZEN_HACK "hack for AMD Ryzen FPU bug (support FMA3 and FMA4 in FPU,
 if (NOT MSVC)
 option(STATIC_LINK_RUNTIME "link statically against compiler runtime, standard library and pthreads")
 endif()
-option(NON_PC_TARGET "non-pc target build: iphone, andriod, embedded non-i386 SBC, etc" )
+option(NON_PC_TARGET "non-pc target build: iOS, Android, embedded non-i386 SBC, etc" )
```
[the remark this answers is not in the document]
despair86: you can use AVX2 in i386 protected mode @liilac

crypto/salsa20/stream_salsa20.c
despair86: does the iPhone simulator not exist? last time I checked, its target is `[i586|x86_64]-apple-iphoneos*` or similar

ios/Loki Network/lokinet-wrapper.cpp
nielsandriesse: You're right, this was leftover from an earlier approach :)

CMakeLists.txt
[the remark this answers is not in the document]
nielsandriesse: I think this is a question for @majestrate, @despair86 or @odilitime

CMakeLists.txt
```diff
@@ -155,7 +161,7 @@ endif()
 
 if(DEBIAN)
   add_definitions(-DDEBIAN)
-elseif(NOT ANDROID AND NOT NON_PC_TARGET)
+elseif(NOT ANDROID AND NOT IOS AND NOT NON_PC_TARGET)
```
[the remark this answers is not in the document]
nielsandriesse: Thoughts, @majestrate / @despair86 / @odilitime?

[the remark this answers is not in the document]
despair86: just add `i[3-7]86` as well as `[x86-|AaMmDd]64` to the match *shrugs*
not exactly sure how regexes work in cmake

nielsandriesse:
Addressed your comments, @liilac

majestrate:
close PR or get it rebased?

gamabuntan:
@majestrate I'll be updating this soon

majestrate:
please don't include binary libs in the repo

PR #543, opened 2019-04-23
Fixup casing

michael-loki:
Done via the github UI


michael-loki:
I would say go for squash and merge here 🤷‍♀️

PR #544, opened 2019-04-23
Remove ability for service::Endpoint to have its own ev

michael-loki:
This functionality isn't used at all and simplifies the headache here

### DIFF
--- a/llarp/handlers/null.hpp
+++ b/llarp/handlers/null.hpp
@@ -31,6 +31,12 @@ namespace llarp
       {
         return false;
       }
+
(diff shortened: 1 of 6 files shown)


majestrate:
this code is required for the android and ios code

michael-loki:
Ah. Will update my notes.

majestrate:
the idea is that the endpoint needs to own its own event loop in 2 cases:
* running when owning its own network namespace (something i want to do soon)
* running when someone else manages the packet io (android/ios)

Issue #547, opened 2019-04-23
client.ini in config.cpp


neuroscr:
**Describe the bug**
```
[NFO] (624) Tue Apr 23 13:30:52 2019 PDT config.cpp:123    Attempting to create config file /Users/rtharp/.lokinet/lokinet.ini
[ERR] (624) Tue Apr 23 13:30:52 2019 PDT config.cpp:131    /Users/rtharp/.lokinet/client.ini currently exists, please use -f to overwrite
[NFO] (624) Tue Apr 23 13:30:52 2019 PDT daemon/main.cpp:272    Running from: /Users/rtharp/Sites/loki-network
[NFO] (624) Tue Apr 23 13:30:52 2019 PDT daemon/main.cpp:273    Using config file: /Users/rtharp/.lokinet/lokinet.ini
[ERR] (624) Tue Apr 23 13:30:52 2019 PDT context.cpp:50    failed to load config file /Users/rtharp/.lokinet/lokinet.ini
```

**To Reproduce**
`sudo ./lokinet` with no `~/.lokinet/lokinet.ini`

**Screenshots Or Logs**
```
[NFO] (624) Tue Apr 23 13:30:52 2019 PDT config.cpp:123    Attempting to create config file /Users/rtharp/.lokinet/lokinet.ini
[ERR] (624) Tue Apr 23 13:30:52 2019 PDT config.cpp:131    /Users/rtharp/.lokinet/client.ini currently exists, please use -f to overwrite
[NFO] (624) Tue Apr 23 13:30:52 2019 PDT daemon/main.cpp:272    Running from: /Users/rtharp/Sites/loki-network
[NFO] (624) Tue Apr 23 13:30:52 2019 PDT daemon/main.cpp:273    Using config file: /Users/rtharp/.lokinet/lokinet.ini
[ERR] (624) Tue Apr 23 13:30:52 2019 PDT context.cpp:50    failed to load config file /Users/rtharp/.lokinet/lokinet.ini
```

**Device and Operating system (please complete the following information):**
 - OS: macos (but probably all)
-  Device: macpro
 - Lokinet Version number or Git commit hash: 0.4

PR #546, opened 2019-04-23
Make help more useful

michael-loki:
```
$ lokinet -h
Lokinet is a private, decentralized and IP based overlay network for the internet
Usage:
  lokinet [OPTION...] positional parameters

  -v, --verbose     Verbose
  -h, --help        help
  -g, --generate    generate config
  -r, --router      run as router
  -f, --force       overwrite
      --config arg  path to configuration file
```

PR #548, opened 2019-04-23
address issue #547, remove client.ini, update ini defaults, and clean…

neuroscr:
… up help message

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -30,7 +30,7 @@ handle_signal(int sig)
 int
 printHelp(const char *argv0, int code = 1)
 {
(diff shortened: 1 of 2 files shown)


daemon/main.cpp
```diff
@@ -30,7 +30,7 @@ handle_signal(int sig)
 int
 printHelp(const char *argv0, int code = 1)
 {
-  std::cout << "usage: " << argv0 << " [-h] [-v] [-g|-c] config.ini"
+  std::cout << "usage: " << argv0 << " [-h] [-v] [-g|-r [-f]] [config.ini]"
```
michael-loki: do you wanna revert and let #546 do this instead?

PR #552, opened 2019-04-23
MacOS package scripts

neuroscr:


neuroscr:
apparently include #551 fug

osx-setup/resources/conclusion.html
```diff
@@ -0,0 +1,3 @@
+IRC or Discord is likely the best place for help with LokiNET
```
michael-loki: solid html here

PR #553, opened 2019-04-23
MacOS scripts to create package installer

neuroscr:


neuroscr:
without other PR now!

PR #564, opened 2019-04-26
fix Issue 563

majestrate:
enables using subdomains on loki and snode addresses
also adds some unit tests

### DIFF
--- a/llarp/service/address.cpp
+++ b/llarp/service/address.cpp
@@ -6,25 +6,62 @@ namespace llarp
 {
   namespace service
   {
+    const std::vector< std::string > Address::AllowedTLDs = {".loki", ".snode"};
+
+    bool
+    Address::PermitTLD(const char* tld)
+    {
+      std::string gtld(tld);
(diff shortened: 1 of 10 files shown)


llarp/service/address.cpp
```diff
@@ -6,25 +6,62 @@ namespace llarp
 {
   namespace service
   {
+    const std::vector< std::string > Address::AllowedTLDs = {".loki", ".snode"};
```
michael-loki: I would say `std::set` for how this is being used

llarp/service/address.cpp
```diff
@@ -6,25 +6,62 @@ namespace llarp
...
+
+    bool
+    Address::PermitTLD(const char* tld)
+    {
+      std::string gtld(tld);
+      std::transform(gtld.begin(), gtld.end(), gtld.begin(), ::tolower);
+      for(const auto& allowed : AllowedTLDs)
```
michael-loki:
```c++
return AllowedTLDS.find(gtld) != AllowedTLDS.end();
```

PR #565, opened 2019-04-27
Add kubernetes config file for test net

ewhal:


contrib/kubernetes/kubernetes.yaml
```diff
@@ -0,0 +1,76 @@
...
+    metadata:
+      annotations:
+        description: loki-network lokinet
+      creationTimestamp: null
+      labels:
+        app: loki-network
+      name: loki-network
+    spec:
+      containers:
+        - image: 092763672147.dkr.ecr.us-east-1.amazonaws.com/loki/loki-network:latest
```
michael-loki: Where is the dockerfile for this ?

ewhal: The docker file I used was docker/router.Dockerfile but the image should definitely be moved to docker hub in the future

majestrate: do we know if loki project has an official dockerhub?

PR #572, opened 2019-04-30
Refactor well named functionality in service::Endpoint into new struct

michael-loki:
Does two things:
- simplifies logic in the main class
- makes these things more testable/composable

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -10,6 +10,7 @@
 #include <nodedb.hpp>
 #include <profiling.hpp>
 #include <router/abstractrouter.hpp>
+#include <service/endpoint_util.hpp>
 #include <service/hidden_service_address_lookup.hpp>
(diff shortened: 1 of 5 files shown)


llarp/service/endpoint.cpp
```diff
@@ -316,57 +271,20 @@ namespace llarp
...
-          {
-            itr->second->Stop();
-            m_DeadSessions.emplace(itr->first, std::move(itr->second));
-            itr = m_RemoteSessions.erase(itr);
-          }
-          else
-            ++itr;
-        }
-      }
+      EndpointUtil::TickRemoteSessions(now, m_RemoteSessions, m_DeadSessions);
```
michael-loki: I don't like this particularly. In C++17 this could just be:

```c++
m_DeadSessions.merge(EndpointUtil::TickRemoteSessions(now, m_RemoteSessions));
```

which I think would cleaner imo

PR #589, opened 2019-05-05
Add AppArmor profile

l-n-s:


majestrate:
will there be a variant for when lokinet is at `/usr/bin/lokinet` ? (for debian packaging)

l-n-s:
AppArmor matches process by the path, so it needs to be renamed and changed. I can make a debian version, sure.

l-n-s:
Now it's debianized

PR #597, opened 2019-05-07
Remove per-platform conditionals in cmake files

michael-loki:
@despair86 can you verify solaris/windows please?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,24 +245,16 @@ add_subdirectory(libutp)
 add_subdirectory(llarp)
 add_subdirectory(libabyss)
 
-if (NOT WIN32)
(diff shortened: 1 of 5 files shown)


despair86:
wait a minute
the platform builds are very *slightly different* in their construction in the first place:

- Windows uses C++17 Filesystem TS, all others do not
- Each platform group has its own event loop implementation
- Windows leaves most of abseil unused in favour of equivalent C++17 functionality

otherwise none of the conditionals would exist in the first place

michael-loki:
Yup, but those definitions are moved to per-arch files (rather than branching in the CMakeLists.txt).

michael-loki:
closing stale review

PR #645, opened 2019-06-13
json logging option

majestrate:


llarp/router/router.cpp
```diff
@@ -923,6 +924,19 @@ namespace llarp
     }
     else if(StrEq(section, "logging"))
     {
+      if(strlen(key) == 0 && strlen(val) == 0)
```
neuroscr: kind of a hack here and breaks how INIs work but probably fine for now

PR #649, opened 2019-06-16
Fix event loop tests post-libuv

michael-loki:
my understanding:

- these tests unconditionally call `loop->add_ev`, and assert the result.
- in `ev_libuv.hpp`, `add_ev` always returns false: https://github.com/loki-project/loki-network/blob/3a7d74b08c8ec347b2efa85e8aae51a4595ef1f4/llarp/ev/ev_libuv.hpp#L79-L84
- `llarp_make_ev_loop` makes a libuv event loop on !windows


despair86:
`llarp_make_ev_loop` has the correct behaviour on Windows as well

michael-loki:
@despair86 - I don't see that here: https://github.com/loki-project/loki-network/blob/3a7d74b08c8ec347b2efa85e8aae51a4595ef1f4/llarp/ev/ev.cpp#L19-L32

despair86:
@michael-loki what do you mean? In the libuv refactor, the win32 codepath was left alone, such that from what i'm looking at there, libuv and windows do the same thing: their respective event loop object is created as a `shared_ptr`?

michael-loki:
because of points 1+2 in my PR. the impl of `add_ev` in `ev_win32` actually *does* something: https://github.com/loki-project/loki-network/blob/3a7d74b08c8ec347b2efa85e8aae51a4595ef1f4/llarp/ev/ev_win32.cpp#L638-L653

despair86:
ohhhh
I read this PR upside down then
my mistake 😅

PR #650, opened 2019-06-17
Move messages code to right subdir

michael-loki:


majestrate:
this may conflict with @tewinget 's LRSM changes

michael-loki:
Since it’s mostly a clean move, I think a merge would probably resolve it?

majestrate:
On Mon, Jun 17, 2019 at 05:37:34AM -0700, michael-loki wrote:
> Since it’s mostly a clean move, I think a merge would probably resolve it?

i'd like to verify with tom first just to be sure.

llarp/messages/link_message_parser.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <router_contact.hpp>
+#include <util/buffer.hpp>
+#include <util/logger.hpp>
+#include <util/metrics.hpp>
+
+#include <memory>
+
+namespace llarp
+{
+  struct InboundMessageParser::msg_holder_t
```
tewinget: given the nature of this PR, perhaps could/should also rename this to LinkMessageParser?

michael-loki: Will do.

tewinget:
for the files which were moved from `messages/` consider adding `_messages` to those which describe messages.

majestrate:
ready?

michael-loki:
yarp

PR #653, opened 2019-06-18
[WIP] Path build status messages

tewinget:
Adds LR_StatusMessage as a Link Message.  This message type is sent back to a path creator in a manner which mirrors the LR_CommitMessage(s) sent to create a path, such that the hops of the path can inform the path creator of the path creation status (success or fail with reason).

It compiles and there are no meaningful warnings added!

LR_StatusMessage is now used in place of PathConfirmMessage

### DIFF
--- a/llarp/messages/relay_status.cpp
+++ b/llarp/messages/relay_status.cpp
@@ -9,6 +9,7 @@
 #include <util/buffer.hpp>
 #include <util/logger.hpp>
 #include <util/logic.hpp>
+#include <util/memfn.hpp>
 
 #include <functional>
 
@@ -31,9 +32,7 @@ namespace llarp
     {
     }
 
-    ~LRSM_AsyncHandler()
-    {
-    }
+    ~LRSM_AsyncHandler() = default;
 
     void handle()
     {
@@ -48,10 +47,6 @@ namespace llarp
 
   };
 
-
-  LR_StatusMessage::~LR_StatusMessage()
-  {
-  }
 
   bool
   LR_StatusMessage::DecodeKey(const llarp_buffer_t& key, llarp_buffer_t* buf)
@@ -90,14 +85,7 @@ namespace llarp
   void
   LR_StatusMessage::Clear()
   {
-    frames[0].Clear();
-    frames[1].Clear();
-    frames[2].Clear();
-    frames[3].Clear();
-    frames[4].Clear();
-    frames[5].Clear();
-    frames[6].Clear();
-    frames[7].Clear();
+    std::for_each(frames.begin(), frames.end(), [](auto& f) { f.Clear(); });
   }
 
   bool
@@ -162,8 +150,7 @@ namespace llarp
                                   const RouterID nextHop, const SharedSecret pathKey,
                                   uint64_t status)
   {
-    typedef std::shared_ptr< LR_StatusMessage > LRSM_ptr;
-    LRSM_ptr message = std::make_shared< LR_StatusMessage >();
+    auto message = std::make_shared< LR_StatusMessage >();
 
     message->status = status & LR_StatusRecord::SUCCESS;
     message->pathid = pathid;
@@ -233,31 +220,24 @@ namespace llarp
                                 std::shared_ptr< LR_StatusMessage > msg)
   {
     llarp::LogDebug("Attempting to send LR_Status message to (", nextHop, ")");
-    if(router->HasSessionTo(nextHop))
+    if(not router->HasSessionTo(nextHop))
     {
-      if (router->SendToOrQueue(nextHop, msg.get()))
-      {
-        return;
-      }
-      llarp::LogError("Sending LR_Status message, SendToOrQueue to ", nextHop, " failed");
+      llarp::LogError("Sending LR_Status message, but no connection to previous hop (", nextHop, ")");
       return;
     }
-    llarp::LogError("Sending LR_Status message, but no connection to previous hop (", nextHop, ")");
-    return;
+    if (not router->SendToOrQueue(nextHop, msg.get()))
+    {
+      llarp::LogError("Sending LR_Status message, SendToOrQueue to ", nextHop, " failed");
+    }
   }
 
   bool
   LR_StatusRecord::BEncode(llarp_buffer_t* buf) const
   {
-    if(!bencode_start_dict(buf))
-      return false;
-
-    if(!BEncodeWriteDictInt("s", status, buf))
-      return false;
-    if(!bencode_write_version_entry(buf))
-      return false;
-
-    return bencode_end(buf);
+    return bencode_start_dict(buf)
+        && BEncodeWriteDictInt("s", status, buf)
+        && bencode_write_version_entry(buf)
+        && bencode_end(buf);
   }
 
   bool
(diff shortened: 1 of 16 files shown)


llarp/messages/relay_status.cpp
```diff
@@ -0,0 +1,295 @@
...
+    void queue_handle()
+    {
+      auto func = std::bind(&llarp::LRSM_AsyncHandler::handle, shared_from_this());
+      llarp_threadpool_queue_job(router->pathContext().Worker(), func);
+    }
+
+  };
+
+
+  LR_StatusMessage::~LR_StatusMessage()
```
michael-loki: empty destructors aren't required.

llarp/messages/relay_status.cpp
```diff
@@ -0,0 +1,295 @@
...
+        return false;
+      }
+    }
+
+    return read;
+  }
+
+  void
+  LR_StatusMessage::Clear()
+  {
```
michael-loki:
```c++
std::for_each(frames.begin(), frames.end(), & EncryptedFrame::Clear);
```

llarp/messages/relay_status.cpp
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  // call this from a worker thread
+  bool
+  LR_StatusMessage::CreateAndSend(AbstractRouter *router, const PathID_t pathid,
+                                  const RouterID nextHop, const SharedSecret pathKey,
+                                  uint64_t status)
+  {
+    typedef std::shared_ptr< LR_StatusMessage > LRSM_ptr;
+    LRSM_ptr message = std::make_shared< LR_StatusMessage >();
```
michael-loki: remove the typedef (which should be [`using`](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using), and then:

`auto message = std::make_shared< LR_StatusMessage >();`

llarp/messages/relay_status.cpp
```diff
@@ -0,0 +1,295 @@
...
+  bool
+  LR_StatusMessage::AddFrame(const SharedSecret& pathKey, uint64_t status)
+  {
+    frames[7] = frames[6];
+    frames[6] = frames[5];
+    frames[5] = frames[4];
+    frames[4] = frames[3];
+    frames[3] = frames[2];
+    frames[2] = frames[1];
+    frames[1] = frames[0];
```
michael-loki: I think this is equivalent:

```
std::copy(frames.begin(), frames.end() - 1, frames.begin() + 1);
auto& frame = frames[0];
```

llarp/messages/relay_status.cpp
```diff
@@ -0,0 +1,295 @@
...
+    auto func = std::bind(&LR_StatusMessage::SendMessage, router, nextHop, msg); 
+    router->pathContext().logic()->queue_func(func);
+  }
+
+  void
+  LR_StatusMessage::SendMessage(AbstractRouter *router, const RouterID nextHop,
+                                std::shared_ptr< LR_StatusMessage > msg)
+  {
+    llarp::LogDebug("Attempting to send LR_Status message to (", nextHop, ")");
+    if(router->HasSessionTo(nextHop))
```
michael-loki: I would inverse the flow here:

```c++
if (not router->HasSessionTo(nextHop)) 
{
  llarp::LogError("Sending LR_Status message, but no connection to previous hop (", nextHop, ")");
  return;
} 

if (not router->SendToOrQueue(nextHop, msg.get())) 
{
  llarp::LogError("Sending LR_Status message, SendToOrQueue to ", nextHop, " failed");
  return;
}
```

llarp/messages/relay_status.cpp
```diff
@@ -0,0 +1,295 @@
...
+  {
+    if(!bencode_start_dict(buf))
+      return false;
+
+    if(!BEncodeWriteDictInt("s", status, buf))
+      return false;
+    if(!bencode_write_version_entry(buf))
+      return false;
+
+    return bencode_end(buf);
```
michael-loki: nit (since this is how most of the code does it):

```c++
return bencode_start_dict(buf) && 
    BEncodeWriteDictInt("s", status, buf) && 
    bencode_write_version_entry(buf) && 
    bencode_end(buf);
```

llarp/messages/relay_status.cpp
```diff
@@ -0,0 +1,295 @@
...
+      return false;
+
+    return read;
+  }
+
+  bool
+  LR_StatusRecord::BDecode(llarp_buffer_t* buf)
+  {
+    using namespace std::placeholders;
+    return bencode_read_dict(std::bind(&LR_StatusRecord::OnKey, this, _1, _2),
```
michael-loki:
```c++
return bencode_read_dict(util::memFn(&LR_StatusRecord::OnKey, this), buf);
```

llarp/messages/relay_status.cpp
tewinget: copied from LR_CommitMessage, should I change both?

llarp/messages/relay_status.cpp
tewinget: As above, copied from LR_CommitMessage, should I change both?

michael-loki: i think they might be fixed in #650

llarp/messages/relay_status.cpp
tewinget: With auto I may even be able to omit the `using` bit.  Will revisit.

llarp/messages/relay_status.cpp
tewinget: probably, will have to consider which reads better.  not too worried about a few extra lines.

I assume they would compile to equivalently efficient code, but that may be a poor assumption.

llarp/messages/relay_status.cpp
tewinget: fair enough, will revise

llarp/messages/relay_status.cpp
tewinget: copied (and modified) from LR_CommitMessage, should I change both?

llarp/messages/relay_status.cpp
tewinget: will do; I don't think that bit was merged yet when I was working on this and I forgot it was added/changed.

llarp/messages/relay_status.cpp
tewinget: addressed

llarp/messages/relay_status.cpp
tewinget: addressed

llarp/messages/relay_status.cpp
tewinget: addressed

llarp/messages/relay_status.cpp
tewinget: kept as-is for now, still not convinced the brevity here is worth the change in readability, but perhaps this is a result of my not having used std::copy very much.

llarp/messages/relay_status.cpp
tewinget: addressed

llarp/messages/relay_status.cpp
tewinget: addressed

llarp/messages/relay_status.cpp
tewinget: addressed

majestrate:
this will contain the failure cases when they are done too right?

tewinget:
supplanted by #732.

PR #657, opened 2019-06-19
Fix macos build system for libuv

neuroscr:


michael-loki:
I'm not sure why this is required. this is the sort of change which is required out-of-build.

majestrate:
@michael-loki was it broken on your osx ?

michael-loki:
Nope. but my `c++`  is in `/usr/local/bin`.

this is the kind of thing which should be solved with a toolchain file (I think).

neuroscr:
> I'm not sure why this is required. 

because macos by default doesn't search /usr/local for the dependencies. This fixes being able to find and link against libuv

>this is the sort of change which is required out-of-build.

what do you mean by that?

> Nope. but my `c++` is in `/usr/local/bin`.

my c++ is in `/usr/bin`

>this is the kind of thing which should be solved with a toolchain file (I think).

I just resolved it how Jeff resolved the FreeBSD implementation, if you don't like the pattern talk to him.

majestrate:
I think the right way is probably a custom cmake toolchain file for your setup.

neuroscr:
Can someone give me an example or pattern to copy?

michael-loki:
I'll make the changes in #668

michael-loki:
@neuroscr see https://github.com/loki-project/loki-network/pull/668/commits/07377b8d5f371eeffe9ce7aa8898c00f33e8574f

and then when you invoke `cmake`, pass in `-DCMAKE_LIBRARY_PATH=/usr/local/lib`

PR #660, opened 2019-06-20
Tidy up pathbuilder code

michael-loki:
Pathbuilding doesn't (directly) depend on the DHT.


majestrate:
you sure that change with the keygen stuff will work?

michael-loki:
It was working for me on bunnynet

PR #662, opened 2019-06-21
fix up libuv close logic

majestrate:


llarp/ev/ev_libuv.cpp
```diff
@@ -574,8 +620,34 @@ namespace libuv
...
+            static_cast< glue* >(h->data)->Close();
+          }
+        },
+        nullptr);
+  }
+
+  void
+  Loop::stopped()
+  {
+    tick(50);
```
michael-loki: Why 50?

llarp/ev/ev_libuv.cpp
```diff
@@ -3,12 +3,19 @@
 
 namespace libuv
 {
+  struct glue
+  {
+    virtual ~glue(){};
```
michael-loki: `= default;`

llarp/ev/ev_libuv.cpp
majestrate: arbitrary

PR #668, opened 2019-06-27
Remove cppbackport from vendor dir

michael-loki:
> 6 additions and 5,019 deletions


jagerman:
Isn't this going to result in compilation problems with non-cutting-edge compilers?

majestrate:
debian stable is okay with it so that's a reasonable cutoff.

majestrate:
we require c++14 features anyways

majestrate:
@jagerman is this acceptable merge given debian stable handles it?

neuroscr:
what about macos and windows?

despair86:
Filesystem TS is a C++17 feature, `CMAKE_CXX_STANDARD` should be bumped up

looks fine to me, seeing as this never worked on windows anyway -rick

michael-loki:
most compilers have been shipping it under `experimental` for a while, so we don't need to bump to 17.

PR #669, opened 2019-06-28
Refactor config

michael-loki:
This builds on #668 because I needed a working build 😉


llarp/net/net_addr.cpp
```diff
@@ -81,25 +81,17 @@ namespace llarp
   }
 
   bool
-  Addr::from_char_array(const char* in)
+  Addr::from_char_array(string_view in)
```
michael-loki: this could do with a review.

majestrate:
this needs to be updated to reflect the fs stuff

michael-loki:
ack

michael-loki:
@majestrate done.

PR #674, opened 2019-07-01
close after 30 seconds of dead

majestrate:


llarp/utp/session.cpp
```diff
@@ -189,16 +189,13 @@ namespace llarp
...
-      {
-        if(now <= lastSend)
-          return false;
-        return now - lastSend > 5000;
-      }
-      // let utp manage this
-      return state == eClose;
+      if(now <= lastSend)
+        return false;
+      return now - lastSend > 30000;
```
neuroscr: what about receives? are those last active?

PR #687, opened 2019-07-08
Allow environment variable override of config

michael-loki:


llarp/router/router.cpp
```diff
@@ -770,9 +770,9 @@ namespace llarp
   Router::fromConfig(Config *conf)
   {
     // Set netid before anything else
-    if(!conf->router.netid.empty())
+    if(!conf->router.netId().empty())
     {
-      const auto &netid = conf->router.netid;
+      const auto &netid = conf->router.netId();
```
majestrate: should this be `const auto` instead of `const auto &` ?

michael-loki: Either is fine: https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/

PR #686, opened 2019-07-08
merge IPv6 support branch and gutt use of "auto" in config

majestrate:
this merges my dev branch so that i don't have to worry about additional divergence.

PR #690, opened 2019-07-09
Remove redundant else blocks

michael-loki:
might conflict with #686, this can just be applied afterwards


llarp/exit/endpoint.cpp
```diff
@@ -98,7 +98,7 @@ namespace llarp
       auto lastPing = path->LastRemoteActivityAt();
       if(lastPing == 0 || (now > lastPing && now - lastPing > timeout))
         return now > m_LastActive && now - m_LastActive > timeout;
-      else if(lastPing)
+      if(lastPing)
```
majestrate: this part here is needed because of the other return path

michael-loki: which other return path? they’re functionally equivalent afaict.

michael-loki: Corrected either way.

Issue #689, opened 2019-07-09
Windows implementation of SetThreadName


michael-loki:
It’s done on other platforms, just not windows.

despair86:
I have one already
I'll pull it back out, it's wrapped in `#if 0` block atm

despair86:
and in fact, the only reason you'd need one is for MSVC builds as `pthread_setname_np(3)` is available under the MSYS2 build environment

despair86:
fixed in my branch (see 19571cce1e29917d5a119f13ce263e3c55414376)

PR #691, opened 2019-07-10
fix freebsd build

majestrate:
this SHOULD fix the freebsd build that was broken by #689

### DIFF
--- a/llarp/util/threading.cpp
+++ b/llarp/util/threading.cpp
@@ -21,15 +21,16 @@ namespace llarp
       const int rc = pthread_setname_np(name.c_str());
 #else
 #ifdef __FreeBSD__
-      pthread_setname_np(pthread_self(), name.c_str());
+      const int rc = pthread_setname_np(pthread_self(), name.c_str(), nullptr);
 #else
       const int rc = pthread_setname_np(pthread_self(), name.c_str());
+#endif
       if(rc)
       {
         LogError("Failed to set thread name to ", name, " errno = ", rc,
                  " errstr = ", strerror(rc));
       }
-#endif
+
 #endif
 #else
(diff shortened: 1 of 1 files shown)


llarp/util/threading.cpp
```diff
@@ -16,14 +19,18 @@ namespace llarp
 #ifdef POSIX
 #ifdef __MACH__
       const int rc = pthread_setname_np(name.c_str());
+#else
+#ifdef __FreeBSD__
+      pthread_setname_np(pthread_self(), name.c_str());
```
michael-loki: based on https://www.freebsd.org/cgi/man.cgi?query=pthread_setname_np&sektion=3&apropos=0&manpath=OSF1+V5.1%2falpha

this should be:

```c++
const int rc = pthread_setname_np(pthread_self(), name.c_str(), nullptr);
```

llarp/util/threading.cpp
```diff
@@ -24,6 +30,8 @@ namespace llarp
         LogError("Failed to set thread name to ", name, " errno = ", rc,
                  " errstr = ", strerror(rc));
       }
+
+#endif
```
neuroscr: is this a stray endif? why was it aadded?

michael-loki: because of the `#else` on line 22.

llarp/util/threading.cpp
```diff
@@ -16,6 +19,9 @@ namespace llarp
 #ifdef POSIX
 #ifdef __MACH__
       const int rc = pthread_setname_np(name.c_str());
+#else
+#ifdef __FreeBSD__
```
michael-loki: `#elif defined(__FreeBSD__)`, and then that addresses @neuroscr’s comment

PR #705, opened 2019-07-13
Matrix build for travis

michael-loki:


.travis.yml
```diff
@@ -1,58 +1,44 @@
...
+  - STATIC_LINK=OFF NINJA=ninja NETNS=ON
+  - STATIC_LINK=ON NINJA=ninja NETNS=ON
+  - STATIC_LINK=OFF SHARED_LIB=ON
+  - STATIC_LINK=ON SHARED_LIB=ON
+  - STATIC_LINK=OFF NINJA=ninja SHARED_LIB=ON
+  - STATIC_LINK=ON NINJA=ninja SHARED_LIB=ON
+  - STATIC_LINK=OFF NETNS=ON SHARED_LIB=ON
+  - STATIC_LINK=ON NETNS=ON SHARED_LIB=ON
+  - STATIC_LINK=OFF NINJA=ninja NETNS=ON SHARED_LIB=ON
+  - STATIC_LINK=ON NINJA=ninja NETNS=ON SHARED_LIB=ON
```
michael-loki: do these combinations really make sense? we probably don’t need to test with/without ninja etc

neuroscr:
You removed:
- having a unified travis-ci testing framework (from bitcoincore inherited through Monero) across all loki projects (but I'm not sure any one desires this beyond me)
- builder and host agnostism (and testing cross platform building)
- No reliance on timestamps (previous macos sdk image has specifically modified build tools, we're working on updating this image to be newer, you could have helped the loki project too if you had worked on that)
- The previous system can be made to make builds deterministic but we haven't gotten around to making the dependency package files

.travis.yml
neuroscr: Probably right, I don't think it's support breaks too often (I'm not 100% it ever has but memory think either it did or something adjacent to it did) but would be nice to know when it does (if that's possible)

michael-loki:
Copy paste is not unification.

michael-loki:
> having a unified travis-ci testing framework (from bitcoincore inherited through Monero) across all loki projects (but I'm not sure any one desires this beyond me)

How is it unified? beyond sharing some (unused!) environment variables, what value does it add?

> builder and host agnostism (and testing cross platform building)

sort of goes against what travis provides. cross platform build *could* be part of the matrix build, and I want to add *actual* macOS and windows builds here, so we don’t regress on those platforms.

> No reliance on timestamps

I don’t see how the current build does that either

> The previous system can be made to make builds deterministic but we haven't gotten around to making the dependency package files

So could this one.

neuroscr:
Part of me really likes ditching that tightly wound system, so I don't deeply care one way or another. Your expertise here is appreciated but I did want to communicate what the advantages of the existing system over this (because it was a bit of a rabbit hole for me to figure out and clearly not obvious).

>what value does [consistency across project] add?

The idea would be having more staff that can pitch in and work cross-project, so we have less dependencies on specific people.

>So could this one.

Agreed, anything can be improved. Travis only having Xenial is painful though.

neuroscr:
Also thought mac/win was running on gitlab-ci but looks like that's all disabled. Even had to remove alpine, so if you have any time to look at that one too, it's also very easy to get macos going over there (The office has a mac runner).

And with our windows build VM, we can set up a windows runner too.

michael-loki:
It looks like this branch (awkwardly) isn’t triggering CI anymore. Will take a look later today.

PR #715, opened 2019-07-17
current windows fixes

### LINKED ISSUE #689: Windows implementation of SetThreadName

despair86:
not quite ready to merge since i'm waiting for all the other huge changes to land first

so far we have: 

- fix #689 
- fix most automated builds, including LTO and cross-compiled LTO
- core dump handler allows end-users to send us dumped core images to aid debugging on windows (WIP)


majestrate:
can we not commit more binary assets to the repo? thnx.

despair86:
@majestrate need debug helper for core dump generation on Windows, do i just serve that up JIT from a download point?

majestrate:
why does it need to be in the repo?

despair86:
Right now, it is packaged straight into the installer, just like the other libraries there. Unlike 7-zip and the registry helper lib, this one is sadly closed source...

(`libdbghelp` is a redistributable library and one should not depend on the target system even _having_ its own copy in the system folders to begin with)

majestrate:
merge conflicts

despair86:
yeah I'll fix those in the afternoon 👌🏻✅👍🏻

despair86:
backed out all the cross-LTO stuff
travis CI integration broke it all

despair86:
woooo
it works

despair86:
@majestrate merge at will, might be blocking #726

despair86:
my normal rebase broke this time
git for windows treats shell scripts and binaries with the same name equally 😢

cmake/win32.cmake
```diff
@@ -18,8 +18,10 @@ if(NOT MSVC_VERSION)
   # to .r[o]data section one after the other!
   add_compile_options(-fno-ident -Wa,-mbig-obj)
   link_libraries( -lshlwapi -ldbghelp )
-  add_definitions(-DWINVER=0x0500 -D_WIN32_WINNT=0x0500)
-  # set(FS_LIB stdc++fs)
+  add_definitions(-DWINVER=0x0500 -D_WIN32_WINNT=0x0500)
+  if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 9.0)
```
michael-loki: Try this instead - it's more portable to things like clang on macOS:

```cmake
find_library(FS_LIB_NAME NAMES stdc++fs)
if(NOT FS_LIB_NAME STREQUAL FS_LIB_NAME-NOTFOUND)
  set(FS_LIB ${FS_LIB_NAME})
endif()
```

llarp/win32/win32_intrnl.c
```diff
@@ -1,572 +1,676 @@
...
+{
+  return getInterfaceIndexTableInt(FALSE);
+}
+
+#endif
+
+// there's probably an use case for a _newer_ implementation
+// of pthread_setname_np(3), in fact, I may just merge _this_
+// upstream...
+#ifdef _MSC_VER
```
michael-loki: looks like the line endings changed in this file, was that intentional?

despair86: yeah, i was using visual studio, then git changed it to UNIX line ends

PR #716, opened 2019-07-17
reduce logging

majestrate:
reduce verbosity of default logging


michael-loki:
Ready for review yet?

michael-loki:
not sure why that test failed in release. should be good to merge.

PR #719, opened 2019-07-18
only call utp_issue_deferred_acks before timing out sessions

majestrate:
this helps prevent some unneeded timeouts on utp layer.

### DIFF
--- a/llarp/service/sendcontext.cpp
+++ b/llarp/service/sendcontext.cpp
@@ -57,6 +57,7 @@ namespace llarp
     {
       SharedSecret shared;
       ProtocolFrame f;
+      f.C.Zero();
       f.N.Randomize();
       f.T = currentConvoTag;
       f.S = ++sequenceNo;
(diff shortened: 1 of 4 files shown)


llarp/service/sendcontext.cpp
```diff
@@ -57,6 +57,7 @@ namespace llarp
     {
       SharedSecret shared;
       ProtocolFrame f;
+      f.C.Zero();
```
michael-loki: Move this to `ProtocolFrame`‘s constructor instead?

llarp/utp/session.hpp
```diff
@@ -33,9 +33,9 @@ namespace llarp
       /// tx session key
       SharedSecret txKey;
       /// timestamp last active
-      llarp_time_t lastActive;
+      llarp_time_t lastActive = 0;
```
michael-loki: 👍

llarp/service/sendcontext.cpp
majestrate: nvm it's already in it using the `Clear()` function

PR #720, opened 2019-07-18
Make hop length and quantity configurable

majestrate:
builds on #719


llarp/path/pathset.hpp
```diff
@@ -269,9 +271,9 @@ namespace llarp
         }
       }
 
-     protected:
-      size_t m_NumPaths;
+      size_t numPaths;
```
michael-loki: why rename this? It would have been cleaner to just keep it as `m_numPaths`.

PR #725, opened 2019-07-20
Basic docs translations to spanish, more to go later

rjmalagon:
From time to time I will translate this dirctory to spanish.
I would use a pseudo academic and formal enough style with some literal and some cultural translation.
Strict spellcheck.


majestrate:
can we use `docs/es/` as the base folder?

PR #728, opened 2019-07-22
Refactor router code

tewinget:
This PR refactors a lot of the functionality that is "kitchen sink"ed into router.cpp into separate classes.

This PR will be squashed before merge.

There are also some minor behavior changes along with this refactor as I came across behavior that was either incorrect or sub-optimal.

### DIFF
--- a/llarp/link/i_link_manager.hpp
+++ b/llarp/link/i_link_manager.hpp
@@ -1,0 +1,85 @@
+#ifndef LLARP_I_LINK_MANAGER_HPP
+#define LLARP_I_LINK_MANAGER_HPP
+
+#include <link/server.hpp>
+#include <util/types.hpp>
+#include <util/logic.hpp>
+
+#include <functional>
+
+struct llarp_buffer_t;
+
+namespace llarp
+{
+  using Logic_ptr = std::shared_ptr< Logic >;
+
+  struct RouterContact;
+  struct ILinkSession;
+  struct IOutboundSessionMaker;
+  struct RouterID;
+
+  namespace util
+  {
+    struct StatusObject;
+  }  // namespace util
+
+  struct ILinkManager
+  {
+    virtual ~ILinkManager() = default;
+
+    virtual LinkLayer_ptr
+    GetCompatibleLink(const RouterContact &rc) const = 0;
+
+    virtual IOutboundSessionMaker *
+    GetSessionMaker() const = 0;
+
+    virtual bool
+    SendTo(const RouterID &remote, const llarp_buffer_t &buf) = 0;
+
+    virtual bool
+    HasSessionTo(const RouterID &remote) const = 0;
+
+    virtual void
+    PumpLinks() = 0;
+
+    virtual void
+    AddLink(LinkLayer_ptr link, bool inbound = false) = 0;
+
+    virtual bool
+    StartLinks(Logic_ptr logic) = 0;
+
+    virtual void
+    Stop() = 0;
+
+    virtual void
+    PersistSessionUntil(const RouterID &remote, llarp_time_t until) = 0;
+
+    virtual void
+    ForEachPeer(std::function< void(const ILinkSession *, bool) > visit,
+                bool randomize = false) const = 0;
+
+    virtual void
+    ForEachPeer(std::function< void(ILinkSession *) > visit) = 0;
+
+    virtual void
+    ForEachInboundLink(std::function< void(LinkLayer_ptr) > visit) const = 0;
+
+    virtual size_t
+    NumberOfConnectedRouters() const = 0;
+
+    virtual size_t
+    NumberOfConnectedClients() const = 0;
+
+    virtual bool
+    GetRandomConnectedRouter(RouterContact &router) const = 0;
+
+    virtual void
+    CheckPersistingSessions(llarp_time_t now) = 0;
+
+    virtual util::StatusObject
+    ExtractStatus() const = 0;
+  };
+
+}  // namespace llarp
+
+#endif  // LLARP_I_LINK_MANAGER_HPP
--- a/llarp/link/link_manager.cpp
+++ b/llarp/link/link_manager.cpp
@@ -1,0 +1,362 @@
+#include <link/link_manager.hpp>
+
+#include <router/i_outbound_session_maker.hpp>
+#include <crypto/crypto.hpp>
+
+#include <algorithm>
+#include <set>
+
+namespace llarp
+{
+  LinkLayer_ptr
+  LinkManager::GetCompatibleLink(const RouterContact &rc) const
+  {
+    if(stopping)
+      return nullptr;
+
+    for(auto &link : outboundLinks)
+    {
+      // TODO: may want to add some memory of session failures for a given
+      //      router on a given link and not return that link here for a
+      //      duration
+      if(!link->IsCompatable(rc))
+        continue;
+
+      return link;
+    }
+
+    return nullptr;
+  }
+
+  IOutboundSessionMaker *
+  LinkManager::GetSessionMaker() const
+  {
+    return _sessionMaker;
+  }
+
+  bool
+  LinkManager::SendTo(const RouterID &remote, const llarp_buffer_t &buf)
+  {
+    if(stopping)
+      return false;
+
+    auto link = GetLinkWithSessionTo(remote);
+    if(link == nullptr)
+    {
+      return false;
+    }
+
+    return link->SendTo(remote, buf);
+  }
+
+  bool
+  LinkManager::HasSessionTo(const RouterID &remote) const
+  {
+    return GetLinkWithSessionTo(remote) != nullptr;
+  }
+
+  void
+  LinkManager::PumpLinks()
+  {
+    for(const auto &link : inboundLinks)
+    {
+      link->Pump();
+    }
+    for(const auto &link : outboundLinks)
+    {
+      link->Pump();
+    }
+  }
+
+  void
+  LinkManager::AddLink(LinkLayer_ptr link, bool inbound)
+  {
+    util::Lock l(&_mutex);
+
+    if(inbound)
+    {
+      inboundLinks.emplace(link);
+    }
+    else
+    {
+      outboundLinks.emplace(link);
+    }
+  }
+
+  bool
+  LinkManager::StartLinks(Logic_ptr logic)
+  {
+    LogInfo("starting ", outboundLinks.size(), " outbound links");
+    for(const auto &link : outboundLinks)
+    {
+      if(!link->Start(logic))
+      {
+        LogWarn("outbound link '", link->Name(), "' failed to start");
+        return false;
+      }
+      LogDebug("Outbound Link ", link->Name(), " started");
+    }
+
+    if(inboundLinks.size())
+    {
+      LogInfo("starting ", inboundLinks.size(), " inbound links");
+      for(const auto &link : inboundLinks)
+      {
+        if(!link->Start(logic))
+        {
+          LogWarn("Link ", link->Name(), " failed to start");
+          return false;
+        }
+        LogDebug("Inbound Link ", link->Name(), " started");
+      }
+    }
+
+    return true;
+  }
+
+  void
+  LinkManager::Stop()
+  {
+    if(stopping)
+    {
+      return;
+    }
+
+    util::Lock l(&_mutex);
+
+    LogInfo("stopping links");
+    stopping = true;
+
+    for(const auto &link : outboundLinks)
+      link->Stop();
+    for(const auto &link : inboundLinks)
+      link->Stop();
+  }
+
+  void
+  LinkManager::PersistSessionUntil(const RouterID &remote, llarp_time_t until)
+  {
+    if(stopping)
+      return;
+
+    util::Lock l(&_mutex);
+
+    m_PersistingSessions[remote] =
+        std::max(until, m_PersistingSessions[remote]);
+    LogDebug("persist session to ", remote, " until ",
+             m_PersistingSessions[remote]);
+  }
+
+  void
+  LinkManager::ForEachPeer(
+      std::function< void(const ILinkSession *, bool) > visit,
+      bool randomize) const
+  {
+    if(stopping)
+      return;
+
+    for(const auto &link : outboundLinks)
+    {
+      link->ForEachSession(
+          [visit](const ILinkSession *peer) { visit(peer, true); }, randomize);
+    }
+    for(const auto &link : inboundLinks)
+    {
+      link->ForEachSession(
+          [visit](const ILinkSession *peer) { visit(peer, false); }, randomize);
+    }
+  }
+
+  void
+  LinkManager::ForEachPeer(std::function< void(ILinkSession *) > visit)
+  {
+    if(stopping)
+      return;
+
+    for(const auto &link : outboundLinks)
+    {
+      link->ForEachSession([visit](ILinkSession *peer) { visit(peer); });
+    }
+    for(const auto &link : inboundLinks)
+    {
+      link->ForEachSession([visit](ILinkSession *peer) { visit(peer); });
+    }
+  }
+
+  void
+  LinkManager::ForEachInboundLink(
+      std::function< void(LinkLayer_ptr) > visit) const
+  {
+    for(const auto &link : inboundLinks)
+    {
+      visit(link);
+    }
+  }
+
+  size_t
+  LinkManager::NumberOfConnectedRouters() const
+  {
+    std::set< RouterID > connectedRouters;
+
+    auto fn = [&connectedRouters](const ILinkSession *session, bool) {
+      if(session->IsEstablished())
+      {
+        const RouterContact rc(session->GetRemoteRC());
+        if(rc.IsPublicRouter())
+        {
+          connectedRouters.insert(rc.pubkey);
+        }
+      }
+    };
+
+    ForEachPeer(fn);
+
+    return connectedRouters.size();
+  }
+
+  size_t
+  LinkManager::NumberOfConnectedClients() const
+  {
+    std::set< RouterID > connectedClients;
+
+    auto fn = [&connectedClients](const ILinkSession *session, bool) {
+      if(session->IsEstablished())
+      {
+        const RouterContact rc(session->GetRemoteRC());
+        if(!rc.IsPublicRouter())
+        {
+          connectedClients.insert(rc.pubkey);
+        }
+      }
+    };
+
+    ForEachPeer(fn);
+
+    return connectedClients.size();
+  }
+
+  bool
+  LinkManager::GetRandomConnectedRouter(RouterContact &router) const
+  {
+    std::unordered_map< RouterID, RouterContact, RouterID::Hash >
+        connectedRouters;
+
+    ForEachPeer(
+        [&connectedRouters](const ILinkSession *peer, bool unused) {
+          (void)unused;
+          connectedRouters[peer->GetPubKey()] = peer->GetRemoteRC();
+        },
+        false);
+
+    const auto sz = connectedRouters.size();
+    if(sz)
+    {
+      auto itr = connectedRouters.begin();
+      if(sz > 1)
+      {
+        std::advance(itr, randint() % sz);
+      }
+
+      router = itr->second;
+
+      return true;
+    }
+
+    return false;
+  }
+
+  void
+  LinkManager::CheckPersistingSessions(llarp_time_t now)
+  {
+    if(stopping)
+      return;
+
+    std::vector< RouterID > sessionsNeeded;
+
+    {
+      util::Lock l(&_mutex);
+
+      auto itr = m_PersistingSessions.begin();
+      while(itr != m_PersistingSessions.end())
+      {
+        auto link = GetLinkWithSessionTo(itr->first);
+        if(now < itr->second)
+        {
+          if(link)
+          {
+            LogDebug("keepalive to ", itr->first);
+            link->KeepAliveSessionTo(itr->first);
+          }
+          else
+          {
+            sessionsNeeded.push_back(itr->first);
+          }
+          ++itr;
+        }
+        else
+        {
+          const RouterID r(itr->first);
+          LogInfo("commit to ", r, " expired");
+          itr = m_PersistingSessions.erase(itr);
+        }
+      }
+    }
+
+    for(const auto &router : sessionsNeeded)
+    {
+      _sessionMaker->CreateSessionTo(router, nullptr);
+    }
+  }
+
+  util::StatusObject
+  LinkManager::ExtractStatus() const
+  {
+    std::vector< util::StatusObject > ob_links, ib_links;
+    std::transform(inboundLinks.begin(), inboundLinks.end(),
+                   std::back_inserter(ib_links),
+                   [](const auto &link) -> util::StatusObject {
+                     return link->ExtractStatus();
+                   });
+    std::transform(outboundLinks.begin(), outboundLinks.end(),
+                   std::back_inserter(ob_links),
+                   [](const auto &link) -> util::StatusObject {
+                     return link->ExtractStatus();
+                   });
+
+    util::StatusObject obj{{"outbound", ob_links}, {"inbound", ib_links}};
+
+    return obj;
+  }
+
+  void
+  LinkManager::Init(IOutboundSessionMaker *sessionMaker)
+  {
+    stopping      = false;
+    _sessionMaker = sessionMaker;
+  }
+
+  LinkLayer_ptr
+  LinkManager::GetLinkWithSessionTo(const RouterID &remote) const
+  {
+    if(stopping)
+      return nullptr;
+
+    for(const auto &link : inboundLinks)
+    {
+      if(link->HasSessionTo(remote))
+      {
+        return link;
+      }
+    }
+    for(const auto &link : outboundLinks)
+    {
+      if(link->HasSessionTo(remote))
+      {
+        return link;
+      }
+    }
+
+    return nullptr;
+  }
+
+}  // namespace llarp
--- a/llarp/link/link_manager.hpp
+++ b/llarp/link/link_manager.hpp
@@ -1,0 +1,100 @@
+#ifndef LLARP_LINK_MANAGER_HPP
+#define LLARP_LINK_MANAGER_HPP
+
+#include <link/i_link_manager.hpp>
+
+#include <util/compare_ptr.hpp>
+#include <util/threading.hpp>
+#include <link/server.hpp>
+
+#include <unordered_map>
+#include <set>
+#include <atomic>
+
+namespace llarp
+{
+  struct IRouterContactManager;
+
+  struct LinkManager final : public ILinkManager
+  {
+   public:
+    ~LinkManager() = default;
+
+    LinkLayer_ptr
+    GetCompatibleLink(const RouterContact &rc) const override;
+
+    IOutboundSessionMaker *
+    GetSessionMaker() const override;
+
+    bool
+    SendTo(const RouterID &remote, const llarp_buffer_t &buf) override;
+
+    bool
+    HasSessionTo(const RouterID &remote) const override;
+
+    void
+    PumpLinks() override;
+
+    void
+    AddLink(LinkLayer_ptr link, bool inbound = false) override;
+
+    bool
+    StartLinks(Logic_ptr logic) override;
+
+    void
+    Stop() override;
+
+    void
+    PersistSessionUntil(const RouterID &remote, llarp_time_t until) override;
+
+    void
+    ForEachPeer(std::function< void(const ILinkSession *, bool) > visit,
+                bool randomize = false) const override;
+
+    void
+    ForEachPeer(std::function< void(ILinkSession *) > visit) override;
+
+    void
+    ForEachInboundLink(
+        std::function< void(LinkLayer_ptr) > visit) const override;
+
+    size_t
+    NumberOfConnectedRouters() const override;
+
+    size_t
+    NumberOfConnectedClients() const override;
+
+    bool
+    GetRandomConnectedRouter(RouterContact &router) const override;
+
+    void
+    CheckPersistingSessions(llarp_time_t now) override;
+
+    virtual util::StatusObject
+    ExtractStatus() const override;
+
+    void
+    Init(IOutboundSessionMaker *sessionMaker);
+
+   private:
+    LinkLayer_ptr
+    GetLinkWithSessionTo(const RouterID &remote) const;
+
+    std::atomic< bool > stopping;
+    mutable util::Mutex _mutex;  // protects m_PersistingSessions
+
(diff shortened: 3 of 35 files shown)


llarp/link/i_link_manager.hpp
```diff
@@ -0,0 +1,85 @@
+#ifndef LLARP_I_LINK_MANAGER_HPP
```
michael-loki: fwiw the convention I've tried to stick with is `AbstractX` (e.g. `AbstractRouter`)

michael-loki: also introduce a `.cpp` to go with this, which just `#includes` this file. this just verifies this header file compiles in a standalone fashion.

llarp/link/link_manager.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+namespace llarp
+{
+
+  LinkLayer_ptr
+  LinkManager::GetCompatibleLink(const RouterContact &rc) const
+  {
+    {
+      util::Lock l(&_mutex);
+      if (stopping) return nullptr;
```
michael-loki: you don't need to lock to check an atomic

llarp/link/link_manager.cpp
```diff
@@ -0,0 +1,353 @@
...
+  LinkManager::GetSessionMaker() const
+  {
+    return _sessionMaker;
+  }
+
+  bool
+  LinkManager::SendTo(const RouterID& remote, const llarp_buffer_t& buf)
+  {
+    {
+      util::Lock l(&_mutex);
```
michael-loki: lock again

llarp/link/link_manager.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+  void
+  LinkManager::Stop()
+  {
+    util::Lock l(&_mutex);
+
+    if (!stopping)
+    {
+      LogInfo("stopping links");
+      stopping.store(true);
```
michael-loki: `stopping = true;`

llarp/link/link_manager.cpp
```diff
@@ -0,0 +1,353 @@
...
+      }
+    }
+
+    return true;
+  }
+
+  void
+  LinkManager::Stop()
+  {
+    util::Lock l(&_mutex);
```
michael-loki: don't need the lock here

llarp/link/link_manager.cpp
```diff
@@ -0,0 +1,353 @@
...
+        m_PersistingSessions[remote]);
+  }
+
+  void
+  LinkManager::ForEachPeer(std::function< void(const ILinkSession *, bool) > visit,
+                           bool randomize) const
+  {
+    {
+      util::Lock l(&_mutex);
+      if (stopping) return;
```
michael-loki: and here

llarp/link/link_manager.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <unordered_map>
+#include <set>
+#include <atomic>
+
+namespace llarp
+{
+
+  struct IRouterContactManager;
+
+  struct LinkManager : public ILinkManager
```
michael-loki: `final` if this is final

llarp/link/link_manager.hpp
```diff
@@ -0,0 +1,102 @@
...
+    mutable util::Mutex _mutex; // protects m_PersistingSessions
+
+    using LinkSet = std::set< LinkLayer_ptr, ComparePtr< LinkLayer_ptr > >;
+
+    LinkSet outboundLinks;
+    LinkSet inboundLinks;
+
+    // sessions to persist -> timestamp to end persist at
+    std::unordered_map< RouterID, llarp_time_t, RouterID::Hash >
+        m_PersistingSessions GUARDED_BY(_mutex);
```
michael-loki: ❤️

llarp/router/i_outbound_message_handler.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef LLARP_ROUTER_I_OUTBOUND_MESSAGE_HANDLER_HPP
```
michael-loki: abstract instead?

llarp/router/outbound_message_handler.cpp
```diff
@@ -0,0 +1,209 @@
...
+  {
+    std::array< byte_t, MAX_LINK_MSG_SIZE > linkmsg_buffer;
+    llarp_buffer_t buf(linkmsg_buffer);
+
+    if (!EncodeBuffer(msg, buf))
+    {
+      return false;
+    }
+
+    Message message = std::make_pair(buf.sz, callback);
```
michael-loki: just go with `Message message{buf.sz, callback};`

llarp/router/outbound_message_handler.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    memcpy(message.first.data(), buf.base, buf.sz);
```
michael-loki: `std::copy_n` instead?

llarp/router/outbound_message_handler.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    return true;
+  }
+
+  //TODO: this
+  util::StatusObject
+  OutboundMessageHandler::ExtractStatus() const
+  {
+    util::StatusObject status;
+    return status;
```
michael-loki: `return {};` instead?

llarp/router/outbound_message_handler.cpp
```diff
@@ -0,0 +1,209 @@
...
+  bool
+  OutboundMessageHandler::Send(const RouterID& remote,
+                               const Message& msg)
+  {
+    llarp_buffer_t buf(msg.first);
+    if (_linkManager->SendTo(remote, buf))
+    {
+      DoCallback(msg.second, SendStatus::Success);
+    }
+    DoCallback(msg.second, SendStatus::Congestion);
```
michael-loki: is this what you mean? we always call the congestion func rn

llarp/router/outbound_message_handler.hpp
```diff
@@ -0,0 +1,99 @@
...
+struct llarp_buffer_t;
+
+namespace llarp
+{
+
+  struct LinkManager;
+  struct Logic;
+  enum class SessionResult;
+
+  struct OutboundMessageHandler : public IOutboundMessageHandler
```
michael-loki: `final`?

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,293 @@
...
+  bool
+  OutboundSessionMaker::OnSessionEstablished(ILinkSession *session)
+  {
+    //TODO: do we want to keep it
+
+    const auto &router = session->GetPubKey();
+
+    LogInfo("session with ", router, " established");
+
+    if (not _rcLookup->RemoteIsAllowed(router))
```
michael-loki: I actually prefer the `not` token over `!`

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,293 @@
...
+    {
+      FinalizeRequest(router, SessionResult::InvalidRouter);
+    }
+
+    auto func = std::bind(&OutboundSessionMaker::VerifyRC,
+                          this,
+                          session->GetRemoteRC());
+    _threadpool->addJob(func);
```
michael-loki: fwiw, I think naming `func` detracts from the semantics of the call.

```c++
_threadpool->addJob(std::bind(&OutboundSessionMaker::VerifyRC, this, session->GetRemoteRC()));
```

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,293 @@
...
+  //TODO: this
+  util::StatusObject
+  OutboundSessionMaker::ExtractStatus() const
+  {
+    util::StatusObject status;
+    return status;
+  }
+
+  void
+  Init(LinkManager *linkManager,
```
michael-loki: I hate init methods. why not a constructor which does this?

llarp/router/outbound_session_maker.cpp
jagerman: Does MSVC support it yet (without `#include <ciso646>`)?

michael-loki: Actually good question. I'm not sure. you just need `iso646.h` (since c99) but msvc is weird and is a c++ exclusive compiler 🤷‍♀

llarp/router/rc_lookup_handler.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+  void
+  RCLookupHandler::PeriodicUpdate(llarp_time_t now)
+  {
+    // try looking up stale routers
+    _nodedb->VisitInsertedBefore(
+        [&](const RouterContact &rc) {
+          if(HavePendingLookup(rc.pubkey))
+            return;
+          GetRC(rc.pubkey, nullptr);
```
majestrate: GetRC calls stuff that acquires a lock that is held while visiting.

llarp/router/rc_lookup_handler.cpp
```diff
@@ -0,0 +1,323 @@
...
+      if(!s->IsEstablished())
+        return;
+      const RouterContact rc = s->GetRemoteRC();
+      if(rc.IsPublicRouter())
+        _dht->impl->ExploreNetworkVia(dht::Key_t{rc.pubkey});
+    });
+  }
+
+  void
+  RCLookupHandler::Init(llarp_dht_context *dht,
```
majestrate: this should be the constructor.

llarp/link/link_manager.cpp
```diff
@@ -0,0 +1,359 @@
...
+                     return link->ExtractStatus();
+                   });
+
+    util::StatusObject obj{{"outbound", ob_links}, {"inbound", ib_links}};
+
+    return obj;
+  }
+
+  void
+  LinkManager::Init(IOutboundSessionMaker *sessionMaker)
```
majestrate: this should ALSO be the constructor.

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,299 @@
...
+  //TODO: this
+  util::StatusObject
+  OutboundSessionMaker::ExtractStatus() const
+  {
+    util::StatusObject status{};
+    return status;
+  }
+
+  void
+  OutboundSessionMaker::Init(ILinkManager *linkManager,
```
majestrate: also this should be a constructor. :>

llarp/router/outbound_message_handler.cpp
```diff
@@ -0,0 +1,217 @@
...
+  //TODO: this
+  util::StatusObject
+  OutboundMessageHandler::ExtractStatus() const
+  {
+    util::StatusObject status{};
+    return status;
+  }
+
+  void
+  OutboundMessageHandler::Init(ILinkManager* linkManager,
```
majestrate: also a constructor

llarp/link/link_manager.cpp
```diff
@@ -0,0 +1,359 @@
...
+    ForEachPeer( [&connectedRouters]
+                     (const ILinkSession *peer, bool unused)
+                     {
+                       (void)unused;
+                       connectedRouters[peer->GetPubKey()] = peer->GetRemoteRC();
+                     },
+                 false
+               );
+
+    auto sz = connectedRouters.size();
```
majestrate: this should be `const`

llarp/link/link_manager.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+    IOutboundSessionMaker *_sessionMaker;
```
majestrate: should this be a `const *` ?

llarp/router/outbound_message_handler.cpp
```diff
@@ -0,0 +1,217 @@
...
+      case SessionResult::RouterNotFound:
+        OnRouterNotFound(router);
+        break;
+      case SessionResult::InvalidRouter:
+        OnInvalidRouter(router);
+        break;
+      case SessionResult::NoLink:
+        OnNoLink(router);
+        break;
+      default:
```
majestrate: this  should probably error on default case "just in case"

llarp/router/outbound_message_handler.hpp
```diff
@@ -0,0 +1,100 @@
...
+      void
+      FinalizeRequest(const RouterID &router,
+                      SendStatus status);
+
+      mutable util::Mutex _mutex; // protects outboundMessageQueue
+
+      std::unordered_map< RouterID, MessageQueue, RouterID::Hash >
+          outboundMessageQueue GUARDED_BY(_mutex);
+
+      ILinkManager* _linkManager;
```
majestrate: should this also be a `const *`?

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,299 @@
...
+    switch(result)
+    {
+      case RCRequestResult::Success:
+        if (rc)
+        {
+          GotRouterContact(router, *rc);
+        }
+        else
+        {
+          //TODO: shouldn't happen, how to handle?
```
majestrate: maybe `abort()` ? probably log error

llarp/router/outbound_session_maker.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+      std::unordered_map< RouterID,
+                          CallbacksQueue,
+                          RouterID::Hash >
+          pendingCallbacks GUARDED_BY(_mutex);
+
+      ILinkManager *_linkManager;
+      I_RCLookupHandler *_rcLookup;
+      std::shared_ptr< Logic > _logic;
+      llarp_nodedb *_nodedb;
```
majestrate: `const *`

llarp/router/outbound_session_maker.hpp
```diff
@@ -0,0 +1,110 @@
...
+                          std::shared_ptr< PendingSession >,
+                          RouterID::Hash >
+          pendingSessions GUARDED_BY(_mutex);
+
+      std::unordered_map< RouterID,
+                          CallbacksQueue,
+                          RouterID::Hash >
+          pendingCallbacks GUARDED_BY(_mutex);
+
+      ILinkManager *_linkManager;
```
majestrate: `* const` (i kept on saying `const *` >.>)

llarp/router/rc_lookup_handler.cpp
```diff
@@ -0,0 +1,323 @@
...
+    // update dht if required
+    if(_dht->impl->Nodes()->HasNode(dht::Key_t{rc.pubkey}))
+    {
+      _dht->impl->Nodes()->PutNode(rc);
+    }
+
+    // update nodedb if required
+    if (rc.IsPublicRouter())
+    {
+      _nodedb->UpdateAsyncIfNewer(rc);
```
majestrate: this doesn't update the inserted timestamp so it either should be `InsertAsync` or make `UpdateAsyncIfNewer` change insertion timestamps.

llarp/router/rc_lookup_handler.cpp
```diff
@@ -0,0 +1,323 @@
...
+    auto itr = pendingCallbacks.find(router);
+
+    if (itr != pendingCallbacks.end())
+    {
+      
+      auto& callbackQueue = itr->second;
+
+      while (not callbackQueue.empty())
+      {
+        callbackQueue.front()(router, rc, result);
```
majestrate: what happens if the callback eventually wants to acquire the RCLookupHandler lock again?

llarp/router/rc_lookup_handler.hpp
```diff
@@ -0,0 +1,117 @@
...
+      RemoteInBootstrap(const RouterID& remote) const;
+
+      void
+      FinalizeRequest(const RouterID &router,
+                      const RouterContact * const rc,
+                      RCRequestResult result);
+
+      mutable util::Mutex _mutex; // protects pendingCallbacks, whitelistRouters
+
+      llarp_dht_context *_dht = nullptr;
```
majestrate: `* const`

llarp/link/link_manager.cpp
tewinget: So for objects which need to point to one another you would end up mixing and matching this style and in-constructor init of pointers.  *Ideally* that situation wouldn't occur, but it's non-trivial to avoid.  In order to have consistency among these new classes I opted for this style.

llarp/link/link_manager.hpp
tewinget: fair enough.

llarp/link/link_manager.cpp
tewinget: true

llarp/link/link_manager.cpp
tewinget: will address all of these at once.

llarp/link/link_manager.hpp
tewinget: only if I change it to be set in the constructor as @michael-loki suggested, but in case / because some of these things need to point to one another I opted to set all of their pointers in an Init function.  Naturally the pointers are private to at least mostly mitigate the risk of accidentally changing them rather than using them.

llarp/router/i_outbound_message_handler.hpp
tewinget: Yeah, abstract is fine.  Will change for all.

llarp/router/outbound_message_handler.cpp
tewinget: Yeah, I wasn't 100% sure about this bit.  Feels weird to copy into a vector for some reason...but this is how it was being done in router.cpp so I kept it this way.

llarp/router/outbound_message_handler.cpp
tewinget: eh, this is fine for now.  I can change it if it really matters but since the function is still a TODO and the compiler should do the same here regardless it's probably fine.

llarp/router/outbound_message_handler.cpp
tewinget: I guess, but that would be a very interesting case.

llarp/router/outbound_session_maker.cpp
tewinget: If it weren't the line just before I'd agree.

llarp/router/outbound_session_maker.hpp
tewinget: which the compiler corrected several times as a result...

llarp/router/rc_lookup_handler.cpp
tewinget: I think changing UpdateAsyncIfNewer is cleaner, which do you prefer?

llarp/router/rc_lookup_handler.cpp
tewinget: That's...gonna be fun to resolve.

llarp/router/rc_lookup_handler.cpp
tewinget: fair point.  I don't particularly like the idea of collecting the callbacks in a local container first (and then unlocking) but that's probably for the best.

llarp/router/rc_lookup_handler.cpp
majestrate: changing `UpdateAsyncIfNewer` works fine

llarp/link/link_manager.cpp
majestrate: i am segfaulting in my local testing because a timer is called with a member being nullptr and i assumed it's happpening before calling init

tewinget: The `Init`s are run during `Router::FromConfig`.  If that's not early enough then the lack of segfault before was just masking something using `Router` before it was fully configured/initialized.

llarp/link/link_manager.hpp
tewinget: I suppose I could make these pointers `const` and cast the `const` away in the `Init` function, but that also feels messy.  Thoughts?

llarp/router/outbound_message_handler.cpp
tewinget: I'm not sure what you mean by this, but I don't see any congestion function anywhere in the old `SendToOrQueue` codepath.

llarp/link/i_link_manager.hpp
tewinget: @majestrate has used `IClassName` and I guess he prefers that (as do I) so I didn't change that.  Not saying I won't, just haven't yet pending further discussion/convincing.

empty includes done.

llarp/link/link_manager.cpp
tewinget: changed.

llarp/link/link_manager.cpp
tewinget: changed.

llarp/link/link_manager.cpp
tewinget: changed.

llarp/link/link_manager.cpp
tewinget: changed.

llarp/link/link_manager.cpp
tewinget: changed.

llarp/link/link_manager.cpp
tewinget: changed.

llarp/link/link_manager.hpp
tewinget: changed.

llarp/router/i_outbound_message_handler.hpp
tewinget: See comment above, not changed.

llarp/router/outbound_message_handler.cpp
tewinget: changed.

llarp/router/outbound_message_handler.cpp
tewinget: Not changed, can still do if necessary.

llarp/link/link_manager.cpp
tewinget: After further discussion, leaving it as-is.

llarp/link/link_manager.hpp
tewinget: not casting const away, no change.

llarp/router/outbound_message_handler.cpp
tewinget: See comment above, not changed.

llarp/router/outbound_message_handler.cpp
tewinget: Added `LogError` and abort.

llarp/router/outbound_message_handler.hpp
tewinget: changed.

llarp/router/outbound_message_handler.hpp
tewinget: See comment above, not changed.

llarp/router/outbound_session_maker.cpp
tewinget: update on this?  Also, perhaps `make format` can change to the preferred style?  Not sure, haven't really used it, but I imagine it'd be cool if it could.

llarp/router/outbound_session_maker.cpp
tewinget: Not changed.

llarp/router/outbound_session_maker.cpp
tewinget: See comment above, not changed.

llarp/router/outbound_session_maker.cpp
tewinget: added `LogError` but not abort.

llarp/router/outbound_session_maker.hpp
tewinget: See comment above, not changed.

llarp/router/rc_lookup_handler.cpp
tewinget: Changed.

llarp/router/rc_lookup_handler.cpp
tewinget: Addressed.

llarp/router/rc_lookup_handler.cpp
tewinget: See comment above, not changed.

llarp/router/rc_lookup_handler.cpp
tewinget: Addressed.

llarp/router/rc_lookup_handler.hpp
tewinget: See comment above, not changed.

llarp/link/link_manager.cpp
michael-loki: Init functions like this are evils in API design :( but okay

llarp/router/i_outbound_session_maker.hpp
```diff
@@ -0,0 +1,59 @@
...
+  {
+    Establish,
+    Timeout,
+    RouterNotFound,
+    InvalidRouter,
+    NoLink
+  };
+
+  using RouterCallback =
+      std::function< void(const RouterID, const SessionResult) >;
```
michael-loki: `const RouterID&`?

llarp/router/outbound_message_handler.cpp
```diff
@@ -0,0 +1,219 @@
...
+  void
+  OutboundMessageHandler::OnSessionEstablished(const RouterID &router)
+  {
+    FinalizeRequest(router, SendStatus::Success);
+  }
+
+  void
+  OutboundMessageHandler::OnConnectFailure(const RouterID &router)
+  {
+    FinalizeRequest(router, SendStatus::Timeout);
```
michael-loki: is a connect failure always a timeout?

llarp/router/outbound_message_handler.cpp
michael-loki: When I reviewed the function looked like https://github.com/loki-project/loki-network/blob/1d6f010ea9056ff4deee38cd97de3d47727095bb/llarp/router/outbound_message_handler.cpp#L162-L169

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,295 @@
...
+      // made
+      if(pendingSessions.find(router) != pendingSessions.end())
+      {
+        return;
+      }
+    }
+
+    LinkLayer_ptr link = _linkManager->GetCompatibleLink(rc);
+
+    if(link == nullptr)
```
michael-loki: `if(!link)`?

llarp/router/rc_lookup_handler.cpp
```diff
@@ -0,0 +1,342 @@
...
+        if(callbackQueue.size())
+        {
+          collectedCallbacks.reserve(callbackQueue.size());
+        }
+
+        while(not callbackQueue.empty())
+        {
+          collectedCallbacks.push_back(callbackQueue.front());
+          callbackQueue.pop();
+        }
```
michael-loki: I think this could be simplified to be:

```c++
    std::vector< RCRequestCallback > collectedCallbacks;
    {
      util::Lock l(&_mutex);

      auto itr = pendingCallbacks.find(router);

      if(itr != pendingCallbacks.end())
      {
        auto &callbackQueue = itr->second;

        if(callbackQueue.size())
        {
          collectedCallbacks.reserve(callbackQueue.size());
        }

        collectedCallbacks.assign(callbackQueue.begin(), callbackQueue.end());

        pendingCallbacks.erase(itr);
      }
    }  // lock

    for(const auto &callback : collectedCallbacks)
    {
      callback(router, rc, result);
    }
```

llarp/link/link_manager.cpp
tewinget: I can definitely agree with arguments against doing it this way.  It's definitely something to revisit in the future.

llarp/router/i_outbound_session_maker.hpp
tewinget: Yea, will change now.

llarp/router/rc_lookup_handler.cpp
tewinget: yeah, probably.  neat.

llarp/router/outbound_message_handler.cpp
```diff
@@ -25,8 +25,6 @@ namespace llarp
       return false;
     }
 
-    buf.sz = buf.cur - buf.base;
```
majestrate: ebin af

llarp/router/outbound_message_handler.cpp
tewinget: when this func is called, yes.  Changing func name.

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,317 @@
...
+    {
+    }
+  };
+
+  bool
+  OutboundSessionMaker::OnSessionEstablished(ILinkSession *session)
+  {
+    // TODO: do we want to keep it
+
+    const auto &router = RouterID(session->GetPubKey());
```
majestrate: `const RouterID router{session->GetPubKey()};`

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    if(HavePendingSessionTo(router))
+    {
+      return;
+    }
+
+    CreatePendingSession(router);
+
+    LogWarn("Creating session establish attempt to ", router, " .");
```
majestrate: this is a stray `LogWarn`

llarp/router/outbound_session_maker.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+    return true;
+  }
+
+  void
+  OutboundSessionMaker::OnConnectTimeout(ILinkSession *session)
+  {
+    // TODO: retry/num attempts
+
+    LogWarn("Session establish attempt to ", RouterID(session->GetPubKey()),
```
majestrate: also here

majestrate: nvm this is fine

llarp/router/rc_lookup_handler.hpp
```diff
@@ -0,0 +1,112 @@
...
+    void
+    FinalizeRequest(const RouterID &router, const RouterContact *const rc,
+                    RCRequestResult result);
+
+    mutable util::Mutex _mutex;  // protects pendingCallbacks, whitelistRouters
+
+    llarp_dht_context *_dht = nullptr;
+    llarp_nodedb *_nodedb   = nullptr;
+    std::shared_ptr< llarp::thread::ThreadPool > _threadpool;
+    service::Context *_hiddenServiceContext;
```
majestrate: this should set to `nullptr` like the other bare pointers

llarp/router/rc_lookup_handler.hpp
```diff
@@ -0,0 +1,112 @@
...
+    ILinkManager *_linkManager;
```
majestrate: also this too

majestrate:
100 comments and a 3 way merge conflict ain't one :>

majestrate:
YOLO!

PR #729, opened 2019-07-24
make android compile again

majestrate:
this has the `make android` target pull in libuv source code which we could probably do for ios too.


vendor/libtuntap-master/tuntap-unix-linux.c
```diff
@@ -187,13 +187,15 @@ tuntap_sys_set_ipv4(struct device *dev, t_tun_in_addr *s4, uint32_t bits)
 
   return 0;
 }
+#if defined(ANDROID)
 
+#else
```
michael-loki: explain why this is commented out?

majestrate: it's included in android

PR #734, opened 2019-07-26
run main in it's own thread and set its process name.

majestrate:
this would allow in the future final clean up tasks to fire.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -19,6 +19,7 @@
 #endif
 
 struct llarp_main *ctx = 0;
+std::promise< int > exit_code;
 
 void
 handle_signal(int sig)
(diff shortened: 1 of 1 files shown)


tewinget:
I see nothing wrong, but I'll let @michael-loki review and hit the approval button.

daemon/main.cpp
```diff
@@ -273,27 +299,20 @@ main(int argc, char *argv[])
...
+  auto ftr = exit_code.get_future();
+  do
   {
-    signal(SIGINT, handle_signal);
-    signal(SIGTERM, handle_signal);
-#ifndef _WIN32
-    signal(SIGHUP, handle_signal);
-#endif
+    // do periodic non lokinet related tasks here
+  } while(ftr.wait_for(std::chrono::seconds(1)) != std::future_status::ready);
```
michael-loki: Would be nice to take this period as a command line argument, but can be a future extension point.

PR #736, opened 2019-07-26
lokinet-bootstrap improvements

jagerman:
- takes an optional second argument to control where to download
- print usage info if called with >2 arguments or an argument starting with `-`
- Add `set -e` so if unchecked things fail (e.g. the mkdir) the script fails.

I have no idea how to update the `lokinet-bootstrap.exe` file.


jagerman:
This is needed mainly for the debs to be able to download to the system location during package configuration.

PR #743, opened 2019-07-30
Add ASAN CI

michael-loki:
use recent llvm ASAN on macOS (since it's the easiest way to get the most recent compiler)


michael-loki:
@majestrate R4M

PR #744, opened 2019-07-31
Use travis_wait because builds/linkers are slow

michael-loki:
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,11 +93,11 @@ before_install:
 
 script:
   - if [ "$TRAVIS_OS_NAME" == "windows" ]; then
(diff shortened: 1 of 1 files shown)


majestrate:
looks like it still fails, does adding LTO help?

michael-loki:
I can’t imagine doing more work in the linker would make it faster.

majestrate:
is there a flag we can pass to the linker to make it more verbose?

despair86:
maybe object libraries are starting to look like a bad idea
since the link editor can just search the library index in a static archive faster
instead of studying each individual object closely to find missing symbols, one by one

despair86:
besides, what *was* the original motivation for switching to object libraries over static archives?

michael-loki:
we aren't using object libraries at all? so idk what you mean?

PR #746, opened 2019-08-01
Add fail-able docker build for gcc trunk

michael-loki:
Note I've not fixed every warning in trunk just yet, but this deals with a bunch of noise from other compilers too.

This won't fail any builds, but means we can "keep an eye" on things breaking. Builds on top of #744


llarp/net/net_addr.cpp
```diff
@@ -23,12 +23,6 @@ namespace llarp
   {
   }
 
-  Addr::Addr(const Addr& other)
-  {
-    memcpy(&_addr, &other._addr, sizeof(sockaddr_in6));
```
michael-loki: this is what the default copy constructor gives you - bitwise copying.

llarp/ev/ev_libuv.cpp
```diff
@@ -517,7 +519,9 @@ namespace libuv
     bool
     Init(uv_loop_t* loop)
     {
-      strncpy(m_Device->if_name, m_Tun->ifname, sizeof(m_Device->if_name));
+      std::copy(std::begin(m_Tun->ifname), std::end(m_Tun->ifname),
+                m_Device->if_name);
+      memcpy(m_Device->if_name, m_Tun->ifname, sizeof(m_Device->if_name));
```
majestrate: redundant copy?

llarp/path/path.cpp
```diff
@@ -103,8 +103,8 @@ namespace llarp
     }
 
     bool
-    Path::HandleLRSM(uint64_t status, std::array< EncryptedFrame, 8 >& frames,
-                     AbstractRouter* r)
+    Path::HandleLRSM(ABSL_ATTRIBUTE_UNUSED uint64_t status,
```
majestrate: `status` is not unused

llarp/service/info.hpp
```diff
@@ -66,18 +66,6 @@ namespace llarp
...
 
-      ServiceInfo&
-      operator=(const ServiceInfo& other)
-      {
-        enckey  = other.enckey;
-        signkey = other.signkey;
-        version = other.version;
-        vanity  = other.vanity;
-        version = other.version;
-        UpdateAddr();
```
majestrate: `UpdateAddr` needs to be called on assignment to recompute the cached address

vendor/libtuntap-master/tuntap-unix.c
```diff
@@ -169,7 +169,7 @@ tuntap_set_ifname(struct device *dev, const char *ifname)
   }
 
   (void)memset(dev->if_name, 0, IF_NAMESIZE);
-  (void)strncpy(dev->if_name, ifname, len);
+  (void)strncpy(dev->if_name, ifname, len + 1);
```
majestrate: should be `sizeof(dev->if_name)` right?

majestrate: rather `min(sizeof(dev->ifname), len)`

llarp/service/info.hpp
michael-loki: Does it? Just naive copying should have the same effect?

vendor/libtuntap-master/tuntap-unix.c
michael-loki: that's wrong those, because if `ifname` is as long as `IF_NAMESIZE`, there is no room for the trailing '\0'

majestrate: `len` is calculated based of `ifname` and it it exceeds that it needs to truncate

majestrate: `ifname` not to be confused with `dev->if_name`

llarp/service/info.hpp
michael-loki: Just for citation, I think this is fine, because `UpdateAddr` just calls:

- `BEncode` into a temp array (which is stable)
- `crypto_generichash_blake2b` into `m_CachedAddr` (which is meant to be stable since it's a hash function)

this is replicable by just copying `m_CachedAddr`

vendor/libtuntap-master/tuntap-unix.c
michael-loki: So consider this case:

1. `IF_NAMESIZE` is 4 (just to make this easier for me to type out)
2. `ifname` is `loki\0` (making `strlen(ifname)` == 4)


```
char arr[IF_NAMESACE];
strncpy(arr, ifname, strlen(ifname));
```

at the end of this, `arr` is `['l', 'o', 'k', 'i']`, without any trailing '\0' character (making it an invalid c-string)

llarp/service/info.hpp
majestrate: okay yeah that's fine

vendor/libtuntap-master/tuntap-unix.c
majestrate: `len` is calculated based on `ifname`  (user input) not `dev->if_name` (target buffer)

PR #747, opened 2019-08-01
bug fixes

majestrate:
* fix utp socket leak
* track path build fails and backoff on fail
* have ip packets `srcv6` and `dstv6` return expanded v4 address if ipv4
* remove internal utp connection limit `:^)`

### DIFF
--- a/libutp/utp_internal.cpp
+++ b/libutp/utp_internal.cpp
@@ -3416,7 +3416,7 @@ utp_process_udp(utp_context *ctx, const byte *buffer, size_t len,
 
       return 1;
     }
-
+    /*
     if(ctx->utp_sockets->GetCount() > 3000)
     {
 #if UTP_DEBUG_LOGGING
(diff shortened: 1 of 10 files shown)


libutp/utp_internal.cpp
```diff
@@ -3416,7 +3416,7 @@ utp_process_udp(utp_context *ctx, const byte *buffer, size_t len,
 
       return 1;
     }
-
+    /*
```
tewinget: Probably best to remove code if it's not going to be used, rather than commenting it out.

llarp/path/path.cpp
```diff
@@ -143,7 +143,7 @@ namespace llarp
         ++index;
       }
 
-      if((currentStatus & LR_StatusRecord::SUCCESS) == 1)
```
tewinget: lol woops

tewinget:
Seems fine other than the block of code commented out rather than removed.

libutp/utp_internal.cpp
majestrate: disagree, we may want to re-enable it later (maybe)

tewinget: I suppose I can live with that for now, but that policy merits further discussion.

PR #749, opened 2019-08-04
Set fast_finish so we don't need to wait for the sloooow 'make windows' target

michael-loki:


michael-loki:
damn, looks like the image hasn't pulled it in yet.

PR #750, opened 2019-08-05
WIP on docker-compose workflow

michael-loki:
to build and run every:

```
$ docker-compose -f ./docker/compose/docker-compose.yml --project-directory . up --scale router=10
```

will start 1 bootstrap and 10 normal nodes.

### DIFF
--- a/docker/compose/bootstrap.ini
+++ b/docker/compose/bootstrap.ini
@@ -1,0 +1,76 @@
+# this configuration was auto generated with 'sane' defaults
+# change these values as desired
+
+
+[router]
+# number of crypto worker threads
+threads=4
+# path to store signed RC
+contact-file=/root/.lokinet/self.signed
+# path to store transport private key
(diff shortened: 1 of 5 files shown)


docker/compose/bootstrap.ini
```diff
@@ -0,0 +1,76 @@
...
+
+# network settings
+[network]
+profiles=/root/.lokinet/profiles.dat
+enabled=true
+exit=false
+#exit-blacklist=tcp:25
+#exit-whitelist=tcp:*
+#exit-whitelist=udp:*
+ifaddr=10.200.0.1/8
```
majestrate: why `/8`?

michael-loki: It’s what’s in https://github.com/loki-project/loki-network/blob/2bdafc66033591e2342d01862ef8b08bfd6c6f4b/lokinet-docker.ini#L77, can probably be tweaked

PR #754, opened 2019-08-06
Disable metrics by default

majestrate:


llarp/config/config.cpp
```diff
@@ -266,13 +266,21 @@ namespace llarp
   void
   MetricsConfig::fromSection(string_view key, string_view val)
   {
-    if(key == "disable-metrics")
+    if(key == "enable-metrics")
     {
-      disableMetrics = true;
+      disableMetrics = IsFalseValue(val);
```
michael-loki: I think if we’re doing it like this, we should rename the variable to `enableMetrics`

majestrate: i added an option for that too

michael-loki: I mean, if metrics are disabled by default, we should make the variable “enableMetrics”, with a default value of false. 

Positive bools are easier to understand imo.

majestrate: i agree

PR #760, opened 2019-08-07
CI to catch missing `make format`

### LINKED ISSUE #727: Add a travis-ci build to check for code formatting issues
We have periodic merge conflicts happening from time to time because of code formatting PRs; it would be nice to avoid them by having an extra travis-ci "meta" job that doesn't actually build anything but instead just runs a bunch of style/documentation/code fixes.  Basically a test job for checking all the non-compiled parts of the project.

That way rather than PR -> merge -> reformat -> PR -> merge we can just get a failure in the original PR build for the formatting problems which can then be fixed in the original PR.

michael-loki:
What @jagerman wants, @jagerman gets. fixes #727

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,6 +243,9 @@ abyss: debug
 format:
 	clang-format -i $$(find jni daemon llarp include libabyss | grep -E '\.[h,c](pp)?$$')
 
+format-verify: format
+	git diff --quiet || echo 'Please run make format!!' && git --no-pager diff && false
+
 analyze-config: clean
 	mkdir -p '$(BUILD_ROOT)'
 	$(ANALYZE_CONFIG_CMD)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -243,6 +243,9 @@ abyss: debug
 format:
 	clang-format -i $$(find jni daemon llarp include libabyss | grep -E '\.[h,c](pp)?$$')
 
+format-verify: format
+	git diff --quiet || echo "Please run make format!!" && git --no-pager diff && false
```
majestrate: shouldn't this be `git format --quiet` ? or something?

michael-loki: Not sure what you mean:

```
$ git format --help
git: 'format' is not a git command. See 'git --help'.
```

what this PR essentially does:

- run the normal `make format` check
- run `git diff`. If there are any changes and there shouldn't be, because we're perfect and never forget to run `make format`, and travis does a clean checkout.
- print the diff if there are any changes

PR #761, opened 2019-08-08
don't close utp session as client if write direction is full

majestrate:


llarp/utp/session.cpp
```diff
@@ -212,7 +212,9 @@ namespace llarp
         // don't time out the connection if backlogged in downstream direction
         // for clients dangling off the side of the network
         const auto timestamp =
-            remoteRC.IsPublicRouter() ? lastSend : lastActive;
+            remoteRC.IsPublicRouter() && parent->GetOurRC().IsPublicRouter()
```
michael-loki: Once the ternary is on 4 lines, I think we want:

```c++
const bool isParent = remoteRC.IsPublicRouter() && parent->GetOurRC().IsPublicRouter();
const auto timestamp = isParent ? lastSend : lastActive;
```

jagerman: I rather like multiline ternaries (with the `?` and `:` leading on separate lines like this).

PR #764, opened 2019-08-09
Lokinet monitor

majestrate:
add in contrib a python ncurses monitor application


contrib/py/admin/.gitignore
```diff
@@ -0,0 +1 @@
+v/
```
michael-loki: missing newline here.

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,166 @@
...
+      self.update_data()
+      self.display_data()
+      time.sleep(0.5)
+
+
+
+if __name__ == '__main__':
+  import sys
+  c = Canvas(len(sys.argv) > 1 and sys.argv[1] or "http://127.0.0.1:1190/jsonrpc")
+  c.run()
```
michael-loki: missing newline here.

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env python3
```
michael-loki: it would be nice to run this through something like [`black`](https://black.readthedocs.io/en/stable/) so we have it nicely formatted.

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,166 @@
...
+
+  def update_data(self):
+    """update data from lokinet"""
+    try:
+      j = self.jsonrpc("llarp.admin.dumpstate", {})
+      self.data = j['result']
+    except:
+      self.data = None
+
+  def _render_path(self,y, path, name):
```
michael-loki: Probably not worth it at this scale, but `mypy` style typings could be useful here (since I have no idea what `y`, `path` or `name` are meant to be)

majestrate: yeah this was just a quick weekend project to do

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+
+    _speedSamples = 8
+    _globalspeed = []
+
+    def __init__(self, url):
+        self.data = dict()
+        self.win = curses.initscr()
+        self._url = url
+        while len(self._globalspeed) < self._speedSamples:
+            self._globalspeed.append((0, 0))
```
jagerman:
```Python
self._globalspeed += [(0, 0)] * (self._speedSamples - len(self._globalspeed))
```

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+                "params": params,
+            },
+        )
+        return r.json()
+
+    def update_data(self):
+        """update data from lokinet"""
+        try:
+            j = self.jsonrpc("llarp.admin.dumpstate", {})
+            self.data = j["result"]
```
jagerman:
```Python
self.data = self.jsonrpc("llarp.admin.dumpstate", {})["result"]
```

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+        except Exception as ex:
```
jagerman: Don't need `as ex` if you aren't using `ex`

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+    def _render_path(self, y, path, name):
+        """render a path at current position"""
+        self.win.move(y, 1)
+        self.win.addstr("({}) ".format(name))
+        y += 1
+        self.win.move(y, 1)
+        y += 1
+        self.win.addstr("me -> ")
+        for hop in path["hops"]:
+            self.win.addstr(" {} ->".format(hop["router"][:4]))
```
jagerman: Loops are overrated in python:
```Python
self.win.addstr(" -> ".join(("me", *(hop["router"] for hop in path["hops"])))
```

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+        now = time.time() * 1000
+        return "{} seconds".format(int((ts - now) / 1000))
+
+    def speedOf(self, rate):
+        """turn int speed into string formatted"""
+        units = ["B", "KB", "MB", "GB"]
+        idx = 0
+        while rate > 1000 and idx < len(units):
+            rate /= 1000.0
+            idx += 1
```
jagerman:
```Python
import math
idx = int(math.log10(rate)) // 3
```

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+        return "{} {}ps".format("%.2f" % rate, units[idx])
```
jagerman:
```Python
return "{:.2f} {}ps".format(rate, units[idx])
```

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def display_service(self, y, name, status):
+        """display a service at current position"""
+        self.win.move(y, 1)
+        self.win.addstr("service [{}]".format(name))
+        build = status["buildStats"]
+        ratio = build["success"] / (build["attempts"] or 1)
```
jagerman: I'm not sure what will be in `build` here, but if `"attempts"` is not present this will throw a KeyError.  It's fine if it'll be present and falsey (False, None, 0), but if it can be not present you'd need something (ugly) like:
```Python
... / (build["attempts"] or 1 if "attempts" in build else 1)
```
or else set it beforehand.  (But maybe it's always set so this doesn't matter).

contrib/py/admin/dumpstats.py
```diff
@@ -0,0 +1,266 @@
...
+        y += 1
+        self.win.move(y, 1)
+        self.win.addstr("build success: {} %".format(int(100 * ratio)))
```
jagerman: Maybe useful to have lower than integer precision here?
```Python
"build success: {:.1f} %".format(100 * ratio)
```

contrib/py/admin/dumpstats.py
majestrate: the `or 1` is to prevent division by 0

majestrate:
some stray commits were included, force pushed

PR #768, opened 2019-08-12
Sane limits

majestrate:
implement limiting


llarp/constants/limits.cpp
```diff
@@ -0,0 +1,13 @@
+#include <constants/limits.hpp>
+
+namespace llarp
+{
+  namespace limits
+  {
+    /// snode limit parameters
+    const LimitParameters snode = {6, 60};
```
michael-loki: is this something we would want to override at the command line/environment at all?

majestrate: yes and we do

llarp/router/router.cpp
```diff
@@ -655,16 +658,16 @@ namespace llarp
 
     const size_t connected = NumberOfConnectedRouters();
     const size_t N         = nodedb()->num_loaded();
-    if(N < minRequiredRouters)
+    if(N < 4)
```
tewinget: err...changed *from* a named amount to a hardcoded amount?

majestrate:
On Mon, Aug 12, 2019 at 05:06:32PM -0700, Thomas Winget wrote:
> tewinget commented on this pull request.
> 
> 
> 
> > @@ -655,16 +658,16 @@ namespace llarp
>  
>      const size_t connected = NumberOfConnectedRouters();
>      const size_t N         = nodedb()->num_loaded();
> -    if(N < minRequiredRouters)
> +    if(N < 4)
> 
> err...changed *from* a named amount to a hardcoded amount?
that was a mistake hmmm

llarp/router/router.cpp
majestrate: fixed

llarp/config/config.hpp
```diff
@@ -105,7 +105,7 @@ namespace llarp
     size_t m_minConnectedRouters = 2;
 
     /// hard upperbound limit on the number of router to router connections
-    size_t m_maxConnectedRouters = 2000;
+    size_t m_maxConnectedRouters = 5;
```
jagerman: Was this intentional?  5 seems really low.

llarp/router/outbound_session_maker.hpp
```diff
@@ -52,11 +52,20 @@ namespace llarp
...
+
     void
     Init(ILinkManager *linkManager, I_RCLookupHandler *rcLookup,
          std::shared_ptr< Logic > logic, llarp_nodedb *nodedb,
          std::shared_ptr< llarp::thread::ThreadPool > threadpool);
 
+    /// always maintain this many connections to other routers
+    size_t minConnectedRouters = 4;
+    /// hard upperbound limit on the number of router to router connections
+    size_t maxConnectedRouters = 6;
```
jagerman: ... and likewise here (but also 6 here vs 5 earlier; perhaps a constant for this instead?)

llarp/config/config.hpp
majestrate: remember this applies to clients too

llarp/router/outbound_session_maker.hpp
majestrate: overriden by limits later

jagerman: Actually I think this is just the comment that is wrong -- this looks like the client limits rather than router limits.

majestrate: yeah the comments are wrong, i could initialize with zero but that'd be weird. the values are overrided by limits later in the code.

PR #777, opened 2019-08-15
windows bug fixes

despair86:
- added `lokinet-rcutil(8)` to windows install pkg
- fixed console colours using win32 legacy console
- restored cmake link flags for `*-*-winnt-gnu` (if this causes travis ci to fail `*-*-winnt-msvc` build, let me know so i can add the relevant pragma directly)


cmake/win32.cmake
```diff
@@ -21,9 +21,7 @@ if(NOT MSVC_VERSION)
   # GNU ld sees fit to merge *all* the .ident sections in object files
   # to .r[o]data section one after the other!
   add_compile_options(-fno-ident -Wa,-mbig-obj)
-  find_library(shlwapi_lib shlwapi)
-  find_library(dbghelp_lib dbghelp)
-  link_libraries( ${shlwapi_lib} ${dbghelp_lib} )
+  link_libraries( -lshlwapi -ldbghelp )
```
michael-loki: why revert this? adding `-l` to `link_libraries` is bad practice.

despair86: abseil writes link-editor directives straight into the object file for MSVC, so only clang for windows and gcc require it in cmake

majestrate:
is this ready?

despair86:
holy hell i almost lost track of this one. yeah it should work
idk why macintosh decided to throw a hissy fit overnight

majestrate:
travis being meh i would assume

PR #780, opened 2019-08-19
Replace StatusObject with underlying JSON type

michael-loki:


llarp/handlers/exit.cpp
```diff
@@ -51,9 +51,9 @@ namespace llarp
       util::StatusObject exitsObj{};
       for(const auto &item : m_ActiveExits)
       {
-        exitsObj.Put(item.first.ToHex(), item.second->ExtractStatus());
+        exitsObj[item.first.ToHex()] = item.second->ExtractStatus();
```
majestrate: while you're here it ideally should be `ToString()` not `ToHex()` so it reports the snode address

llarp/service/endpoint_state.cpp
```diff
@@ -116,31 +116,40 @@ namespace llarp
...
+      std::transform(
+          m_SNodeSessions.begin(), m_SNodeSessions.end(),
+          std::back_inserter(obj["snodeSessions"]),
+          [](const auto& item) { return item.second.first->ExtractStatus(); });
 
       util::StatusObject sessionObj{};
 
       for(const auto& item : m_Sessions)
       {
         std::string k = item.first.ToHex();
```
majestrate: also `ToString()`

michael-loki:
Will do.

llarp/service/endpoint_state.cpp
michael-loki: `No member named 'ToString' in 'llarp::AlignedBuffer<16>'`

majestrate: okay nvm there then

michael-loki:
@majestrate fixed up most of the instances of `ToHex`

PR #782, opened 2019-08-19
replace endl with '\n in config.cpp

michael-loki:
`\n` is 11 fewer characters than ` << std::endl`


jagerman:
Ah, but it's not equivalent because it doesn't flush.  It's better: because there's no need to flush at all.

PR #787, opened 2019-08-27
make build internally consistent, bump windows version to 0.5.0

despair86:
- abseil, cppbackport (if selected), gtest, and gmock now build with the same options as the rest of the project
- added build config for embedded ix86 (requires a 486 or later)
- stripped target selection flags from MSVC build
- bump version to 0.5.0
- make sure sun service manifest matches systemd unit file
- clang-format all crypto code
- use `__asm__` instead of `asm` in avx2 port since we insist on straight C99 without extensions (`asm` isn't special in C99, but `__asm__` is)
- fix win32 tap bug
- UI fixes for windows


cmake/unix.cmake
```diff
@@ -26,7 +26,11 @@ include_directories(${LIBUV_INCLUDE_DIRS})
 
 function(check_working_cxx_atomics64 varname)
   set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
-  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=c++14")
+  if (EMBEDDED_CFG)
```
michael-loki: Can remove the duplication + else block here by moving the -std=c++14 out of the if.

michael-loki: what I meant was:

```cmake
set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=c++14")
if (EMBEDDED_CFG)
  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -m32 -march=i486")
endif()
```

cmake/unix.cmake
```diff
@@ -62,10 +67,13 @@ function(link_libatomic)
   message(FATAL_ERROR "Host compiler must support 64-bit std::atomic!")
 endfunction()
 
+if(EMBEDDED_CFG)
```
michael-loki: remove line 77, and then have:

```cmake
if (EMBEDDED_CFG OR ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
  link_libatomic()
endif()
```

cmake/win32.cmake
```diff
@@ -25,8 +66,47 @@ if(NOT MSVC_VERSION)
   add_definitions(-DWINVER=0x0500 -D_WIN32_WINNT=0x0500)
   # Wait a minute, if we're not Microsoft C++, nor a Clang paired with Microsoft C++,
   # then the only possible option has to be GNU or a GNU-linked Clang!
-  if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 9.0 OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    set(FS_LIB stdc++fs)
+  set(FS_LIB stdc++fs)
+  # If we're building a lokinet for windows, but need to target ancient hardware  
+  function(check_working_cxx_atomics64 varname)
```
michael-loki: this seems like we should hoist this function to its own file, and not duplicate between all the different flavours here.

llarp/win32/abyss.rc
```diff
@@ -58,8 +60,8 @@ END
 //
 
 VS_VERSION_INFO VERSIONINFO
- FILEVERSION 0,4,0,0
- PRODUCTVERSION 0,4,0,0
```
majestrate: i think we should use the `llarp/constants/version.hpp` file for these

llarp/constants/version.hpp
```diff
@@ -33,5 +37,5 @@ struct Version
 {
   static const char LLARP_NET_ID[];
 };
-
+#endif
```
majestrate: that `#endif` should be a bit higher up so we have the Version struct and netid in windows

despair86: it doesn't affect normal builds, hence `#ifdef RC_INVOKED`
when `windres(1)` or `rc(1)` run, `/lib/cpp` or rc's internal pre-ASA cpp define the symbol `RC_INVOKED`, which allows you to hide anything that RC may not understand
this is because its internal cpp dates back to the '80s and predates ASA C

majestrate: oh

PR #789, opened 2019-08-27
bump handover window up and space out path builds more evenly

majestrate:


llarp/service/intro.hpp
```diff
@@ -30,7 +30,7 @@ namespace llarp
       }
 
       bool
-      ExpiresSoon(llarp_time_t now, llarp_time_t dlt = 15000) const
+      ExpiresSoon(llarp_time_t now, llarp_time_t dlt = 30000) const
```
michael-loki: maybe we should name this instead somewhere. 30000 *what*

michael-loki: my eyes are terrible, once a number has more than 2 zeros, I can't tell.  :P 

Is this 3 seconds, 30 seconds or 5 minutes (also `dlt` isn't a great variable name)...

llarp/dht/recursiverouterlookup.cpp
```diff
@@ -76,11 +76,10 @@ namespace llarp
         parent->DHTSendTo(
             whoasked.node.as_array(),
             new GotRouterMessage({}, whoasked.txid, valuesFound, false), false);
-
-        // store this in our nodedb for caching
-        if(valuesFound.size() > 0)
-          parent->StoreRC(valuesFound[0]);
       }
+      // store this in our nodedb for caching
+      if(valuesFound.size() > 0)
```
michael-loki: `if(not valuesFound.empty())`?

PR #792, opened 2019-08-29
bump versions and call timeout hook

majestrate:
prepares for 0.5.0 release
cc @despair86 prepare windows for next release too.


majestrate:
yeah let's squash

PR #796, opened 2019-08-30
optimize

majestrate:
this makes it faster


majestrate:
the culprit was `std::back_inserter`

majestrate:
something like that

PR #799, opened 2019-09-03
disable link layer locking

majestrate:
i found no lock contention on the link layer mutexes with tracy so we should revert these into Null{Lock,Mutex} like they were before.


majestrate:
i think we should start a dev branch for these changes

PR #812, opened 2019-09-05
macOS installer bump / Other windows fixes

despair86:
for @neuroscr 

also:
- don't warn when netid == default netid (some scripts set netid explicitly)
- windows error feedback for @KeeJef (instead of showing random network setup errors, tells the user to check their config file, just like all the other platforms were already doing)

### DIFF
--- a/llarp/ev/ev.cpp
+++ b/llarp/ev/ev.cpp
@@ -93,7 +93,6 @@ llarp_ev_udp_sendto(struct llarp_udp_io *udp, const sockaddr *to,
 bool
 llarp_ev_add_tun(struct llarp_ev_loop *loop, struct llarp_tun_io *tun)
 {
-#if !defined(_WIN32)
   if(tun->ifaddr[0] == 0 || strcmp(tun->ifaddr, "auto") == 0)
   {
     LogError("invalid ifaddr on tun: ", tun->ifaddr);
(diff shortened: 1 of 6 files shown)


llarp/ev/ev.cpp
```diff
@@ -93,7 +93,6 @@ llarp_ev_udp_sendto(struct llarp_udp_io *udp, const sockaddr *to,
 bool
 llarp_ev_add_tun(struct llarp_ev_loop *loop, struct llarp_tun_io *tun)
 {
-#if !defined(_WIN32)
   if(tun->ifaddr[0] == 0 || strcmp(tun->ifaddr, "auto") == 0)
   {
     LogError("invalid ifaddr on tun: ", tun->ifaddr);
```
despair86: we should make this more descriptive?

osx-setup/distribution.xml
```diff
@@ -10,7 +10,7 @@
   <conclusion file="conclusion.html" mime-type="text/html" />
   <!-- List all component packages -->
   <pkg-ref id="network.loki.lokinet"
-           version="0.4"
+           version="0.6"
```
majestrate: 0.5?

despair86: @majestrate this PR is to be merged no earlier than tomorrow, once the v0.5 release is out?

Issue #816, opened 2019-09-09
remove all exe files from source tracking and build them from source.


majestrate:
cc @despair86

despair86:
mostly done
windows drivers, 7-zip, bootstrap all moved to off-site location and downloaded at runtime, bootstrap is downloaded at build time (see `win32-setup/Makefile`)
registry helper lib built on-demand

despair86:
move to #825

despair86:
closed in #825

PR #825, opened 2019-09-11
remove binaries from source tree

despair86:
- fix a bug in network auto config on windows
- bump to v0.5.1
- download windows bits at runtime
- ~~download a known good lokinet-bootstrap at installer build time as it's non-trivial to build for windows (requires TLS)~~
- ~~refactor lokinet-bootstrap for windows, more robust http response parsing~~
- ryan's macOS install bits
see #816


contrib/lokinet-bootstrap-winnt/bootstrap.c
```diff
@@ -2452,22 +2453,136 @@ const void *b;
 size_t bn;
 size_t s;
 {
-  char *p = malloc(s * (an + bn));
```
majestrate: please use 2 spaces not tabs kthnx

win32-setup/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+all: regdbhelper.dll lokinet-bootstrap.exe
+
+default: all
+
+regdbhelper.dll:
+	i686-w64-mingw32-gcc regdb_helper.c -o $@ -shared -Os -s
+
+lokinet-bootstrap.exe:
+	wget https://snowlight.net/loki/win32-dist/lokinet-bootstrap.exe
```
majestrate: this target should build the executable not download it.

majestrate:
honestly it should just require libcurl

despair86:
ok custom code replaced with a JIT build of libcurl+mbedtls to match all other platforms

despair86:
`LICENSE` changed to have windows line endings to placate UI about box

despair86:
ok made crlf conversion run at installer build time instead

PR #827, opened 2019-09-16
Add initial macOS app

michael-loki:
The shutdown behaviour is less-than-perfect right now (the resetting of the DNS settings doesn't always kick in).


majestrate:
missing some files it seems
also this should go into dev

michael-loki:
good spot(s)

PR #828, opened 2019-09-16
Add initial macOS app

michael-loki:


majestrate:
stray file? `ui-macos/.DS_Store`

despair86:
ew resource forks

despair86:
also, isn't the xcode/mpw project stuff machine specific, i.e. you can't take the project files into source control

ui-macos/lokinet/DNSManager.swift
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    func setNewSettings() throws {
+        let netprocess = Process()
+        netprocess.executableURL = DNSManager.netSetup
+
+        netprocess.arguments = ["-setdnsservers", self.interface]
+        netprocess.arguments?.append("127.0.0.1")
+        netprocess.arguments?.append(contentsOf: oldDNSSettings)
```
majestrate: the dns server should be ONLY lokinet so that leaks don't make lookups return false positives.

KeeJef:
Bump on this, should be included in either a point release or 0.6.0. What are we waiting on before merge?

michael-loki:
@KeeJef I thought I pushed my windows fixes, done now.

PR #834, opened 2019-09-19
try fixing handover

majestrate:
compiles and tests pass


majestrate:
nah merge dev into master after this is merged

PR #842, opened 2019-09-23
Enable more sanitisers

michael-loki:


llarp/iwp/linklayer.cpp
```diff
@@ -90,6 +90,7 @@ namespace llarp
       }
       else
       {
+        Lock lock(&m_AuthedLinksMutex);
```
neuroscr: I thought we said this lock was not needed on discord?

michael-loki: We use the null lock/null mutex here now.

PR #847, opened 2019-09-30
Distinct ios build

michael-loki:
Rather than litter the main cmake build with tonnes of iOS specific things, create a distinct new "top level" file, and refactor.


majestrate:
travis complains:
`CMake 3.15 or higher is required.  You are running version 3.14.2`

PR #855, opened 2019-10-04
Introduce --background to only start JSON RPC

### LINKED ISSUE #853: Make lokinet startup do nothing until it receives an enable/disable

michael-loki:
fixes #853

what's currently broken:
- ctrl-c to kill if you've not sent the start command(??)

### DIFF
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -110,11 +110,18 @@ namespace llarp
   util::StatusObject
   Router::ExtractStatus() const
   {
-    return util::StatusObject{
-        {"dht", _dht->impl->ExtractStatus()},
-        {"services", _hiddenServiceContext.ExtractStatus()},
-        {"exit", _exitContext.ExtractStatus()},
-        {"links", _linkManager.ExtractStatus()}};
+    if(_running)
+    {
+      return util::StatusObject{
+          {"dht", _dht->impl->ExtractStatus()},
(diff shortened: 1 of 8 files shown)


llarp/router/router.cpp
```diff
@@ -110,11 +110,18 @@ namespace llarp
...
+      return util::StatusObject{
+          {"dht", _dht->impl->ExtractStatus()},
+          {"services", _hiddenServiceContext.ExtractStatus()},
+          {"exit", _exitContext.ExtractStatus()},
+          {"links", _linkManager.ExtractStatus()}};
+    }
+    else
+    {
+      return util::StatusObject{{"notRunning", true}};
+    }
```
majestrate: this should set `running` to `true` if running and return only an object with `running` as `false` in the case it's not.

llarp/rpc/rpc.cpp
```diff
@@ -271,6 +278,10 @@ namespace llarp
       HandleJSONRPC(Method_t method,
                     ABSL_ATTRIBUTE_UNUSED const Params& params) override
       {
+        if(method == "llarp.admin.start")
+        {
+          return StartRouter();
+        }
```
majestrate: this should be `llarp.admin.wakeup` so we can have a `llarp.admin.suspend` as well

michael-loki: I was thinking start + stop, versus wakeup. will defer to you though

PR #857, opened 2019-10-07
Add rpc call to get current addresses for our node

michael-loki:
Builds on top of #856 

Also, add the ability to use libcurl to hit the jsonrpc endpoint (the libcurl dependency is optional, if it's not available on the system, support simply isn't compiled in).


daemon/rcutil.cpp
```diff
@@ -55,26 +122,31 @@ main(int argc, char* argv[])
 
   options.add_options()
       ("v,verbose", "Verbose", cxxopts::value<bool>())
-      ("c,colour", "colour output", cxxopts::value<bool>()->default_value("true"))
       ("h,help", "help", cxxopts::value<bool>())
-      ("j,json", "output in json", cxxopts::value<bool>())
```
michael-loki: based on feedback from @jagerman, the non-json output isn't that useful, so make it the default.

llarp/config/config.cpp
```diff
@@ -475,6 +475,23 @@ namespace llarp
...
   }
 
+  fs::path
+  GetDefaultConfigDir()
+  {
+#ifdef _WIN32
+    fs::path homedir = fs::path(getenv("APPDATA"));
+#else
+    fs::path homedir = fs::path(getenv("HOME"));
+#endif
```
majestrate: can these be `const` ?

michael-loki: yes

majestrate:
looks like ci wants the package: `libcurl4-openssl-dev` not `libcurl-dev`

necro-nemesis:
Glad to see this.

PR #859, opened 2019-10-09
Windows build fixes

michael-loki:


cmake/win32.cmake
```diff
@@ -24,7 +24,7 @@ if(NOT MSVC_VERSION)
   add_definitions(-DWINVER=0x0500 -D_WIN32_WINNT=0x0500)
   # Wait a minute, if we're not Microsoft C++, nor a Clang paired with Microsoft C++,
   # then the only possible option has to be GNU or a GNU-linked Clang!
-  set(FS_LIB stdc++fs)
+  # set(FS_LIB stdc++fs)
```
despair86: but why tho

michael-loki: Latest mingw ships with a non-broken stdlib

despair86: ok apparently only mainstream targets get this special treatment

t. builds on a bunch of niche, arcane, yet up to date platforms that still ship a separate c++17 filesystem ts lib

despair86: wrap it around a library check or version check at least

despair86:
Merging, and reversing the patch listed above
thank you for your service mike

PR #860, opened 2019-10-11
Null mutex clarity (via comments / log statement)

notlesh:
Leaving some additional breadcrumbs for anyone who stumbles across NullMutex might help it achieve its purpose.

### DIFF
--- a/llarp/util/thread/threading.hpp
+++ b/llarp/util/thread/threading.hpp
@@ -31,12 +31,12 @@ namespace llarp
   namespace util
   {
     /// a mutex that does nothing
-    /// 
-    /// this exists to convert mutexes that were initially in use (but may no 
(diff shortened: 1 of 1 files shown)


llarp/util/thread/threading.hpp
```diff
@@ -31,9 +31,16 @@ namespace llarp
   namespace util
   {
     /// a mutex that does nothing
+    /// 
+    /// this exists to convert "old" mutexes that may no longer be necessary
```
michael-loki: nit: I would note that previously they were *real* mutexes too.

notlesh: added even more comments in a3c48e2

majestrate:
`make format`

PR #862, opened 2019-10-11
Add numNodesKnown to llarp.admin.dumpstate RPC endpoint

notlesh:
This adds the number of nodes in the `nodedb` to the `llarp.admin.dumpstate` endpoint. 

It might be better / more consistent to add an `ExtractStatus()` method to `nodedb`, though there doesn't seem to be much else in `nodedb` we might want to know.

Also note that `num_loaded()` is going to acquire the `nodedb`'s `access mutex`; would this be problematic?

### DIFF
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -114,6 +114,7 @@ namespace llarp
     {
       return util::StatusObject{
           {"running", true},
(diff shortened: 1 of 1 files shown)


majestrate:
the `access` mutex acquisition should not be a problem

PR #863, opened 2019-10-11
Fix build on xenial

jagerman:
- xenial's cmake version (3.5.1) builds everything fine and test suite
  passes, so lower the minimum to that.

- add a hack for xenial's kernel header & glibc version breaking if both
  net/if.h and linux/if.h get included.  The only thing we actually need
  from net/if.h that linux/if.h doesn't have is `if_nametoindex`, so
  just hack that definition in for xenial's specific glibc/kernel header
  versions.

This travis-ci build doesn't hit either of these issues because it isn't fully xenial - it's xenial plus updated headers and cmake versions because WTF travis.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
-# Lowest version - android ndk 3.6.0
-cmake_minimum_required(VERSION 3.6.0)
+cmake_minimum_required(VERSION 3.5.1)  # xenial's cmake version
 
 find_program(CCACHE_PROGRAM ccache)
 if(CCACHE_PROGRAM)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
-# Lowest version - android ndk 3.6.0
-cmake_minimum_required(VERSION 3.6.0)
+cmake_minimum_required(VERSION 3.5.1)
```
michael-loki: Can you comment why we need this version 3.5.1 is :(

jagerman: I assume you mean in the code?  Added a comment to CMakeLists.txt.

PR #864, opened 2019-10-11
fully integrated windows package

despair86:
with everything and the kitchen sink

a. GCC LTO requires specifiying CXXFLAGS in LDFLAGS
b. not my fault your machine doesn't link pthread properly (What does this affect? Raspbian users?)
c. grab everything and the kitchen sink at installer build time
d. devendor libsodium from win32 by generating patchset to be applied to upstream
e. remove vendored libsodium


win32-setup/Makefile
```diff
@@ -29,22 +29,40 @@ curl:
...
+	wget https://snowlight.net/loki/win32-dist/dbghelp64.dll
 
+else
 regdbhelper.dll:
-	scp despair@10.10.10.6:loki-network/win32-setup/*.dll .
+	scp despair@10.10.10.3:web/loki/win32-dist/*.dll .
 
 lokinet-bootstrap.exe:
-	scp despair@10.10.10.6:loki-network/win32-setup/*.exe .
+	scp despair@10.10.10.3:web/loki/win32-dist/*.exe .
```
majestrate: this should be buildable for everyone not just maintainers.

despair86: yeah that was left over when i moved the prebuilt assets to a TLS server

libabyss/src/server.cpp
```diff
@@ -50,8 +56,11 @@ namespace abyss
         _conn->read   = &ConnImpl::OnRead;
         _conn->tick   = &ConnImpl::OnTick;
         _conn->closed = &ConnImpl::OnClosed;
-        m_Bad         = false;
-        m_State       = eReadHTTPMethodLine;
+#ifdef _WIN32
+        _conn->write = &TCPWrite;
+#endif
```
majestrate: this should be set by the event loop instead.

despair86: ok

despair86:
ok if we merge this we can safely get rid of the vendored libsodium
even my avx2 patches on salsa20 etc are present somehow in upstream
so i just need to patch out `undocumented function ` on win32

PR #869, opened 2019-10-21
android jni shim and vpn api for mobile

majestrate:
`make android` target should yield some `apk`s


llarp/handlers/tun.cpp
```diff
@@ -585,80 +584,120 @@ namespace llarp
...
+        llarp::LogInfo(Name(), " waiting for vpn to start");
+        vpn   = m_LazyVPNPromise.get_future().get();
+        vpnif = vpn.io;
+        if(vpnif == nullptr)
+        {
+          llarp::LogError(Name(), " failed to recieve vpn interface");
+          return false;
+        }
+        llarp::LogInfo(Name(), " got vpn interface");
+        auto self = shared_from_this();
```
notlesh: Why `shared_from_this()` instead of just capturing `this`?

majestrate: would using `this` be safe?

notlesh: I didn't dig too deep there initially, but it looks like `shared_from_this()` has the effect of tying the lifetime of `this` to the `uv_check_t` that gets injected into `libuv`.

majestrate: yes that is why i did it

jni/lokinet_jni_common.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+  const size_t length = env->GetArrayLength(stringJbytes);
+  jbyte* pBytes       = env->GetByteArrayElements(stringJbytes, NULL);
+
+  T result = visit(llarp::string_view((const char*)pBytes, length));
+
+  env->ReleaseByteArrayElements(stringJbytes, pBytes, JNI_ABORT);
+  env->DeleteLocalRef(stringJbytes);
+
+  return std::move(result);
```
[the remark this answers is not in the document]
jagerman: Yes, and in fact returning an rvalue reference like this is now a warning in some recent compiler versions.

PR #884, opened 2019-10-30
WIP: Json rpc improvements

notlesh:
This is a work in progress. The goal is to remove `libabyss` and replace it with `libjson-rpc-cpp`, restoring the original functionality along the way.


llarp/CMakeLists.txt
```diff
@@ -61,7 +61,9 @@ if(ANDROID)
   set(LOG_LIB log)
 endif()
 
-target_link_libraries(${UTIL_LIB} PUBLIC ${CRYPTOGRAPHY_LIB} ${FS_LIB} ${LOG_LIB})
+set(JSONCPP_LIBS jsoncpp microhttpd jsonrpccpp-common jsonrpccpp-server)
```
majestrate: isn't microhttpd GPL? also we already have a json parser we shouldn't pull another one in

notlesh: `microhttpd` is LGPL 2.1, and optionally "eCos License" if HTTPS/TLS is disabled. From [here](https://www.gnu.org/software/libmicrohttpd/):


> Libmicrohttpd is free software; you can redistribute it and/or modify it under the GNU LGPL v2.1 or at your option any later version. If you disable HTTPS/TLS support, you can also choose the second license, the eCos License. If you have questions about licensing, please contact the maintainer.

Agreed on the JSON parser; mixing parsers is certainly not ideal. What I think the library buys us is flexibility in the transport mechanism (HTTP/HTTPS, TCP, socket, etc.) IMO, the rest of it is of little value; implementing JSON-RPC isn't tricky, and this library doesn't do a great job at that.

While I don't want to spend much time on this, I do want something that doesn't suck, so I'm up for suggestions.

despair86:
As the guy maintaining the windows port this is already looking good

despair86:
we could just nuke `JSON for Modern C++` altogether
it doesn't even optimise that well to begin with

llarp/router/router.hpp
```diff
@@ -253,7 +253,6 @@ namespace llarp
     std::string rpcBindAddr = DefaultRPCBindAddr;
 
     /// lokid caller
-    std::unique_ptr< rpc::Caller > rpcCaller;
```
majestrate: we need this in order to talk with lokid for obtaining the service node list

PR #888, opened 2019-10-30
fix typo in cmake

majestrate:
also disable abyss-main


notlesh:
0843256 might conflict with my recent PR (#884), as it rips out libabyss. probably minor, though.

PR #899, opened 2019-11-04
make path builds work

majestrate:
bigger link messages like those for path builds are not delivered fast enough because we wait to flush the encrypt packet buffer after handling decrypted packets until we get more packets from the network.

this should fix that

### DIFF
--- a/llarp/iwp/message_buffer.cpp
+++ b/llarp/iwp/message_buffer.cpp
@@ -163,7 +163,8 @@ namespace llarp
     bool
     InboundMessage::ShouldSendACKS(llarp_time_t now) const
     {
-      return (now > m_LastACKSent && now - m_LastACKSent > 1000);
+      return (now > m_LastACKSent
+              && now - m_LastACKSent > (Session::DeliveryTimeout / 4));
     }
 
     bool
--- a/llarp/iwp/message_buffer.cpp
+++ b/llarp/iwp/message_buffer.cpp
@@ -163,8 +163,7 @@ namespace llarp
     bool
     InboundMessage::ShouldSendACKS(llarp_time_t now) const
     {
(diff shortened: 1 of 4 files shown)


majestrate:
@jagerman when this is merged please rebuild and deploy the debs for this branch so that path builds start working again, thnx.

llarp/iwp/message_buffer.cpp
```diff
@@ -163,7 +163,8 @@ namespace llarp
     bool
     InboundMessage::ShouldSendACKS(llarp_time_t now) const
     {
-      return (now > m_LastACKSent && now - m_LastACKSent > 1000);
+      return (now > m_LastACKSent
+              && now - m_LastACKSent > (Session::DeliveryTimeout / 4));
```
jagerman: Can this not be simplified to:

```C++
return now > m_LastACKSent + Session::DeliveryTimeout / 4;
```

llarp/iwp/session.cpp
```diff
@@ -596,6 +596,14 @@ namespace llarp
         }
       }
       SendMACK();
+      if(m_EncryptNext && !m_EncryptNext->empty())
+      {
+        auto self = shared_from_this();
+        m_Parent->QueueWork([self, data = std::move(m_EncryptNext)] {
```
jagerman: You can drop the `auto self = ...` and just initialize it in the lambda captures:
```C++
m_Parent->QueueWork([self = shared_from_this(), data = std::move(m_EncryptNext)] {
```
which should save one shared pointer reference increment/decrement.

llarp/iwp/message_buffer.cpp
majestrate: it can be, probably

PR #906, opened 2019-11-06
fast allocator for udp event loop

majestrate:
internal tests show significant speedups and may restore lost perf if applied to the entire network.
so far 2x network throughput if applied just locally.
gottagofast.


notlesh:
I'm able to reproduce a double-free on this branch by hitting the JSON-RPC endpoint via curl. I did this while running valgrind; the output should be attached.

[valgrind.out.txt](https://github.com/loki-project/loki-network/files/3842624/valgrind.out.txt)

PR #907, opened 2019-11-08
used lockless queues instead of shared_ptr

majestrate:


llarp/iwp/session.cpp
```diff
@@ -130,36 +146,50 @@ namespace llarp
...
     Session::EncryptAndSend(ILinkSession::Packet_t data)
     {
-      if(m_EncryptNext == nullptr)
-        m_EncryptNext = std::make_shared< CryptoQueue_t >();
-      m_EncryptNext->emplace_back(std::move(data));
-      if(!IsEstablished())
+      if(IsEstablished())
+      {
+        if(m_EncryptQueue.full())
+        {
```
jagerman: This seems like a good place for a warning when compiled in debug mode so that we can see if we are hitting this (given the `/// TODO: no idea if this is the right size` above).

PR #912, opened 2019-11-14
fix logic thread behavior

majestrate:
the logic thread seemed to work in a way that was different from the way i remember it should.

fixed.

### DIFF
--- a/llarp/ev/ev.hpp
+++ b/llarp/ev/ev.hpp
@@ -671,7 +671,8 @@ namespace llarp
 #ifdef _WIN32
 struct llarp_fd_promise
 {
-  void Set(std::pair< int, int >)
+  void
+  Set(std::pair< int, int >)
   {
   }
 
--- a/llarp/util/thread/threadpool.cpp
+++ b/llarp/util/thread/threadpool.cpp
@@ -15,18 +15,13 @@ llarp_init_threadpool(int workers, const char *name)
   return new llarp_threadpool(workers, name);
 }
 
-struct llarp_threadpool *
-llarp_init_same_process_threadpool()
-{
-  return new llarp_threadpool();
-}
-
 void
 llarp_threadpool_join(struct llarp_threadpool *pool)
 {
   llarp::LogDebug("threadpool join");
   if(pool->impl)
-    pool->impl->drain();
+    pool->impl->stop();
+  pool->impl.reset();
 }
 
 void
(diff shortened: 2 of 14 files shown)


llarp/ev/ev.hpp
```diff
@@ -671,7 +671,8 @@ namespace llarp
 #ifdef _WIN32
 struct llarp_fd_promise
 {
-  void Set(std::pair< int, int >)
+  void
```
tewinget: is `make format` capricious, or did someone forget to run it at some point?

majestrate: `make format` should use `clang-format-8` as 9 and higher makes the source get fugged some how.

llarp/util/thread/threadpool.cpp
```diff
@@ -41,64 +36,29 @@ llarp_threadpool_stop(struct llarp_threadpool *pool)
...
+bool
 llarp_threadpool_queue_job(struct llarp_threadpool *pool,
-                           std::function< void() > func)
+                           std::function< void(void) > func)
 {
-  if(pool->impl)
-  {
-    while(!pool->impl->tryAddJob(func))
-    {
-      std::this_thread::sleep_for(std::chrono::microseconds(1000));
```
notlesh: This was likely a big source of latency...

llarp/util/thread/threadpool.h
```diff
@@ -5,48 +5,51 @@
...
+
+  size_t
+  pendingJobs() const;
+
+  size_t
+  numThreads() const;
+
+  /// try to guess how big our job latency is on this threadpool
+  llarp_time_t
+  GuessJobLatency(llarp_time_t granulairty = 1000) const;
```
notlesh: A high level description of granularity would be useful. Also, typo (granulairty)

llarp/util/thread/threadpool.h
```diff
@@ -5,48 +5,51 @@
...
+
+  /// see if this thread is full given lookahead amount
+  bool
```
notlesh: Might be better named "HasSpareCapacity()" or similar

test/link/test_llarp_link.cpp
```diff
@@ -133,6 +142,9 @@ struct LinkLayerTest : public test::LlarpTest< llarp::sodium::CryptoLibSodium >
   void
   SetUp()
   {
+    oldLevel = llarp::LogContext::Instance().minLevel;
+    if(shouldDebug)
+      llarp::SetLogLevel(eLogDebug);
```
notlesh: lol...

PR #925, opened 2019-11-25
reduce number of jobs we put onto the logic thread

majestrate:
over time idle cpu use was increasing, this may help address this (or it may not).


jagerman:
I believe these changes are included in #935

Issue #927, opened 2019-11-25
make master branch build release builds by default


majestrate:
right now master defaults to debug builds, this should do release instead.

PR #930, opened 2019-11-26
make master build release by default

majestrate:


majestrate:
fixes https://github.com/loki-project/loki-network/issues/927

PR #931, opened 2019-11-26
use curl to fetch from lokid rpc the identity key

majestrate:


llarp/router/router.cpp
```diff
@@ -227,14 +229,97 @@ namespace llarp
...
+      LogError("running a service node on mobile device is not possible.");
+      return false;
+#else
+      CURL *curl = curl_easy_init();
+      if(curl)
+      {
+        CURLcode res;
+        std::stringstream ss;
+        ss << "http://" << lokidRPCAddr << "/json_rpc";
+        const auto url = ss.str();
```
jagerman: Simpler here just to use string concatenation: `const auto url = "http://" + lokidRPCAddr + "/json_rpc";`

jagerman: (Assuming lokidRPCAddr is already a `std::string`; otherwise wrap one of these in a `std::string{...}` to make it work)

llarp/router/router.cpp
```diff
@@ -227,14 +229,97 @@ namespace llarp
...
+        curl_easy_setopt(curl, CURLOPT_USERPWD, auth.c_str());
+        curl_slist *list = nullptr;
+        list = curl_slist_append(list, "Content-Type: application/json");
+        curl_easy_setopt(curl, CURLOPT_HTTPHEADER, list);
+
+        nlohmann::json request = {{"id", "0"},
+                                  {"jsonrpc", "2.0"},
+                                  {"method", "get_service_node_privkey"}};
+        const auto data        = request.dump();
+        std::vector< char > resp;
```
jagerman: Can this just be a std::string?

llarp/router/router.cpp
```diff
@@ -227,14 +229,97 @@ namespace llarp
...
+              if(not j.is_object())
+                continue;
+
+              const auto itr = j.find("result");
+              if(itr == j.end())
+                continue;
+              if(not itr->is_object())
+                continue;
+              const auto k =
+                  (*itr)["service_node_ed25519_privkey"].get< std::string >();
```
jagerman: You should `.find()` this and error out if it's not present in case someone runs it against a 5.x lokid.

llarp/router/router.cpp
```diff
@@ -227,14 +232,103 @@ namespace llarp
...
+            catch(nlohmann::json::exception &ex)
+            {
+              LogError("Bad response from lokid: ", ex.what());
+            }
+          }
+          else
+          {
+            LogError("failed to get identity Keys");
+          }
+        } while(res != CURLE_OK);
```
notlesh: Will this result in an infinite loop when we get an error?

PR #934, opened 2019-11-28
Optimizations to help the logic thread not be overwhelmed

tewinget:
This PR is a number of various commits with the goal of optimizing the logic thread usage.  There is still work to be done on this, but the commits here have made solid progress and are enough to merit PR and review before they pile up too much more.  See each commit and commit message for details.


jagerman:
I really should have used this PR rather than recreating it in #935.  Is there anything here that *isn't* in #935?

tewinget:
I don't believe so...

PR #935, opened 2019-11-29
Logic thread fixes

jagerman:
Cherry picked and cleaned up from all the various fixes we applied over the last few days on a mess of branches.

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -336,6 +336,32 @@ namespace llarp
       return msg.questions[0].IsName("localhost.loki");
     }
 
+    template <>
+    bool
+    TunEndpoint::FindAddrForIP(service::Address &addr, huint128_t ip)
+    {
+      auto itr = m_IPToAddr.find(ip);
+      if(itr != m_IPToAddr.end() and not m_SNodes[itr->second])
+      {
+        addr = service::Address(itr->second.as_array());
+        return true;
+      }
+      return false;
+    }
+
+    template <>
+    bool
+    TunEndpoint::FindAddrForIP(RouterID &addr, huint128_t ip)
+    {
+      auto itr = m_IPToAddr.find(ip);
+      if(itr != m_IPToAddr.end() and m_SNodes[itr->second])
+      {
+        addr = RouterID(itr->second.as_array());
+        return true;
+      }
+      return false;
+    }
+
     bool
     TunEndpoint::HandleHookedDNSMessage(
         dns::Message &&msg, std::function< void(dns::Message) > reply)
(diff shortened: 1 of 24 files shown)


llarp/handlers/tun.cpp
```diff
@@ -32,6 +32,13 @@ namespace llarp
       m_NetworkToUserPktQueue.Process(send);
     }
 
+    bool
+    TunEndpoint::ShouldFlushNow(llarp_time_t now) const
+    {
+      static constexpr llarp_time_t FlushInterval = 25;
```
tewinget: I wonder if this (and other constants in other files) should be at the top of the class/file so all constants for a given class are in one place.  Tough call though.

llarp/handlers/tun.cpp
```diff
@@ -35,8 +35,13 @@ namespace llarp
     void
     TunEndpoint::tunifTick(llarp_tun_io *tun)
     {
-      auto *self = static_cast< TunEndpoint * >(tun->user);
-      LogicCall(self->m_router->logic(), [self]() { self->Flush(); });
+      auto *self     = static_cast< TunEndpoint * >(tun->user);
+      const auto now = self->Now();
```
tewinget: so the libuv event loop snapshots a "now" every event loop tick that could be used here (tun->loop->whatever or something like that).  idk if the current Windows event loop supports that concept though.

llarp/util/thread/logic.cpp
```diff
@@ -90,10 +96,8 @@ namespace llarp
     }
     if(m_Thread->LooksFull(5))
     {
-      LogErrorExplicit(TAG, LINE,
-                      "holy crap, we are trying to queue a job onto the logic "
-                      "thread but "
-                      "it looks full");
+      LogErrorExplicit(TAG, LINE, "holy crap, we are trying to queue a job "
```
tewinget: ok, so since this got reformatted anyway, perhaps now is a good time to change the log message >_>

llarp/service/endpoint.cpp
```diff
@@ -805,6 +808,30 @@ namespace llarp
...
+    {
+      do
+      {
+        auto maybe = m_RecvQueue.tryPopFront();
+        if(not maybe.has_value())
+          return;
+        auto ev = std::move(maybe.value());
+        ProtocolMessage::ProcessAsync(ev.fromPath, ev.pathid, ev.msg);
+      } while(true);
+    }
```
notlesh: not a big fan of `do { ... } while`, esp. `while(true)`... just sayin'

llarp/service/protocol.cpp
```diff
@@ -380,39 +388,44 @@ namespace llarp
         dh->path = recvPath;
         return worker->addJob(std::bind(&AsyncFrameDecrypt::Work, dh));
       }
-      SharedSecret shared;
-      if(!handler->GetCachedSessionKeyFor(T, shared))
+
+      auto v = new AsyncDecrypt();
```
notlesh: Smart pointer?

llarp/service/endpoint.cpp
tewinget: It feels verbose/redundant to do the assignment twice (once before the loop, once at the end of each iteration), but I agree that it would be more clear to write the loop as `while (maybe.has_value())`.

llarp/router/router.cpp
```diff
@@ -173,9 +173,17 @@ namespace llarp
...
   Router::PumpLL()
   {
+    static constexpr size_t PumpJobThreshhold  = 50;
+    static constexpr llarp_time_t PumpInterval = 25;
+    const auto now                             = Now();
     if(_stopping.load())
       return;
-
+    if(_logic->numPendingJobs() >= PumpJobThreshhold
+       && _lastPump + PumpInterval >= now)
```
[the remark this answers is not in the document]
tewinget: There may be a more clear way to write it, but what this is saying is "if the logic thread is a certain amount of busy *and* we have pumped recently enough".

PR #943, opened 2019-12-04
Fix MacOS compilation by adding some c++17 fs modernisms to cppbackport

neuroscr:
MacOS failed compilation:

```
/Users/admin/Sites/loki-network-dev/llarp/config/key_manager.cpp:142:17: error: no viable overloaded '+='
        newPath += ext;
        ~~~~~~~ ^  ~~~
/Users/admin/Sites/loki-network-dev/llarp/config/key_manager.cpp:144:17: error: no matching function for call to 'exists'
        if (not fs::exists(newPath))
                ^~~~~~~~~~
/Users/admin/Sites/loki-network-dev/cmake/../vendor/cppbackport-master/lib/fs/filestatus.h:68:6: note: candidate function not
      viable: no known conversion from 'fs::path' to 'cpp17::filesystem::file_status' for 1st argument
bool exists(file_status);
     ^
/Users/admin/Sites/loki-network-dev/cmake/../vendor/cppbackport-master/lib/fs/filestatus.h:69:6: note: candidate function not
      viable: requires 2 arguments, but 1 was provided
bool exists(const path&, std::error_code&);
     ^
/Users/admin/Sites/loki-network-dev/llarp/config/key_manager.cpp:182:41: error: too many arguments to function call, expected 2,
      have 3
      fs::rename(filepath, newFilepath, ec);
      ~~~~~~~~~~                        ^~
/Users/admin/Sites/loki-network-dev/cmake/../vendor/cppbackport-master/lib/fs/rename.h:38:1: note: 'rename' declared here
void rename(const path& from, const path& to);
^
```

we could have just changed keymanager but I figure with Stephen's style, he'll be prone to attempting to use them again, so best to just implement them since it's easy enough.

### DIFF
--- a/vendor/cppbackport-master/lib/fs/rename.cpp
+++ b/vendor/cppbackport-master/lib/fs/rename.cpp
@@ -45,7 +45,17 @@ void rename(const path& from, const path& to)
 		if (copy_file(from, to))
 			cpp17::filesystem::remove(from);
 	}
-}
+} // end rename
+
+void rename(const path& from, const path& to, __attribute__((unused)) std::error_code& ec)
+{
+	if (std::rename(from.c_str(), to.c_str()) != 0)
+	{
+		if (copy_file(from, to))
+			cpp17::filesystem::remove(from);
+	}
+} // end rename with ec
+
 }
 }
 
--- a/llarp/config/key_manager.cpp
+++ b/llarp/config/key_manager.cpp
@@ -2,6 +2,7 @@
 
 #include <system_error>
 #include <util/logging/logger.hpp>
+#include <util/fs.hpp>
 #include "config/config.hpp"
 #include "crypto/crypto.hpp"
 #include "crypto/types.hpp"
--- a/vendor/cppbackport-master/lib/fs/filestatus.cpp
+++ b/vendor/cppbackport-master/lib/fs/filestatus.cpp
@@ -182,6 +182,12 @@ namespace cpp17
     }
 
     bool
+    exists(const path& p)
+    {
+      return exists(status(p));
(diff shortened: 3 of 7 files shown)


vendor/cppbackport-master/lib/fs/rename.cpp
```diff
@@ -45,7 +45,17 @@ void rename(const path& from, const path& to)
 		if (copy_file(from, to))
 			cpp17::filesystem::remove(from);
 	}
-}
+} // end rename
+
+void rename(const path& from, const path& to, __attribute__((unused)) std::error_code& ec)
```
majestrate: what is the point of this function providing a `std::error_code` if it doesn't even set it on error?

vendor/cppbackport-master/lib/fs/rename.cpp
```diff
@@ -45,7 +45,17 @@ void rename(const path& from, const path& to)
...
+{
+	if (std::rename(from.c_str(), to.c_str()) != 0)
```
majestrate: this is literally the same function as the other one. why not just call that, wrap it with try/catch and set error code if it throws?

neuroscr: in case someone does want to implement std:error_code

despair86:
ah yeah
BSDs and Sys5 are complaining

despair86:
workaround: compile with `g++` or `eg++` whatever your platform ships in ports/pkgsrc/homebrew, G++ builds full C++17 as much as possible

jagerman:
Superceded by #952

PR #952, opened 2019-12-06
Replace cppbackport with ghc-filesystem

jagerman:
From https://github.com/gulrak/filesystem which is more up-to-date and looks better maintained than cppbackport.


cmake/unix.cmake
```diff
@@ -29,15 +29,20 @@ if(EMBEDDED_CFG OR ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   link_libatomic()
 endif()
 
+function(use_vendored_filesystem)
+  include_directories("${CMAKE_CURRENT_LIST_DIR}/../vendor/ghc-filesystem/include")
+  add_definitions(-DLOKINET_USE_GHC_FILESYSTEM)
+  set(FS_LIB)
```
despair86: set this to some dummy value

cmake/unix.cmake
```diff
@@ -29,15 +29,20 @@ if(EMBEDDED_CFG OR ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   link_libatomic()
 endif()
 
+function(use_vendored_filesystem)
+  include_directories("${CMAKE_CURRENT_LIST_DIR}/../vendor/ghc-filesystem/include")
+  add_definitions(-DLOKINET_USE_GHC_FILESYSTEM)
+  set(FS_LIB "" PARENT_SCOPE)
```
despair86: `set(FS_LIB ghc_filesystem)`, it is defined....

vendor/ghc-filesystem/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+endif()
+if(CMAKE_CXX_STANDARD LESS 11)
+    message(FATAL_ERROR "CMAKE_CXX_STANDARD is less than 11, ghc::filesystem only works with C++11 and above.")
+endif()
+
+add_library(ghc_filesystem INTERFACE)
```
despair86: ...here @jagerman

jagerman: Yes, I saw that, but I'm not loading that CMakeLists.txt file currently (and don't want to because it'll break compatibility with xenial because of the cmake version requirement at the top).

PR #954, opened 2019-12-09
Reduce log levels at runtime

majestrate:


llarp/util/logging/logger.hpp
```diff
@@ -83,14 +83,22 @@ namespace llarp
...
     const llarp_time_t started;
 
     static LogContext&
     Instance();
+
+    void
+    DropToRuntimeLevel();
+
+    void
+    RevertRuntimeLevel();
```
notlesh: The convention "startup level" and "runtime level" is very clear to me, this function name seems ambiguous though

PR #956, opened 2019-12-09
Implement KeyManager tests

notlesh:


test/test_util.hpp
```diff
@@ -25,9 +25,9 @@ namespace llarp
 
     struct FileGuard
     {
-      const fs::path &p;
+      const fs::path p;
```
notlesh: Note: I'm assuming this was an optimization, which isn't really necessary for unit testing. It was also restrictive in the way it can be used.

test/crypto/test_llarp_key_manager.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+};
+
+TEST_F(KeyManagerTest, EnsureDefaultConfNames)
+{
+  llarp::Config conf;
+
+  // the default config filenames will suffice, this exists as sanity check to
+  // protect against the assumptions made below
+  ASSERT_TRUE(conf.router.ourRcFile() == rcFile);
```
majestrate: `ASSERT_EQ` is a better choice here

notlesh: Good point -- done

PR #962, opened 2019-12-11
win32 tun and libuv for win32

majestrate:


despair86:
very well then. I'll have a libuv build patch for pre-6.0 machines soon:tm:
> current Windows XP (and 5.x at large) share is still 5% in 2019

despair86:
I also have a patch that builds libuv in the `win32-setup` folder along with libcurl etc so you can set `LIBUV_PREFIX` to `win32-setup`

PR #969, opened 2019-12-14
increase ack window

majestrate:
reduces traffic overhead from retransmissions between nodes.

### DIFF
--- a/llarp/iwp/message_buffer.cpp
+++ b/llarp/iwp/message_buffer.cpp
@@ -165,7 +165,7 @@ namespace llarp
     bool
     InboundMessage::ShouldSendACKS(llarp_time_t now) const
     {
-      return now > m_LastACKSent + (Session::DeliveryTimeout / 2);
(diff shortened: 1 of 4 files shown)


llarp/iwp/message_buffer.cpp
```diff
@@ -165,7 +165,7 @@ namespace llarp
     bool
     InboundMessage::ShouldSendACKS(llarp_time_t now) const
     {
-      return now > m_LastACKSent + (Session::DeliveryTimeout / 4);
+      return now > m_LastACKSent + (Session::DeliveryTimeout / 2);
```
jagerman: This `Session::DeliveryTimeout / 2` value should really be another static constexpr value in llarp/iwp/session.hpp.

Actually, it already *IS* a constant in session.hpp, `ACKResendInterval`, but isn't being used anywhere.  Change it there, and use it here.

(Also just noticed there's a typo in one of those constants in llarp/iwp/session.hpp: `Recieval` should be `Receival`)

PR #994, opened 2019-12-19
try to fix gitlab CI (redo 991)

neuroscr:
change to xenial
add debug/release versions
include lokinetctl
use ninja like travis
disable freebsd (due to newer dependencies)


majestrate:
looks like gitlab passes now

PR #1005, opened 2019-12-26
make static linking actually static

majestrate:


cmake/unix.cmake
```diff
@@ -18,7 +18,7 @@ if(NOT IOS AND NOT ANDROID AND NOT WIN32)
     set(CURL_CA_PATH none)
     message(STATUS "using local curl: ${LIBCURL_ROOT}")
     add_subdirectory(${LIBCURL_ROOT})
-    set(CURL_INCLUDE_DIRS ${LIBCURL_ROOT}/include)
+    set(CURL_INCLUDE_DIR ${LIBCURL_ROOT}/include)
```
jagerman: This should stay as plural because `FindCURL` sets it CURL_INCLUDE_DIRS (and if it comes that way it can potentially contain multiple dirs, plus agrees with how other cmake library finding code handles things).

llarp/CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ add_library(${UTIL_LIB} STATIC ${LIB_UTIL_SRC})
 
 add_dependencies(${UTIL_LIB} genversion)
 
-target_include_directories(${UTIL_LIB} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${PROJECT_SOURCE_DIR}/include)
+target_include_directories(${UTIL_LIB} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${PROJECT_SOURCE_DIR}/include ${CURL_INCLUDE_DIR})
```
jagerman: Likewise

llarp/CMakeLists.txt
```diff
@@ -252,7 +252,7 @@ if(TESTNET)
 endif()
 
 add_library(${STATIC_LIB} STATIC ${LIB_SRC})
-target_include_directories(${STATIC_LIB} PUBLIC ${CURL_INCLUDE_DIRS})
+target_include_directories(${STATIC_LIB} PUBLIC ${CURL_INCLUDE_DIR})
```
jagerman: Likewise.

Makefile
```diff
@@ -212,7 +215,16 @@ $(TEST_EXE): debug
...
+static-configure: $(LIBUV_PREFIX) $(LIBCURL_PREFIX)
+	(test x$(TOOLCHAIN) = x && $(CONFIG_CMD) -DCMAKE_BUILD_TYPE=Release -DSTATIC_LINK=ON -DRELEASE_MOTTO="$(shell cat motto.txt)" -DCMAKE_C_FLAGS='$(CFLAGS)' -DCMAKE_CXX_FLAGS='$(CXXFLAGS)' -DLIBUV_ROOT='$(LIBUV_PREFIX)' -DLIBCURL_ROOT='$(LIBCURL_PREFIX)' ) || (test x$(TOOLCHAIN) != x && $(CONFIG_CMD) -DCMAKE_BUILD_TYPE=Release -DSTATIC_LINK=ON -DRELEASE_MOTTO="$(shell cat motto.txt)" -DCMAKE_C_FLAGS='$(CFLAGS)' -DCMAKE_CXX_FLAGS='$(CXXFLAGS)' -DLIBUV_ROOT='$(LIBUV_PREFIX)' -DLIBCURL_ROOT='$(LIBCURL_PREFIX)' -DCMAKE_TOOLCHAIN_FILE=$(TOOLCHAIN) -DNATIVE_BUILD=OFF )
+
+static: static-configure
+	$(MAKE) -C '$(BUILD_ROOT)'
+	cp $(EXE) $(REPO)/lokinet-static
 
+$(LIBCURL_PREFIX):
+	mkdir -p $(BUILD_ROOT)
+	git clone -b "$(LIBCURL_VERSION)" https://github.com/curl/curl "$(LIBCURL_PREFIX)"
```
jagerman: It's probably quite a bit faster to download `https://github.com/curl/curl/releases/download/curl-7_67_0/curl-7.67.0.tar.xz`, which doesn't need to download the full repo with history but just the release code.

readme.md
```diff
@@ -93,6 +93,17 @@ alternatively make a debian package with:
 
 this puts the built packages in `../`
 
+
+#### Static Linux
+
+static native optimized:
+
+    $ make static STATIC_LINK=ON DOWNLOAD_SODIUM=ON
```
jagerman: Does this work if I have libsodium-dev installed?  I thought that DOWNLOAD_SODIUM is more of an "it's okay to download if you need to" option.  Perhaps we need some sort of `DOWNLOAD_SODIUM=ALWAYS` setting for it?

majestrate: libsodium will be statically linked if installed the `DOWNLOAD_SODIUM` part is probably not needed in this case.

PR #1006, opened 2019-12-27
prevent crash on exit.

majestrate:
dont close timers on closeall to prevent libuv abort on exit since closing the event loop also closes timers, who knew!?

### DIFF
--- a/llarp/ev/ev_libuv.cpp
+++ b/llarp/ev/ev_libuv.cpp
@@ -1007,10 +1007,6 @@ namespace libuv
           {
             static_cast< glue* >(h->data)->Close();
           }
-          else if(h->type == UV_TIMER)
-          {
-            CloseUVTimer((uv_timer_t*)h);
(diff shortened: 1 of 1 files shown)


tewinget:
I don't know why libuv would be aborting on close just because the timers it was using are already closed.  That said, if this fixes it and the timers will be closed anyway, then I'm all for it as it makes the code marginally simpler anyway.  Though there is an argument to be made for explicitly stopping the timers...

majestrate:
On Fri, Dec 27, 2019 at 10:33:58AM -0800, Thomas Winget wrote:
> I don't know why libuv would be aborting on close just because the timers it was using are already closed.  That said, if this fixes it and the timers will be closed anyway, then I'm all for it as it makes the code marginally simpler anyway.  Though there is an argument to be made for explicitly stopping the timers...
yeah idk why either but this makes it not broke.

PR #1008, opened 2019-12-28
remove -e flag from shebang

### LINKED ISSUE #1007: Debian: Installing on NFS root fails
**Describe the bug**
on my debian (technically raspbian) NFS root installation lokinet-bin package fails to install
I think issue is in https://github.com/loki-project/loki-network/blob/master/debian/postinst script, maybe it needs `exit 0` at the end, or `-e` flag removed.

**To Reproduce**
have debian install ontop of NFS
try installing lokinet from official lokinet-provided debian repo

**Screenshots Or Logs**
```
Setting up lokinet-bin (0.6.1-1~deb10) ...
adduser: Warning: The home directory `/var/lib/lokinet' does not belong to the user you are currently creating.
Failed to set capabilities on file `/usr/bin/lokinet' (Operation not supported)
The value of the capability argument is not permitted for a file. Or the file is not a regular (non-symlink) file
failed to setcap lokinet
dpkg: error processing package lokinet-bin (--configure):
 installed lokinet-bin package post-installation script subprocess returned error exit status 1
dpkg: dependency problems prevent configuration of lokinet:
 lokinet depends on lokinet-bin (= 0.6.1-1~deb10); however:
  Package lokinet-bin is not configured yet.

dpkg: error processing package lokinet (--configure):
 dependency problems - leaving unconfigured
```

 - OS: Debian/Raspbian on NFS
-  Device: irrelevant, should affect any NFS-rooted installation
 - Lokinet Version number or Git commit hash: 0.6.1

majestrate:
fix #1007

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,3 +1,4 @@
 #!/bin/sh
 setcap cap_net_admin,cap_net_bind_service=+eip /usr/bin/lokinet || echo "failed to setcap lokinet"
 adduser --quiet --system --home /var/lib/lokinet debian-lokinet || echo "debian-lokinet user not added"
+exit 0
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,3 +1,3 @@
-#!/bin/sh -e
+#!/bin/sh
 setcap cap_net_admin,cap_net_bind_service=+eip /usr/bin/lokinet || echo "failed to setcap lokinet"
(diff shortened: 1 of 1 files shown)


cathugger:
I think adding `exit 0` might be even more important for this, but removing `-e` makes sense too as it doesn't really use it.

PR #1015, opened 2019-12-30
Fast forward dev

majestrate:


tewinget:
commits which were pushed directly to master are reflected here.  in future, the *only* case where this should happen is for critical security patches which happen outside the dev->master development cycle.

PR #1017, opened 2019-12-31
IX86 builds should only require SSE3 by default

despair86:
this still allows Debian pkg builds to use their own target specfic flags


despair86:
this also leaves `NATIVE_BUILD` untouched which allows source builds to be machine specific

despair86:
`-march=nocona` is generally a synonym for `-march=x86-64` as it is the oldest EM64T CPU without AMD64 exclusive instructionss

PR #1019, opened 2020-01-02
Dht spider code

majestrate:


jagerman:
Needs some user documentation.  What does it do, how is it called?

majestrate:
i'll document it later, just wanted to get it out there.

PR #1023, opened 2020-01-06
add bw stats to iwp sessions

notlesh:
Changes to report IWP session stats in the `dumpstate` API call. Used by the GUI to display instantaneous bandwidth usage.


majestrate:
for `dev` branch plz

notlesh:
> apply this to `dev` instead

Wish that was the default -- is there a way to change that? PR into `master` should definitely be a special occasion...

PR #1026, opened 2020-01-08
lokinet qt5 ui

despair86:
and make conditional on platform support (windows vista or above)


majestrate:
1) commit to `dev` plz
2) we are probably going to gut all of the ui stuff in this repo in favor of an external repo for ui jizz
3) we are probably going to use qt infra for installers too

PR #1027, opened 2020-01-08
prepare to fork installers

despair86:
probably the last time i touch this one for a while since you want to use QT and its godawful installer kit

that way we just build the 6.0+ installer with qt and build old installers as needed


win32-setup/Makefile
```diff
@@ -75,7 +75,10 @@ libuv:
 	make -C libuv -j4 distclean
 	cd libuv; ./configure --host=x86_64-w64-mingw32 --target=x86_64-w64-mingw32 CC=$(CC64) CXX=$(CXX64) CFLAGS="$(CFLAGS)" --disable-shared
 	make -C libuv -j4
-	cp libuv/.libs/libuv.a $(PWD)/lib64
+	cp libuv/.libs/libuv.a $(PWD)/lib64
+
+qt5-ui:
+	wget https://snowlight.net/loki/win32-dist/lokinet-qt5-ui.7z
```
notlesh: We should definitely pull this from github -- what is snowlight anyway?

despair86:
Closing for now, will maintain a fork (like jason does for debian packaging) while i sort this out and other windows specific enhancements for this cycle

PR #1031, opened 2020-01-09
Fix avx2 cpu detection on amd64 (again)

jagerman:
The optional runtime detection code also shouldn't rely on being compiled in avx2 mode.


majestrate:
use `dev` branch :^)

PR #1035, opened 2020-01-10
win32 installer using cpack

majestrate:
Let CI Pass first


despair86:
i still want to pop out inno setup installers for signed releases, and keep the automated builds for _everything_ in between

PR #1036, opened 2020-01-12
reduce disk io

majestrate:


llarp/nodedb.hpp
```diff
@@ -87,9 +87,10 @@ struct llarp_nodedb
   std::string
   getRCFilePath(const llarp::RouterID &pubkey) const;
 
-  /// insert and write to disk
+  /// insert and optionally write to disk
   bool
-  Insert(const llarp::RouterContact &rc) LOCKS_EXCLUDED(access);
+  Insert(const llarp::RouterContact &rc, bool writeToDisk = false)
+      LOCKS_EXCLUDED(access);
```
notlesh: So when is this ever called with `writeToDisk == true`?

jagerman: To be addressed in another PR.

PR #1037, opened 2020-01-13
enable aggressive client profiling

majestrate:
* builds off #1036 
* introduces much more aggressive router profiling when running as client


tewinget:
note for in the morning: latter two commits (b1b00cc and a38c333) should be a separate PR.

majestrate:
i dont think this one is needed anymore because #1036 and #1038 cover it

PR #1038, opened 2020-01-14
disable iterative behavior in DHT

majestrate:
this makes path build not suck and allows the network to full mesh

### DIFF
--- a/llarp/dht/context.cpp
+++ b/llarp/dht/context.cpp
@@ -20,7 +20,7 @@
 #include <util/thread/logic.hpp>
 #include <nodedb.hpp>
 #include <profiling.hpp>
-
+#include <router/i_rc_lookup_handler.hpp>
 #include <vector>
 
 namespace llarp
@@ -41,7 +41,7 @@ namespace llarp
       void
       StoreRC(const RouterContact rc) const override
       {
-        GetRouter()->nodedb()->InsertAsync(rc);
+        GetRouter()->rcLookupHandler().CheckRC(rc);
       }
 
       /// on behalf of whoasked request introset for target from dht router with
--- a/llarp/dht/localrouterlookup.cpp
+++ b/llarp/dht/localrouterlookup.cpp
@@ -43,7 +43,11 @@ namespace llarp
             found = rc;
         }
         valuesFound.clear();
-        valuesFound.emplace_back(found);
+        if(not found.pubkey.IsZero())
+        {
+          valuesFound.resize(1);
+          valuesFound[0] = found;
+        }
       }
       routing::DHTMessage msg;
       msg.M.emplace_back(new GotRouterMessage(parent->OurKey(), whoasked.txid,
--- a/llarp/dht/messages/gotrouter.cpp
+++ b/llarp/dht/messages/gotrouter.cpp
@@ -113,6 +113,8 @@ namespace llarp
         LogDebug("got ", R.size(), " results in GRM for lookup");
         if(R.size() == 0)
           dht.pendingRouterLookups().NotFound(owner, K);
+        else if(R[0].pubkey.IsZero())
+          return false;
         else
           dht.pendingRouterLookups().Found(owner, R[0].pubkey, R);
         return true;
--- a/llarp/dht/messages/gotrouter.cpp
+++ b/llarp/dht/messages/gotrouter.cpp
@@ -4,6 +4,7 @@
 #include <memory>
 #include <path/path_context.hpp>
 #include <router/abstractrouter.hpp>
+#include <router/i_rc_lookup_handler.hpp>
 
 namespace llarp
 {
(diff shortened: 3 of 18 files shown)


llarp/dht/messages/gotrouter.cpp
```diff
@@ -116,8 +117,13 @@ namespace llarp
...
       }
-      llarp::LogWarn("Unwarranted GRM from ", From, " txid=", txid);
-      return false;
+      // store if valid
+      for(const auto &rc : R)
+      {
+        if(not dht.GetRouter()->rcLookupHandler().CheckRC(rc))
+          return false;
+      }
+      return true;
```
tewinget: confirm if changing this return value is correct.

llarp/dht/context.cpp
```diff
@@ -41,7 +41,7 @@ namespace llarp
       void
       StoreRC(const RouterContact rc) const override
```
tewinget: As per discussion, this function can probably go away entirely, as the calling point should no longer need it.

llarp/router/rc_lookup_handler.cpp
```diff
@@ -235,6 +238,26 @@ namespace llarp
...
     }
 
+    if(useWhitelist)
+    {
+      std::set< RouterID > lookupRouters;
+      {
+        static constexpr size_t LookupPerTick = 25;
+        // if we are using a whitelist look up a few routers we don't have
+        util::Lock l(&_mutex);
+        for(const auto &r : whitelistRouters)
```
tewinget: consider starting at a random index rather than iterating, so that each time we do this lookup loop we're probably looking up a different set (we'll eventually get back to any we missed, so may as well not try them multiple times in a row)

jagerman: Should this also be looking up routers even if we don't have them in the nodedb to refresh the RCs (i.e. to keep them from going stale and expiring)?

tewinget: So that was supposed to be happening anyway, but I derped a bit when I refactored that bit out of Router.  It's fixed in this PR as well, see changes to rc_lookup_handler.cpp

llarp/dht/messages/gotrouter.hpp
```diff
@@ -18,8 +18,9 @@ namespace llarp
       {
       }
       GotRouterMessage(const Key_t& from, uint64_t id,
-                       std::vector< RouterContact > results, bool tunneled)
-          : IMessage(from), R(std::move(results)), txid(id), relayed(tunneled)
+                       const std::vector< RouterContact >& results,
+                       bool tunneled)
+          : IMessage(from), R(results), txid(id), relayed(tunneled)
       {
```
notlesh: I really think these are effectively equivalent in terms of safety. The former version is better, however, because the caller could `std::move()` their `std::vector<RouterContact>` which would avoid any copy at all. The latter version requires a minimum one copy in all cases.

llarp/service/endpoint.cpp
```diff
@@ -733,11 +733,11 @@ namespace llarp
...
-        job->nodedb       = Router()->nodedb();
-        job->cryptoworker = Router()->threadpool();
-        job->diskworker   = Router()->diskworker();
-        job->logic        = Router()->logic();
+        llarp_async_verify_rc* job = new llarp_async_verify_rc();
+        job->nodedb                = Router()->nodedb();
+        job->cryptoworker          = Router()->threadpool();
+        job->diskworker            = Router()->diskworker();
+        job->logic                 = Router()->logic();
         job->hook = std::bind(&Endpoint::HandleVerifyGotRouter, this, msg,
```
notlesh: Is this change for clarity, or is does it actually change behavior?

llarp/dht/localrouterlookup.cpp
```diff
@@ -43,7 +43,11 @@ namespace llarp
             found = rc;
         }
         valuesFound.clear();
-        valuesFound.emplace_back(found);
+        if(not found.pubkey.IsZero())
+        {
+          valuesFound.resize(1);
+          valuesFound[0] = found;
+        }
       }
```
notlesh: Maybe a log statement here? This isn't a case we should be hitting, so it deserves attention when we do.

llarp/service/endpoint.cpp
majestrate: code clarity.

llarp/dht/messages/gotrouter.hpp
jagerman: Yeah, the original guarantees at most one copy, the new one can have two.

llarp/dht/localrouterlookup.cpp
notlesh: Added in #1041

llarp/router/rc_lookup_handler.cpp
notlesh: Randomized in #1041

PR #1046, opened 2020-01-16
the second half of dht fixes i thought i got merged but weren't for some reason.

majestrate:
mostly client side fixes.


majestrate:
(not needed for v0.6.4 relay fixes)

PR #1050, opened 2020-01-16
Tweak limits

majestrate:


tewinget:
should be fine, pending confirmation this makes the network less 'splody.

llarp/router/router.cpp
```diff
@@ -697,7 +697,10 @@ namespace llarp
       connected += _linkManager.NumberOfPendingConnections();
     }
 
-    _rcLookupHandler.ExploreNetwork();
+    const int interval = isSvcNode ? 30 : 5;
+
+    if((Uptime() / 1000) % interval == 0)
```
jagerman: This has the potential to miss lookups, for example if its called with Uptime() = 29999 on one call and Uptime() = 31000 on the next call then it won't get called on either one.

majestrate: that's fine

llarp/router/rc_lookup_handler.cpp
```diff
@@ -248,7 +248,7 @@ namespace llarp
 
     if(useWhitelist)
     {
-      static constexpr size_t LookupPerTick = 25;
+      static constexpr size_t LookupPerTick = 5;
```
jagerman: Rename to "LookupPerCall" or something (since we don't call this every tick now).

jagerman: Scratch that.  Instead let's change this logic to still call this explore function every tick but do nothing on 4 out of 5 ticks if a service node:

```C++
int _explore_tick = 0; // (member variable)
```

```C++
if (isSvcNode) {
  // Service nodes explore 1 node every 5 ticks
  if (++_explore_tick < 5)
    return;
  else
    _explore_tick = 0;
}
// else client: explore 1 node every tick
```

llarp/router/router.cpp
```diff
@@ -697,7 +697,10 @@ namespace llarp
       connected += _linkManager.NumberOfPendingConnections();
     }
 
-    _rcLookupHandler.ExploreNetwork();
+    const int interval = isSvcNode ? 30 : 5;
```
jagerman: Instead of this we can keep the unconditional ExploreNetwork() here, and then just skip ticks inside ExploreNetwork to throttle lookups (but still distribute them over time).  (See my comments above).

PR #1051, opened 2020-01-17
CI Updates, Build linux static tarballs automation

neuroscr:
- bring gitlab ci up to date, have it make a static linux release tarball
- update docker files libsodium requirements, so they build in travis-ci again


.gitlab-ci.yml
```diff
@@ -5,34 +5,44 @@ stages:
   - build
   - test
 
-# disabled because cannot find -luv -lcap when building
+# disabled because gcc9 can't build net_if.hpp
```
jagerman: Is this referring to the xenial bug with conflicting net/if.h and linux/if.h includes?  That should be fixed now.

neuroscr: related to those changes, gcc9 gives this:
```
In file included from /builds/neuroscr/loki-network/llarp/ev/ev.h:15,
                  from /builds/neuroscr/loki-network/llarp/ev/ev.hpp:5,
                  from /builds/neuroscr/loki-network/llarp/util/thread/logic.hpp:4,
                  from /builds/neuroscr/loki-network/llarp/util/thread/logic.cpp:1:
 /builds/neuroscr/loki-network/llarp/net/net_if.hpp:8:38: error: expected initializer before '__THROW'
     8 | if_nametoindex(const char* __ifname) __THROW;
       |                                      ^~~~~~~
 make[3]: *** [llarp/CMakeFiles/lokinet-util.dir/build.make:666: llarp/CMakeFiles/lokinet-util.dir/util/thread/logic.cpp.o] Error 1
```

linux-setup/readme.txt
```diff
@@ -0,0 +1,4 @@
+You'll need to setcap the lokinet to make sure you don't have to run it as root
+On debian-based distros, you make sure you have setcap installed first: apt install libcap2-bin
+and then you can:
+setcap cap_net_admin,cap_net_bind_service=+eip lokinet
```
jagerman: We aren't doing this anymore for the debs because it breaks some setups (such as having a NFS root); instead we add capabilities in the service file.

.gitlab-ci.yml
```diff
@@ -5,34 +5,44 @@ stages:
...
+    # alternatively: git describe --tag
+    - LOKINET_VERSION=$(./lokinet-static --version|sed s/lokinet-//)
+    - echo lokinet-linux-x64-v$LOKINET_VERSION
+    - mkdir lokinet-linux-x64-v$LOKINET_VERSION
+    - cp lokinet-bootstrap lokinet-linux-x64-v$LOKINET_VERSION
+    - cp build/daemon/lokinet lokinet-linux-x64-v$LOKINET_VERSION
+    - cp build/daemon/lokinetctl lokinet-linux-x64-v$LOKINET_VERSION
+    - cp linux-setup/readme.txt lokinet-linux-x64-v$LOKINET_VERSION
+    - tar cf lokinet-linux-x64-v$LOKINET_VERSION.tar lokinet-linux-x64-v$LOKINET_VERSION
+    - xz lokinet-linux-x64-v$LOKINET_VERSION.tar
```
jagerman: `tar cJf lokinet-linux-x64-v$LOKINET_VERSION.tar.xz lokinet-linux-x64-v$LOKINET_VERSION`

.gitlab-ci.yml
```diff
@@ -5,34 +5,44 @@ stages:
...
-#    paths:
-#      - "build/daemon/lokinet"
-#      - "build/daemon/lokinetctl"
+build:linux_release:
+  image: ubuntu:xenial
+  tags:
+    - linux
+  stage: build
+  before_script:
+    - apt-get update && apt-get install -y binutils-gold build-essential bsdmainutils ca-certificates cmake curl git libcap-dev pkg-config wget xz-utils libuv1-dev libsodium-dev libcurl4-openssl-dev
```
jagerman: Are all of these needed? libsodium-dev, for instance, doesn't seem necessary.

.gitlab-ci.yml
```diff
@@ -5,34 +5,44 @@ stages:
...
+  script:
+    - make static DOWNLOAD_SODIUM=ON
+    # alternatively: git describe --tag
+    - LOKINET_VERSION=$(./lokinet-static --version|sed s/lokinet-//)
+    - echo lokinet-linux-x64-v$LOKINET_VERSION
```
jagerman: Debugging?

linux-setup/readme.txt
neuroscr: I'm aware, this text was created in late December when I was told I shouldn't add a shell script for installation, that a text file was preferred. My goal here is to get it included into the Repo, so we can all make edits and improve it (not be a file that's only on my containers). We can argue over the contents thru PRs that attempt to make further edits to it.

My personal thought is for those that do have setcap, this seems like a better flow. And not all distros have systemd. I could see including some systemd service files in `contrib/` and also providing some instructions in this file on how to install and use those.

.gitlab-ci.yml
```diff
@@ -5,34 +5,44 @@ stages:
...
+build:linux_release:
+  image: ubuntu:xenial
+  tags:
+    - linux
+  stage: build
+  before_script:
+    - apt-get update && apt-get install -y binutils-gold build-essential bsdmainutils ca-certificates cmake curl git libcap-dev pkg-config wget xz-utils libuv1-dev libsodium-dev libcurl4-openssl-dev
+  script:
+    - make static DOWNLOAD_SODIUM=ON
+    # alternatively: git describe --tag
```
jagerman: I don't think that would work - most of the time won't be building a tag.

.gitlab-ci.yml
neuroscr: Yea verifying assumptions we make about `--version` if something breaks, this is a quick way to see the output. I see no harm in being there.

.gitlab-ci.yml
```diff
@@ -5,34 +5,44 @@ stages:
...
+    - LOKINET_VERSION=$(./lokinet-static --version|sed s/lokinet-//)
+    - echo lokinet-linux-x64-v$LOKINET_VERSION
+    - mkdir lokinet-linux-x64-v$LOKINET_VERSION
+    - cp lokinet-bootstrap lokinet-linux-x64-v$LOKINET_VERSION
+    - cp build/daemon/lokinet lokinet-linux-x64-v$LOKINET_VERSION
+    - cp build/daemon/lokinetctl lokinet-linux-x64-v$LOKINET_VERSION
+    - cp linux-setup/readme.txt lokinet-linux-x64-v$LOKINET_VERSION
```
jagerman: Are these already root owned in the build?

.gitlab-ci.yml
neuroscr: probably not I think `binutils-gold` and/or `bsdmainutils` was from docker, just remnants of history of our CI/docker usage. I was trying to keep the configs in travis and gitlab consistent for a bit, so that artifact made in gitlab would at least match the travis tests. 

But sounds like you're thinking what I was, and I was on the edge, so I'll just go ahead and remove them.

.gitlab-ci.yml
neuroscr: It was just an option @despair86 recommended. It seemed to grab a valid tag set from my experimentation used with it. Returns `v0.6.2-38-gc6fa014a` for my branch currently (which isn't tagged). Found a webpage that describes it as 
> `git describe` command finds the most recent tag that is reachable from a commit

and 

> --tags
> Instead of using only the annotated tags, use any tag found in refs/tags namespace. This option enables matching a lightweight (non-annotated) tag.

from https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---tags

I'd suggest to move the version number into a text file that the build system can inject like the motto. This would simplify version management for build pipelines. @despair86  said that would be fine with his Windows requirements too. But needs further discussion/buy-in with the rest of the team.

.gitlab-ci.yml
neuroscr: Yes:
```
$ tar tvf lokinet-linux-x64-v0.6.3-20d468e8e.tar.xz 
drwxr-xr-x  0 root   root        0 Jan 16 18:10 lokinet-linux-x64-v0.6.3-20d468e8e/
-rwxr-xr-x  0 root   root     1661 Jan 16 18:10 lokinet-linux-x64-v0.6.3-20d468e8e/lokinet-bootstrap
-rwxr-xr-x  0 root   root  6318176 Jan 16 18:10 lokinet-linux-x64-v0.6.3-20d468e8e/lokinet
-rwxr-xr-x  0 root   root  3446640 Jan 16 18:10 lokinet-linux-x64-v0.6.3-20d468e8e/lokinetctl
```

linux-setup/readme.txt
jagerman: Okay, here's some suggested wording:

---
Lokinet needs certain capabilities to run to set up a virtual network interface and provide a DNS server.  The preferred approach to using this is through the linux capabilities mechanism, which allows assigning limited capabilities without needing to run the entire process as root.

There are two main ways to do this:

1. If you are running lokinet via an init system such as systemd, you can specify the capabilities in the service file by adding:

CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_BIND_SERVICE
AmbientCapabilities=CAP_NET_ADMIN CAP_NET_BIND_SERVICE

into the `[Service]` section of the systemd service file.  This will assign the necessary permissions when running the process and allow lokinet to work while running as a non-root user.

2. You can set the capabilities on the binary by using the `setcap` program (if not available you may need to install `libcap2-bin` on Debian/Ubuntu-based systems) and running:

setcap cap_net_admin,cap_net_bind_service=+eip lokinet

This grants the permissions whenever the `lokinet` binary is executed.

.gitlab-ci.yml
jagerman: That's rather interesting, I had no idea that command existed.

jagerman:
I'd love to get something like this that pops out Windows installers for each commit, too!

neuroscr:
Yea, I was talking with @despair86 but looks like I have to iterate on that through an open PR (flooding it with commits) against this main repo (because our windows ci-runner is tied to the main repo and I can't point it to my personal that I used to test this PR). Not sure if this PR is the best for that, please advise.

jagerman:
No, you're right, it should definitely be in a separate PR.  I meant to tag @despair86 in my comment and forgot.

linux-setup/readme.txt
```diff
@@ -0,0 +1,16 @@
+Lokinet needs certain capabilities to run to set up a virtual network interface and provide a DNS server. The preferred approach to using this is through the linux capabilities mechanism, which allows assigning limited capabilities without needing to run the entire process as root.
```
jagerman: Can you move this file to something docs, perhaps `docs/linux-static-readme.txt` rather than creating a linux-setup top-level dir.

.gitlab-ci.yml
jagerman: This is probably not gcc 9 but rather alpine's libc.

linux-setup/readme.txt
neuroscr: Just keeping inline with the Mac and Windows distribution folder system we have x-setup

.gitlab-ci.yml
neuroscr: Well is it forward looking? Do we care about alpine? I can see use caring about making sure our code base works with newer gccs but not sure about how libc works irt this

linux-setup/readme.txt
jagerman: Unlike the macos and windows, this is just a bit of documentation which is *also* useful for people building from source on linux, while the contents of the setup-* have patches and scripts needed to massage the build.

neuroscr: Ok will move later

neuroscr:
They're basically notes on how to fix them from what I learned fixing the others. They're not committed elsewhere and we'll want them when we try to enable those.

PR #1052, opened 2020-01-17
Enable gitlab windows-ci

neuroscr:


neuroscr:
does not work like I thought

PR #1054, opened 2020-01-17
Include outbound message queue stats in dumpState API response

notlesh:
Adds stats about the outbound message queue, particularly designed to help understand where we are hitting self-imposed limits.

@tewinget and @majestrate , please review with consideration of threading safety.

### DIFF
--- a/llarp/router/outbound_message_handler.cpp
+++ b/llarp/router/outbound_message_handler.cpp
@@ -89,7 +89,17 @@ namespace llarp
   util::StatusObject
   OutboundMessageHandler::ExtractStatus() const
   {
-    util::StatusObject status{};
+    util::StatusObject status{ 
+      "queueStats", {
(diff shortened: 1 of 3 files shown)


llarp/router/outbound_message_handler.cpp
```diff
@@ -349,6 +374,10 @@ namespace llarp
         break;
       }
     }
+
+    if (sent_count > m_queueStats.perTickMax)
```
majestrate: std::max ?

llarp/router/outbound_message_handler.cpp
```diff
@@ -201,6 +208,7 @@ namespace llarp
   {
     const llarp_buffer_t buf(msg.first);
     auto callback = msg.second;
+    m_queueStats.sent++;
```
majestrate: this should be done in the handler hook instead probably

PR #1058, opened 2020-01-18
Bump RC expiration constants significantly

notlesh:


llarp/router_contact.cpp
```diff
@@ -32,10 +32,10 @@ namespace llarp
   /// 1 day for real network
   llarp_time_t RouterContact::Lifetime = 24 * 60 * 60 * 1000;
 #endif
-  /// update RCs every 5 minutes
-  llarp_time_t RouterContact::UpdateInterval = 5 * 60 * 1000;
-  /// an RC inserted long enough ago (30 min) is considered stale and is removed
-  llarp_time_t RouterContact::StaleInsertionAge = 30 * 60 * 1000;
+  /// update RCs every hour
+  llarp_time_t RouterContact::UpdateInterval = 60 * 60 * 1000;
```
majestrate: imo update interval should be stale insertion age minus some scalar value which is our update window so we don't spend 3 hours trying to find something that isn't there.

PR #1064, opened 2020-01-21
don't deadlock on path queue removal if full

majestrate:
when we want to remove lots of paths flush queue when full then add to queue


tewinget:
Took a bit to double-check for a possible race condition with the tick calling the same function simultaneously, but as long as we don't have duplicate entries in the removed paths queue it should be fine.

PR #1067, opened 2020-01-22
change dht behavior and client lookup strategies

majestrate:


llarp/dht/context.hpp
```diff
@@ -88,7 +88,8 @@ namespace llarp
 
       virtual void
       LookupIntroSetForPath(const service::Address& addr, uint64_t txid,
-                            const PathID_t& path, const Key_t& askpeer) = 0;
+                            const PathID_t& path, const Key_t& askpeer,
+                            uint64_t R) = 0;
```
jagerman: `R`?

notlesh: :eyes:

llarp/dht/messages/findintro.hpp
```diff
@@ -35,14 +35,14 @@ namespace llarp
...
-                       bool iterate = true)
+                       bool iterate = false)
           : IMessage({}), S(addr), T(txid)
       {
         N.Zero();
         if(iterate)
           R = 0;
         else
-          R = 1;
+          R = 3;
```
jagerman: I don't understand this change (I'm assuming R is the recursion level).  If client C connects to router X and asks for an introset, X knows that A is the closest router to the requested introset, so connects to A to get the introset then passes it back to C.  But this change here looks like it's going to ask A to get it recursively, and that doesn't make sense to me.

llarp/dht/messages/findintro.cpp
```diff
@@ -105,20 +107,23 @@ namespace llarp
...
         // we are recursive
-        if(dht.Nodes()->FindCloseExcluding(target, peer, exclude))
+        const auto rc = dht.GetRouter()->nodedb()->FindClosestTo(target);
         {
+          peer = Key_t(rc.pubkey);
+          if(peer == us)
+          {
+            replies.emplace_back(new GotIntroMessage({}, T));
+            return true;
+          }
```
jagerman: If we are called to be recursive then we need to consider that the closest router may not have the record, even if the closest router is us.  (Maybe we were temporarily done when the introset got published, or maybe we are a brand new router and weren't active 2 minutes ago when the introset got published).

PR #1070, opened 2020-01-23
make all timestamps monotonic and run loopback testnet at 1/5 speed

majestrate:
builds off #1069

### DIFF
--- a/llarp/util/time.cpp
+++ b/llarp/util/time.cpp
@@ -14,29 +14,21 @@ namespace llarp
         .count();
   }
 
-  static llarp_time_t
-  time_started_at()
-  {
-    const static llarp_time_t started =
-        time_since_epoch< std::chrono::milliseconds, Clock_t >();
-    return started;
-  }
+  const static llarp_time_t started_at_system =
+      time_since_epoch< std::chrono::milliseconds, Clock_t >();
 
+  const static llarp_time_t started_at_steady =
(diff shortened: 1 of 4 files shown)


llarp/util/time.cpp
```diff
@@ -6,15 +6,33 @@ namespace llarp
...
+        time_since_epoch< std::chrono::milliseconds, Clock_t >();
+    return started;
+  }
+
+  static llarp_time_t
+  time_since_started()
+  {
+    const static llarp_time_t started =
+        time_since_epoch< std::chrono::milliseconds,
+                          std::chrono::steady_clock >();
```
jagerman: Should use static initialization outside the function (static locals only get initialized when they are first encountered).

llarp/util/time.cpp
```diff
@@ -6,15 +6,33 @@ namespace llarp
...
-               llarp::Clock_t::now().time_since_epoch())
+    return std::chrono::duration_cast< Res >(Clock::now().time_since_epoch())
         .count();
   }
 
+  static llarp_time_t
+  time_started_at()
+  {
+    const static llarp_time_t started =
+        time_since_epoch< std::chrono::milliseconds, Clock_t >();
```
jagerman: Should use static initialization outside the function (static locals only get initialized when they are first encountered).

PR #1072, opened 2020-01-25
add router version

majestrate:
serialize router version in rc and set it when we run as service node

### DIFF
--- a/llarp/router_version.hpp
+++ b/llarp/router_version.hpp
@@ -61,7 +61,7 @@ namespace llarp
 
    private:
     Version_t m_Version     = {{0, 0, 0}};
-    uint64_t m_ProtoVersion = LLARP_PROTO_VERSION;
+    int64_t m_ProtoVersion = LLARP_PROTO_VERSION;
   };
 
   inline std::ostream&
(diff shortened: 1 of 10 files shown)


llarp/router_version.hpp
```diff
@@ -0,0 +1,40 @@
+#ifndef LLARP_ROUTER_VERSION_HPP
+#define LLARP_ROUTER_VERSION_HPP
+
+#include <array>
+#include <util/bencode.hpp>
+
+namespace llarp
+{
+  struct RouterVersion : public std::array< uint64_t, 4 >
```
jagerman: This inheritance from `std::array` feels weird, partly because it's using one array to store two fundamentally different things (the protocol, and the version triplet).  It seems like this interface could be made cleaner using two member composition rather than inheritance, i.e. something like:
```C++
struct RouterVersion {
  uint64_t protocol_version;
  std::array<uint16_t, 3> version;

  // All the various methods already there, but now operating on `version`/`protocol_version` members rather than inheritance
}
```

majestrate: my justification was that it would be serialized as a single array so why bother splitting it up into 2 members etc

notlesh:
Your branch name convention changed...

```bash
`date -I`-branch-name
```

PR #1073, opened 2020-01-30
rc gossiping

majestrate:


llarp/router/rc_gossiper.cpp
```diff
@@ -0,0 +1,68 @@
...
+      // ensure connected session
+      if(not(s && s->IsEstablished()))
+        return;
+      // check if public router
+      const auto other_rc = s->GetRemoteRC();
+      if(not other_rc.IsPublicRouter())
+        return;
+      // dont send it to the owner
+      if(other_rc.pubkey == rc.pubkey)
+        return;
```
jagerman: Sending it back to the owner can actually be useful: it lets the owner know that another router has accepted and processed it.  In lokid gossiping we try to send every 5 minutes until we see it reflected back to us so that we know that at least one other node accepted it, and once that reflected record comes back we fall back to once-per-hour broadcasting.

llarp/router/router.cpp
```diff
@@ -109,6 +110,18 @@ namespace llarp
...
+  void
+  Router::GossipRCIfNeeded(const RouterContact rc)
+  {
+    /// if we are not a service node forget about gossip
+    if(not IsServiceNode())
+      return;
+    /// wait for random uptime
+    if(Uptime() < _randomStartDelay)
+      return;
+    _rcGossiper.GossipRC(rc);
```
jagerman: Unless I'm missing something it seems like this is going to gossip to the network a lot.  We really want to rate limit this to once/hour since the last successful gossip, or something like once per five minutes if the last successful gossip was > an hour ago.

llarp/router/rc_gossiper.cpp
```diff
@@ -0,0 +1,68 @@
+#include <router/rc_gossiper.hpp>
+#include <messages/dht_immediate.hpp>
+#include <dht/messages/gotrouter.hpp>
+
+namespace llarp
+{
+  RCGossiper::RCGossiper()
+      : I_RCGossiper(), m_Filter(RouterContact::UpdateInterval)
```
jagerman: This filter rate should be quite a bit shorter; UpdateInterval is currently 3h55m with RC expiry at 4h; we want gossip to happen quite a bit more frequently so that if, say, a lokinet is temporarily offline and misses one RC gossip broadcast it won't end up expiring the router.  It also means that if a router changes (for example if it moves to a different machine and the IP changes) the lokinet can't get its updated RC out across the network until this timer is hit, i.e. just under 4 hours.

I'd suggest (though maybe these can be tweaked more):
- 30 minutes on the filter.
- 1h55m "normal" rebroadcast time for the RC.
- 5m rebroadcast time on startup until we see our RC accepted by the network (i.e. gossiped back to us).
- stale RC expiry remains at 4h.

That way you have to miss two consecutive RC broadcasts in order to remove a remote router's RC for staleness, and an updated RC from a router reconfiguration can take effect in a little over 30 minutes.

jagerman: Actually we could shorten those times (including the staleness) a bit.  The gossip is serving as a sort of heartbeat to let us know a router is active, so not having received one in a while is a good sign that a router is down for some reason.

Maybe 15min filter/30min rebroadcast/1h5min for stale?

llarp/router/router.cpp
jagerman: Never mind, there's a filter check inside `GossipRC()` that prevents this.

llarp/router/rc_gossiper.cpp
jagerman: Or lokid's uptime numbers:

30min filter / 1h normal broadcast / 2h5m stale time.

llarp/router/rc_lookup_handler.cpp
```diff
@@ -288,7 +288,10 @@ namespace llarp
         GetRC(r, nullptr, true);
       return;
     }
-    // TODO: only explore via random subset
+    // service nodes gossip, not explore
+    if(_dht->impl->GetRouter()->IsServiceNode())
```
notlesh: Isn't `IsServiceNode()` going to be the same as `useWhiteList`? Which would make this a noop since (1) we hit the if block above and (2) it `return`s at the end, right?

llarp/router/router.cpp
```diff
@@ -50,6 +50,7 @@ namespace llarp
       , _dht(llarp_dht_context_new(this))
       , inbound_link_msg_parser(this)
       , _hiddenServiceContext(this)
+      , _randomStartDelay(1000 * (llarp::randint() % 30) + 10)
   {
```
notlesh: I've used a LERP utility function in the past for this sort of thing. Looks like `C++20` is adding LERP, mmmmm:

https://en.cppreference.com/w/cpp/numeric/lerp

jagerman: This looks like it's going to produce a value from {10, 1010, 2010, ..., 29010} to delay (in milliseconds).  Is this the intended logic?  The 10 (especially the initial one) seems a little strange to me.

Perhaps it was meant to be `1000 * ((llarp::randint() % 30) + 10)` to produce a random value from {10000, 11000, 12000, ..., 39000}?

But yeah, with Stephen's suggestion it could be: `10000 + llarp::randint() % 30000` to get something that makes a (nearly, and good enough) uniform random choice of all the integers between 10000 and 39999.

majestrate:
is this ready for merge?

tewinget:
@majestrate there seem to be unresolved comments/change requests, so I'd say no it's not quite ready for merge.

llarp/router/rc_gossiper.cpp
majestrate: addresssed

llarp/router/rc_gossiper.cpp
majestrate: using lokid's numbers

llarp/router/router.cpp
majestrate: i am unsure if milisecond accuracy is needed here in the randomization.

llarp/router/rc_lookup_handler.cpp
majestrate: no.

notlesh: (Follow up: we decided these are "independent" or "orthogonal")

jagerman:
Please rebase against `dev` (to get the updated travis-ci logic).

majestrate:
rebased

Issue #1039, opened 2020-01-14
Refactor: use std::chrono types for intervals


jagerman:
Currently lokinet uses `llarp_time_t` (which is milliseconds) everywhere for times.  Part of the justification was that it makes serialization easier, but conversely it makes other parts of the code a bit more complicated.

It would be nice to refactor so that std::chrono duration types get used instead (for compiler-defined duration cast math), being mindful of the existing serialization requirements.

(This is low priority)

majestrate:
pretty sure this is half done.

PR #1074, opened 2020-01-31
win32 mega-patch for 0.7

despair86:
- bump version
- actually tell the DNS client where our local resolver is bound to from `lokinet.ini` instead of `localhost`
- move static/invariant packaging assets to installer.lokinet
- improve installer ux by letting the user know that lokinet or its ui is still open


despair86:
want to get this merged so i can take a stab at refactoring `llarp_time_t` to `std::chrono` in #1039 

where `llarp_time_t` is retained at serialisation boundaries, but converted to `std::chrono::time_point` internally

despair86:
merge after jason puts through his CI changes, i ripped the sodium patches out of tree

despair86:
ok looks good to me, @majestrate or whoever wants to take one more look at it

despair86:
I'm not exactly sure how to get the abseil patches applied locally, since we're just going to pull from upstream

cmake/DownloadLibSodium.cmake
```diff
@@ -12,12 +12,19 @@ file(MAKE_DIRECTORY ${LIBSODIUM_PREFIX}/include)
 
 include(ExternalProject)
 include(ProcessorCount)
+
+set(SODIUM_CONFIGURE ./configure --prefix=${LIBSODIUM_PREFIX} --enable-static --disable-shared CC=${CMAKE_C_COMPILER})
```
despair86: setting `CC` on native targets does nothing special at least

despair86:
*crosses fingers*

hope i can get the other stuff upstreamed 😅

despair86:
fixed the mersenne twister seeding

despair86:
what revision of https://github.com/jarro2783/cxxopts did we import in 8a8833d0fd8c9e4271f4ade645767e9810626dee

despair86:
found it

despair86:
oops accidentally dropped netbsd patches

llarp/router/rc_lookup_handler.cpp
```diff
@@ -279,7 +279,8 @@ namespace llarp
 
       if(lookupRouters.size() > LookupPerTick)
       {
-        static std::mt19937_64 rng{std::random_device{}()};
+        uint64_t seed = llarp::randint();
+        static std::mt19937_64 rng{seed};
```
jagerman: This should be inlined:
```c++
static std::mt19937_64 rng{llarp::randint()};
```
so that it only calls randint() once (the first time the mt19937_64 is initialized).

despair86: ah ok

despair86:
holy shit i think i fixed cross compile LTO

despair86:
all builds pass except macOS

despair86:
dropped the abseil patches at last

PR #1075, opened 2020-01-31
Dht blinding

jagerman:


llarp/crypto/crypto_libsodium.cpp
```diff
@@ -179,24 +179,23 @@ namespace llarp
...
 
-      randombytes_buf(r_hash_input, 32);
-
-      // r = H(H(k) || M) where here H(k) is random bytes
+      // r = H(s || M) where here s is pseudorandom bytes typically generated as part of hashing the
+      // seed (i.e. [a,s] = H(k)), but for derived PrivateKeys will come from a hash of the root
+      // key's s concatenated with the derivation hash.
       crypto_hash_sha512_init(&hs);
-      crypto_hash_sha512_update(&hs, r_hash_input, 32);
+      crypto_hash_sha512_update(&hs, privkey.data() + 32, 32);
```
notlesh: It might improve readability to have an accessor on `PrivateKey` for this. What would we even call that? The ed25519 paper just refers to it as the right side of the seed hash...

jagerman: "signing hash", basically.  I think that's the only place it is used.

jagerman: Updated.

notlesh: :+1:

PR #1076, opened 2020-02-01
use std::shared_ptr not a bare pointer to prevent memleak

majestrate:
if `EnsurePathToXXXX` returns false it doesn't call the hook which causes a leak.

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -473,7 +473,7 @@ namespace llarp
           }
           else
           {
-            auto replyMsg = std::make_shared<dns::Message>(std::move(msg));
+            auto replyMsg = std::make_shared< dns::Message >(std::move(msg));
             using service::Address;
             using service::OutboundContext;
             return EnsurePathToService(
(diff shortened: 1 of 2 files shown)


jagerman:
Since it doesn't look like there is any sharing rquired here, this might be better as a unique_ptr in the lambda arguments (avoiding the atomic refcounting of a shared_ptr, and avoiding the copy into the lambda), something like:

```c++
return EnsurePathToService(
                addr,
                [=, replyMsg = std::make_unique<dns::Message>(std::move(msg))](const Address &, OutboundContext *ctx) {
...

majestrate:
On Sat, Feb 01, 2020 at 07:04:10AM -0800, Jason Rhinelander wrote:
> Since it doesn't look like there is any sharing rquired here, this might be better as a unique_ptr in the lambda arguments (avoiding the atomic refcounting of a shared_ptr, and avoiding the copy into the lambda), something like:
> 
> ```c++
> return EnsurePathToService(
>                 addr,
>                 [=, replyMsg = std::make_unique<dns::Message>(std::move(msg))](const Address &, OutboundContext *ctx) {
> ...
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/loki-project/loki-network/pull/1076#issuecomment-581038460

i would love to use std::unique_ptr here but it doesn't compile for some reason

majestrate:
for some reason being: std::function needs the lambda to be copyable.

jagerman:
> i would love to use std::unique_ptr here but it doesn't compile for some reason

That's annoying, but indeed looks like a std::function limitation that cannot accept non-copyable lambda at all.  Gross.

PR #1078, opened 2020-02-03
New travis arches + required endian fixes, libsodium build replacement

jagerman:


jagerman:
There are some travis build errors triggered here but I don't think they should hold up merging this because they are identifying issues that need fixes in other PRs, rather than a problem in the build amendments in this PR.

- arm64 passes.
- gcc amd64 release build is failing in the signing code.  This seems like a real problem, likely caused by the upgrade to bionic from xenial.  However this may well be addressed by #1075, and if not we need to fix it.
- ppc64le doesn't understand `-march=native`; it appears to want `-mcpu=native -mtune=native` instead.
- s390x has some test failures which look like to endian problems (s390x is big endian).

despair86:
> travis has CI runners for even more exotic machines

this is news to me

jagerman:
> this is news to me

It's pretty new (and still beta or alpha or something).

.travis.yml
```diff
@@ -7,33 +8,28 @@ matrix:
   include:
   - name: "lint check"
     os: linux
-    dist: xenial
     compiler: gcc
     env: MAKE_TARGET=format-verify PATH="/usr/lib/llvm-8/bin:$PATH"
     addons:
       apt:
         sources:
-          - llvm-toolchain-xenial-8
```
notlesh: Are we dropping support for `16.04`? We do have `20.04` around the corner...

jagerman: No, there is at least one build here still using 16.04.  I just figured it makes more sense to make the default environment 18.04 with a newer gcc.  (For instance - because it is catching that gcc release test failure).

jagerman:
Rebased against current dev.

PR #1079, opened 2020-02-03
make message queue unbound for direct dht messages

majestrate:


llarp/router/outbound_message_handler.cpp
```diff
@@ -284,7 +284,7 @@ namespace llarp
 
       MessageQueue &path_queue = itr_pair.first->second;
 
-      if(path_queue.size() < MAX_PATH_QUEUE_SIZE)
+      if(path_queue.size() < MAX_PATH_QUEUE_SIZE || entry.pathid.IsZero())
```
tewinget: This *seems* fine at first blush, but bear in mind that if we generate a ton of pathid-zero traffic then no other traffic will ever be sent...

majestrate: true

majestrate: `OutboundMessageHandler::SendRoundRobin` does not count pathid zero messages when doing outbound message processing, so it is not as bad as it could be.

notlesh: I can't see why this would cause problems either, but at the same time I don't have a lot of confidence that it won't...

PR #1082, opened 2020-02-06
only mark path build success if delivered to first hop

majestrate:
if we fail to deliver a path build to the first hop mark it as failure.


tewinget:
I post-hoc approve.

PR #1085, opened 2020-02-06
check for current introset being zero (initial state)

majestrate:


llarp/service/outbound_context.cpp
```diff
@@ -93,7 +93,7 @@ namespace llarp
       updatingIntroSet = false;
       if(i.has_value())
       {
-        if(currentIntroSet.T >= i->T)
+        if(currentIntroSet.T != 0 and currentIntroSet.T >= i->T)
         {
```
notlesh: Suggestions:

1) rename `i`
2) use `>` rather than `>=`
3) make the `== 0` check a separate conditional with a more severe warning

notlesh: Looks like 8c698a1 takes care of this :+1:

PR #1086, opened 2020-02-07
update lokinet monitor to work with iwp

majestrate:


llarp/path/path.cpp
```diff
@@ -307,6 +309,8 @@ namespace llarp
                              {"expiresSoon", ExpiresSoon(now)},
                              {"expiresAt", ExpireTime()},
                              {"ready", IsReady()},
+                             {"tx", m_LastTXRate},
+                             {"rx", m_LastRXRate},
```
notlesh: As I encountered in #1088, it would be nice to keep these a little more descriptive so they aren't ambiguous when we add more fields.

notlesh: Looks like 7374f8f takes care of this :+1:

PR #1087, opened 2020-02-07
dont flood dht with requests, if we know of a router don't look it up

majestrate:


majestrate:
(for exploration replies only)

PR #1091, opened 2020-02-10
Redundant dht lookups

majestrate:


llarp/path/path.hpp
```diff
@@ -127,17 +128,44 @@ namespace llarp
...
           return Hash{}(*p);
         }
       };
 
+      /// hash for std::shared_ptr by path endpoint
+      struct Endpoint_Hash
+      {
+        size_t
+        operator()(const Path_ptr& p) const
+        {
```
notlesh: Is a `std::function` or lambda not a valid parameter to `std::unordered_set`'s hash template parameter?

majestrate: no idea

jagerman: No, the template parameter needs to be a default-constructible type.

llarp/service/endpoint.cpp
```diff
@@ -460,9 +460,33 @@ namespace llarp
...
+      do
+      {
+        --tries;
+        const auto path = PickRandomEstablishedPath();
+        if(path)
+          paths.emplace(path);
+      } while(tries > 0 and paths.size() < PublishRedundancy);
+
+      if(paths.size() != PublishRedundancy)
+        return false;
```
notlesh: Do we need a log statement here, or would some upstream caller effectively do that for us?

llarp/service/endpoint.cpp
```diff
@@ -457,36 +457,42 @@ namespace llarp
         m_OnReady->Stop();
     }
 
-    bool
-    Endpoint::PublishIntroSet(const EncryptedIntroSet& i, AbstractRouter* r)
+    template < typename Endpoint_t >
+    static absl::optional< path::Path::UniqueEndpointSet_t >
+    GetManyPathsWithUniqueEndpoints(Endpoint_t* ep, size_t N, size_t tries = 10)
     {
```
notlesh: The `UniqueEndpoints` part of this method doesn't seem to be enforced at all, seems like this is more of `GetPathsToEndpoint` -- or am I missing something?

majestrate: the `path::Path::UniqueEndpointSet_t` takes care of that part.

llarp/service/endpoint_util.hpp
```diff
@@ -41,6 +41,23 @@ namespace llarp
       GetConvoTagsForService(const ConvoMap& sessions, const Address& addr,
                              std::set< ConvoTag >& tags);
     };
+
+    template < typename Endpoint_t >
+    static absl::optional< path::Path::UniqueEndpointSet_t >
+    GetManyPathsWithUniqueEndpoints(Endpoint_t* ep, size_t N, size_t tries = 10)
+    {
```
notlesh: Would it not suffice to return an empty set rather than an empty optional? More specifically, what's the difference between (1) a non-empty optional containing an empty set and (2) an empty optional?

majestrate: this is a good point

jagerman:
Please rebase against `dev` (to get the updated travis-ci logic).

notlesh:
The state of my refactor (b009f19):

* `llarp_nodedb::FindClosestTo()` makes an ugly copy of the entire router db (see comments for why) -- this needs a fix
* need unit tests for same function
* need unit tests for DHT messaging logic
* compiles, but not tested (might work... :thinking: )

llarp/nodedb.cpp
```diff
@@ -112,6 +114,89 @@ llarp_nodedb::FindClosestTo(const llarp::dht::Key_t &location)
...
+  // TODO: avoid this ugly copy
+  std::vector< llarp::RouterContact > all;
+  all.reserve(entries.size());
+  for(auto &entry : entries)
+  {
+    all.push_back(entry.second.rc);
+  }
+
+  std::vector< llarp::RouterContact > closest;
+  closest.resize(numRouters);
```
jagerman: Perhaps instead of storing all the RouterContacts, just store pointers to the `decltype(entries)::value_type*`, then do a `std::partial_sort` on that with a custom sort function that compares distance (of the deferenced pointer), then finally deref and copy those sorted elements into full `RouterContacts` in `closest`.

majestrate: alternatively store the keys in that vector instead and then fetch the record via key after.

notlesh: Good thoughts, either solution is probably an adequate stop-gap.

notlesh: This is just an obnoxious problem -- there are a lot of inflexible pieces at play here (`XorMetric`, `AlignedBuffer` and its inheritance tree, `partial_sort_copy()`,  `NetDBMap_t`, etc).

I suggest:
1) we not let this hold up general testing
2) in parallel, work on a hand-rolled algorithm OR
2b) just specialize the function to look for exactly 2 or 4 routers

llarp/dht/messages/findintro.cpp
```diff
@@ -43,6 +49,10 @@ namespace llarp
       if(!bencode_start_dict(buf))
         return false;
 
+      // relay order
+      if(!BEncodeWriteDictInt("O", relayOrder, buf))
```
majestrate: encoding needs to be done in lexicographical order (with the current way things are done)

llarp/service/endpoint_util.hpp
majestrate: changed

llarp/service/endpoint.cpp
majestrate: it propagates

llarp/nodedb.cpp
jagerman: This is rewritten now to sort pointers.

PR #1092, opened 2020-02-10
Reduce iwp multiack transmission frequency

majestrate:
reduces the number of redundant packets sent between nodes


tewinget:
Upon cursory review, looks correct and like it will make a bit of a difference.

PR #1093, opened 2020-02-11
make publishing introsets optional 

majestrate:
using `reachable=false` to disable publishing introsets

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -111,6 +111,14 @@ namespace llarp
     bool
     TunEndpoint::SetOption(const std::string &k, const std::string &v)
     {
+      if(k == "reachable")
+      {
+        if(IsFalseValue(v))
+        {
+          m_PublishIntroSet = false;
+          LogInfo(Name(), " setting to be not reachable by default");
+        }
+      }
       if(k == "isolate-network" && IsTrueValue(v.c_str()))
       {
 #if defined(__linux__)
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -118,6 +118,17 @@ namespace llarp
           m_PublishIntroSet = false;
           LogInfo(Name(), " setting to be not reachable by default");
         }
+        else if(IsTrueValue(v))
+        {
+          m_PublishIntroSet = true;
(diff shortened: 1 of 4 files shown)


majestrate:
should it be different for windows?

notlesh:
> should it be different for windows?

Default to true? (e.g. don't publish introsets by default on windows)

jagerman:
Maybe it should be default true everywhere.

majestrate:
`reachable=true` is the default and having a double negative like `no-publish=false` is weird.

jagerman:
Whoops, I meant default false everywhere (i.e. no reachable).

majestrate:
what if we make it tied to using a persistent address? i.e. if `keyfile` is set reachable, otherwise not reachable unless explicitly specified.

jagerman:
That sounds good.  Also I don't think we want the option to apply at all to snapps: they should always be reachable.

majestrate:
updated

llarp/handlers/tun.cpp
```diff
@@ -118,6 +118,17 @@ namespace llarp
...
+          m_PublishIntroSet = true;
+          LogInfo(Name(), " setting to be reachable by default");
+        }
+        else
+        {
+          LogError(Name(), " config option reachable = '", v,
+                   "' does not make sense");
+          return false;
+        }
       }
```
notlesh: This is more of an observation, but we now have >= 15 lines of code to handle the 'reachable' config option... Seems like we may be missing some DRY (etc.) somewhere

llarp/service/context.cpp
```diff
@@ -149,7 +149,12 @@ namespace llarp
...
       configOpts.push_back({"type", DefaultEndpointType()});
       // non reachable by default as this is the default endpoint
-      configOpts.push_back({"reachable", "false"});
+      // but only if no keyfile option provided
+      if(opts.count("keyfile") == 0)
+      {
+        configOpts.push_back({"reachable", "false"});
+      }
+
       {
```
notlesh: Will this effectively override an explicit `reachable=true`?

majestrate: only if it's not provided by user

majestrate: (no)

PR #1094, opened 2020-02-12
monkey patch upstream dns to rewrite .loki cname answers

majestrate:


llarp/handlers/tun.cpp
```diff
@@ -380,12 +380,54 @@ namespace llarp
     {
       // llarp::LogInfo("Tun.HandleHookedDNSMessage ", msg.questions[0].qname, "
       // of type", msg.questions[0].qtype);
+      std::string qname;
+      if(msg.answers.size() > 0)
+      {
+        const auto &answer = msg.answers[0];
+        if(answer.HasCNameForTLD(".loki"))
```
jagerman: Shouldn't this have a `and not (has A record)`?

majestrate: it will strip any A records out

jagerman:
Also should handle .snode

llarp/dns/message.cpp
```diff
@@ -120,36 +121,32 @@ namespace llarp
...
-        if(!ar.Decode(buf))
+        if(!rr.Decode(buf))
         {
-          llarp::LogError("failed to decode additonal");
+          llarp::LogError("failed to decode additional");
           return false;
         }
-        llarp::LogDebug(ar);
       }
+      */
```
jagerman: I feel like this commented-out chunk of code is going to bit rot until we forget why it's commented out.

majestrate: all lost, like tears in rain.

PR #1097, opened 2020-02-13
Submodules

jagerman:
Converts all our unchanged (or very minimally changed) vendors libs to submodules.

This moves them to `external` instead of `vendor`, keeping vendor for things that we actually vendor (like libtuntap).  (Also this avoids git mess when trying to check out a commit across the devendor/submodule commits when the removed code and submodule are in the same place.)

Everything here also gets updated to the latest stable upstream versions:
- abseil LTS 2019_08_08
- nlohmann 3.7.3
- gtest v1.10.0 (this required updating some deprecated test macro usage that broke)
- cxxopts 2.2.0
- ghc-filesystem 1.3.0

libtuntap-master remains vendored because we have significant local changes in it.

It will also make future external updates much easier: just cd into the submodule, git pull, switch to the commit you want, then commit back in the parent repo.


majestrate:
that's a big diff... `._.`

jagerman:
> that's a big diff... `._.`

That's what she said.

majestrate:
waiting on CI

PR #1098, opened 2020-02-14
dont clear our introset every time we fail to make a new set of intros

majestrate:
this may help with introset publishing being off

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -86,18 +86,19 @@ namespace llarp
           ManualRebuild(1);
         return;
       }
-      introSet().I.clear();
+      std::vector< Introduction > intros;
       for(auto& intro : I)
       {
-        introSet().I.emplace_back(std::move(intro));
+        intros.emplace_back(std::move(intro));
       }
-      if(introSet().I.size() == 0)
+      if(intros.size() == 0)
       {
         LogWarn("not enough intros to publish introset for ", Name());
         if(ShouldBuildMore(now) || forceRebuild)
           ManualRebuild(1);
         return;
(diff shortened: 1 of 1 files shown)


llarp/service/endpoint.cpp
```diff
@@ -86,18 +86,19 @@ namespace llarp
           ManualRebuild(1);
         return;
       }
-      introSet().I.clear();
+      std::vector< Introduction > intros;
       for(auto& intro : I)
       {
-        introSet().I.emplace_back(std::move(intro));
+        intros.emplace_back(std::move(intro));
```
notlesh: I doubt the `std::move()` is doing anything useful here (yes, I realize that's not new to this change). Changing `GetCurrentIntrductionsWithFilter()` to take a `std::set` or changing `introSet().I` to be a `std::vector` might avoid extra copies/logic.

Then again, this isn't a hot path and `working - optimized` > `optimized - working`

llarp/service/endpoint.cpp
```diff
@@ -86,18 +86,19 @@ namespace llarp
...
       }
-      if(introSet().I.size() == 0)
+      if(intros.size() == 0)
       {
         LogWarn("not enough intros to publish introset for ", Name());
         if(ShouldBuildMore(now) || forceRebuild)
           ManualRebuild(1);
         return;
       }
+      introSet().I     = intros;
```
notlesh: I'm guessing the big difference here (between `introSet().I.clear()` earlier and `=` here) is that we leave the introset intact if we hit the `return` above. Assuming that's the motivation for this, what good does having a stale introset do?

PR #1105, opened 2020-02-17
spread out path builds

majestrate:


llarp/service/endpoint.cpp
```diff
@@ -1252,23 +1252,25 @@ namespace llarp
     bool
     Endpoint::ShouldBuildMore(llarp_time_t now) const
     {
+      static constexpr auto buildSpread = path::default_lifetime / 4;
       if(path::Builder::BuildCooldownHit(now))
         return false;
-      const bool should = path::Builder::ShouldBuildMore(now);
+      const bool shouldBuild = NumPathsExistingAt(now + buildSpread) < numPaths
+          and NumInStatus(path::ePathBuilding) < numPaths;
```
jagerman: This can just `return true` early if the condition is satisfied; nothing done below can ever result in anything other than a true return value.

majestrate: you're right.

llarp/service/outbound_context.cpp
```diff
@@ -358,22 +358,20 @@ namespace llarp
...
       llarp_time_t t = 0;
       ForEachPath([&t](path::Path_ptr path) {
         if(path->IsReady())
           t = std::max(path->ExpireTime(), t);
       });
       if(t <= now)
-        return should;
-      return should && t - now >= path::default_lifetime / 2;
+        return canBuild;
+      return canBuild && t - now >= path::default_lifetime / 4;
```
jagerman: You could simplify these three lines to:
```C++
return t >= now + path::default_lifetime / 4;
```

PR #1106, opened 2020-02-18
Use endpoint path for outbound context introset update

majestrate:
outbound context does not override `pathset::HandleGotIntroMessage` who's default behavior is to drop messages when not overrided. thus we should use the parent endpoint's paths for lookups as it will properly handle replies.

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -947,12 +947,26 @@ namespace llarp
       return true;
     }
 
+    void
+    Endpoint::MarkAddressOutbound(const Address& addr)
+    {
+      m_state->m_OutboundSessions.insert(addr);
+    }
+
+    bool
+    Endpoint::WantsOutboundSession(const Address& addr) const
+    {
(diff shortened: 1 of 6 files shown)


llarp/path/path.cpp
```diff
@@ -707,12 +707,8 @@ namespace llarp
...
     {
       MarkActive(r->Now());
-      routing::DHTMessage reply;
-      if(!msg.HandleMessage(r->dht(), reply.M))
-        return false;
-      if(reply.M.size())
-        return SendRoutingMessage(reply, r);
-      return true;
+      std::vector< std::unique_ptr< dht::IMessage > > msgs;
+      return msg.HandleMessage(r->dht(), msgs);
```
notlesh: I'm assuming the fact that `msgs` is temporary implicitly drops the replies. This is fairly subtle, so a comment might help. Alternatively, an overloaded `HandleMessage` that doesn't offer any replies would be more explicit.

majestrate: i did not mean to include that change

PR #1111, opened 2020-02-20
Redundant introset propagation

notlesh:
TODO:
1) Ensure this won't create a loop (maybe I shouldn't have ripped out `depth` ...)
2) ~~Related: handle case where `relay == true` and we happen to be one of the closest 4~~
3) Test/Unit test
4) Handle case where our node doesn't have a full Node DB

### DIFF
--- a/llarp/dht/messages/pubintro.cpp
+++ b/llarp/dht/messages/pubintro.cpp
@@ -71,20 +71,32 @@ namespace llarp
           return true;
       }
 
+      const auto& us = dht.OurKey();
+
       // function to identify the closest 4 routers we know of for this introset
-      auto propagateToClosestFour = [&, this]() {
+      auto propagateToClosestFour = [&]() {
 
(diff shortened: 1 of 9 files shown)


llarp/dht/messages/pubintro.cpp
```diff
@@ -40,44 +40,90 @@ namespace llarp
...
+          replies.emplace_back(new GotIntroMessage({}, txID));
+          return true;
       }
+
+      // function to identify the closest 4 routers we know of for this introset
+      auto propagateToClosestFour = [&, this]() {
+
+        // grab 1st & 2nd if we are relayOrder == 0, 3rd & 4th otherwise
+        auto rc0 = (relayOrder == 0 ? closestRCs[0] : closestRCs[2]);
+        auto rc1 = (relayOrder == 0 ? closestRCs[1] : closestRCs[3]);
```
jagerman: `const auto&` to avoid a copy

notlesh: dff9aeb takes care of this.

PR #1113, opened 2020-02-20
JSONRPC updates

jagerman:
(On top of #1110, only the last commit here is relevant).

- Fixes the JSONRPC callable to take the json parameters argument

- Removes the Response/Params/Method_t typedefs as they seemed to
  obscure the underlying types (json/json/string) rather than usefully
  abstract them.


jagerman:
Rebased.

jagerman:
Obsolete by #1306

PR #1116, opened 2020-02-20
path builder prints hops, rest print short name

tewinget:


llarp/path/pathbuilder.cpp
```diff
@@ -440,8 +440,11 @@ namespace llarp
       ctx->router  = m_router;
       auto self    = GetSelf();
       ctx->pathset = self;
-      auto path    = std::make_shared< path::Path >(hops, self.get(), roles);
-      LogInfo(Name(), " build ", path->HopsString());
+      std::string path_shortName = "[" + m_router->ShortName() + "-";
+      path_shortName += m_router->PathBuildNumber() + "]";
+      auto path = std::make_shared< path::Path >(hops, self.get(), roles,
+                                                 path_shortName);
```
jagerman: `std::move(path_shortName)`

PR #1118, opened 2020-02-21
testnet target fixes

majestrate:
* testnet target in makefile depend on building the binary and not do a full rebuild
* fix deadlock in testnet mode


jagerman:
Stale and obsolete.

PR #1122, opened 2020-02-22
De-abseil, part 2: mutex, locks, (most) time

jagerman:
- util::Mutex is now a std::shared_timed_mutex, which is capable of exclusive and shared locks.

- util::Lock is still present as a std::lock_guard<util::Mutex>.

- the locking annotations are preserved, but updated to the latest supported by clang rather than using abseil's older/deprecated ones.

- ACQUIRE_LOCK macro is gone since we don't pass mutexes by pointer into locks anymore (WTF abseil).

- ReleasableLock is gone.  Instead there are now some llarp::util helper methods to obtain unique and/or shared locks:
    - `auto lock = util::unique_lock(mutex);` gets an RAII-but-also unlockable object (`std::unique_lock<T>`, with `T` inferred from `mutex`).
    - `auto lock = util::shared_lock(mutex);` gets an RAII shared (i.e. "reader") lock of the mutex.
    - `auto lock = util::unique_locks(mutex1, mutex2, mutex3);` can be used to atomically lock multiple mutexes at once (returning a tuple of the locks).
  This are templated on the mutex which makes them a bit more flexible than using a concrete type: they can be used for any type of lockable mutex, not only util::Mutex.  (Some of the code here uses them for getting locks around a std::mutex).  Until C++17, using the RAII types is painfully verbose:
  ```C++
  // pre-C++17 - needing to figure out the mutex type here is annoying:
  std::unique_lock<util::Mutex> lock(mutex);
  // pre-C++17 and even more verbose (but at least the type isn't needed):
  std::unique_lock<decltype(mutex)> lock(mutex);
  // our compromise:
  auto lock = util::unique_lock(mutex);
  // C++17:
  std::unique_lock lock(mutex);
  ```
  All of these functions will also warn (under gcc or clang) if you discard the return value.  You can also do fancy things like `auto l = util::unique_lock(mutex, std::adopt_lock)` (which lets a lock take over an already-locked mutex).

- metrics code is gone, which also removes a big pile of code that was only used by metrics:
  - llarp::util::Scheduler
  - llarp::thread::TimerQueue
  - llarp::util::Stopwatch


jagerman:
Also pushed a couple more optimizations here: a much slimmer implementation of `util::memFn`, and remove a bunch of unnecessary empty .cpp files from compilation which will speed up the builds.

PR #1123, opened 2020-02-23
De-abseil, part3: string_view, macros, dates

jagerman:
(Builds on top of #1122)

- string_view is replaced with the string_view from lokimq (I had to fill it out a bit -- it was missing most of the find* methods, but I added tests in loki-mq for the whole class).  When we import loki-mq we can just switch it to use lokimq's version directly.
- a bunch of useless abseil macros are removed.
- the last of abseil date/time code is gone.  Hinnart's `date.h` is added instead; it's a bit overkill since we only need it for one line, but it's header-only *and* is the basis of C++20's date support.
- abseil string manip functions (StrCat, StripAsciiWhitespace) are replaced.  StrCat isn't needed, and StripAsciiWhitespace (now called TrimWhitespace) is easily implementable.

With this, the only abseil left is the 128-bit integer used in network addresses (and its dependency on the abseil hashing library).  @majestrate had some ideas about replacing that.

A few related changes here that *aren't* just de-abseiling (but were warranted for the de-abseiling work):

- the "check" target now runs both test suites (gtest and catch).  There's a new "catch" target to run just the catch2 tests, and "rungtest" to run gtest tests (in the top-level makefile this is just "gtest"; in CMake the `gtest` target is already taken by the gtest static lib, hence `rungtest`).
- `using namespace std::literals;` is now active everywhere in `llarp` (so you can use time or string suffixes without importing literals).
- times printed are now UTC (because it's a nuisance to depend on enough to bring in timezone handling).
- the uptime in the log statements got prettified to `[+1h12m07.123s]` instead of `[+4327123 ms]`.


jagerman:
(This reduces the lokinet binary size on my system from 4888584 to 4182416, i.e. 15%).

jagerman:
I just flipped the testnet with this because... yolo (they've already been running fine with #1122 for several hours now).

despair86:
> times printed are GMT
they were already on Windows, pretty much as soon as I was forced to port abseil (abseil doesn't handle timezones properly on windows to begin with)

majestrate:
what's up with the memory sanitizer CI target?

jagerman:
> what's up with the memory sanitizer CI target?

It's on old ubuntu that hates the GENERATOR expression in the test

jagerman:
Closing (everything here is in #1124).

PR #1124, opened 2020-02-23
De-abseil finale: no more abseil

jagerman:
Builds on #1123 

Final removal of all remaining abseil bits: this adds a uint128_t for what we need so we can stop needing anything from abseil.


despair86:
does this work on 32-bit machines

jagerman:
> does this work on 32-bit machines

It uses two `uint64_t`'s, so it should.  (It also doesn't implement multiplication/division/modulus because we have no use for those here and are considerably more complicated than anything else integers need to do).

despair86:
ah ok fair enough. Sure, shit on i386 all you want but a bunch of ARMshit is still using the 32-bit v7 architecture

despair86:
in the end should the need arise for a full int128 impl with arithmetic operators I _do_ have a copy of int128 mul/div/mod/etc that I scraped from epee then stripped of all epee-ness when I imported it to my session text-mode skeleton project

jagerman:
> re: uint128_t

Good to know.  At the moment we're mainly using it as a container for IPv6 addresses, and so it really only needs bit manipulation; I threw addition/subtraction in just because they are easy.

PR #1127, opened 2020-02-24
Spread snapp path builds evenly

notlesh:
Notice what a PITA it is to deal with `std::chrono` and `llarp_time_t` at the same time...

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -1278,12 +1278,13 @@ namespace llarp
         return false;
 
       size_t numBuilding = NumInStatus(path::ePathBuilding);
-      if (numBuilding > 0)
+      if(numBuilding > 0)
         return false;
 
       static constexpr auto buildSpread = path::default_lifetime / 4;
-      const auto sinceEpoch = m_lastBuildStarted.time_since_epoch();
-      const auto sinceEpochMs = std::chrono::duration_cast< std::chrono::milliseconds >(sinceEpoch);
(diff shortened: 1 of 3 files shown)


notlesh:
Re: 42967db: This is probably insufficient to change path timeouts for SNApp introset publishing. There is a lot of code that is hard-coded to use `default_lifetime`, and so this will probably have unintended consequences. Ideally we would ensure that a client can pick an its own path lifetime and that it's properly serialized, respected by other clients, etc.

majestrate:
42967db is a protocol breaking change that affects relays.

llarp/service/endpoint.cpp
```diff
@@ -75,10 +75,10 @@ namespace llarp
     Endpoint::RegenAndPublishIntroSet(bool forceRebuild)
     {
       const auto now = llarp::time_now_ms();
-      std::set< Introduction > I;
+      std::set< Introduction > introset;
       if(!GetCurrentIntroductionsWithFilter(
-             I, [now](const service::Introduction& intro) -> bool {
-               return not intro.ExpiresSoon(now, 2 * 60 * 1000);
+             introset, [now](const service::Introduction& intro) -> bool {
+               return not intro.ExpiresSoon(now, 10 * 60 * 1000);
```
majestrate: this can be `std::chrono::milliseconds(path::default_lifetime) / 2` instead of changing the path lifetime

notlesh: I'll make this a constant value (grouped with other similar ones) once this is rebased on top of the `chrono` PR. Similarly with the `buildSpread` variable.

PR #1128, opened 2020-02-24
make llarp_time_t use std::chrono::milliseconds

majestrate:
no more `uint64_t`

### DIFF
--- a/llarp/constants/path.hpp
+++ b/llarp/constants/path.hpp
@@ -4,6 +4,7 @@
 #include <cstddef>
 
 #include <util/types.hpp>
+#include <util/time.hpp>
 
 namespace llarp
 {
@@ -16,15 +17,14 @@ namespace llarp
     /// pad messages to the nearest this many bytes
     constexpr std::size_t pad_size = 128;
     /// default path lifetime in ms
-    constexpr llarp_time_t default_lifetime = 10 * 60 * 1000;
+    static constexpr auto default_lifetime = 10min;
     /// after this many ms a path build times out
-    constexpr llarp_time_t build_timeout = 30000;
+    static constexpr auto build_timeout = 30s;
 
     /// measure latency every this interval ms
-    constexpr llarp_time_t latency_interval = 5000;
-
+    static constexpr auto latency_interval = 5s;
     /// if a path is inactive for this amount of time it's dead
-    constexpr llarp_time_t alive_timeout = 60000;
+    static constexpr auto alive_timeout = 30s;
   }  // namespace path
 }  // namespace llarp
 
--- a/llarp/exit/session.cpp
+++ b/llarp/exit/session.cpp
@@ -38,7 +38,7 @@ namespace llarp
     BaseSession::ExtractStatus() const
     {
       auto obj            = path::Builder::ExtractStatus();
-      obj["lastExitUse"]  = m_LastUse;
+      obj["lastExitUse"]  = m_LastUse.count();
       auto pub            = m_ExitIdentity.toPublic();
       obj["exitIdentity"] = pub.ToString();
       return obj;
@@ -55,7 +55,7 @@ namespace llarp
     {
       const size_t expect = (1 + (numPaths / 2));
       // check 30 seconds into the future and see if we need more paths
-      const llarp_time_t future = now + (30 * 1000) + buildIntervalLimit;
+      const llarp_time_t future = now + 30s + buildIntervalLimit;
       return NumPathsExistingAt(future) < expect && !BuildCooldownHit(now);
     }
 
@@ -91,7 +91,7 @@ namespace llarp
     bool
     BaseSession::CheckPathDead(path::Path_ptr, llarp_time_t dlt)
     {
-      return dlt >= 10000;
+      return dlt >= 10s;
     }
 
     void
@@ -127,7 +127,7 @@ namespace llarp
     bool
     BaseSession::HandleGotExit(llarp::path::Path_ptr p, llarp_time_t b)
     {
-      if(b == 0)
+      if(b == 0s)
       {
         llarp::LogInfo("obtained an exit via ", p->Endpoint());
         CallPendingCallbacks(true);
--- a/llarp/exit/session.cpp
+++ b/llarp/exit/session.cpp
@@ -38,7 +38,7 @@ namespace llarp
     BaseSession::ExtractStatus() const
     {
       auto obj            = path::Builder::ExtractStatus();
-      obj["lastExitUse"]  = m_LastUse.count();
+      obj["lastExitUse"]  = to_json(m_LastUse);
       auto pub            = m_ExitIdentity.toPublic();
       obj["exitIdentity"] = pub.ToString();
       return obj;
--- a/llarp/iwp/session.hpp
+++ b/llarp/iwp/session.hpp
@@ -18,19 +18,19 @@ namespace llarp
     CreatePacket(Command cmd, size_t plainsize, size_t min_pad = 16,
                  size_t pad_variance = 16);
     /// Time how long we try delivery for
-    static constexpr auto DeliveryTimeout = 500ms;
+    static constexpr llarp_time_t DeliveryTimeout = 500ms;
     /// Time how long we wait to recieve a message
-    static constexpr auto ReceivalTimeout = (DeliveryTimeout * 8) / 5;
+    static constexpr llarp_time_t ReceivalTimeout = (DeliveryTimeout * 8) / 5;
     /// How long to keep a replay window for
-    static constexpr auto ReplayWindow = (ReceivalTimeout * 3) / 2;
+    static constexpr llarp_time_t ReplayWindow = (ReceivalTimeout * 3) / 2;
     /// How often to acks RX messages
-    static constexpr auto ACKResendInterval = DeliveryTimeout / 2;
+    static constexpr llarp_time_t ACKResendInterval = DeliveryTimeout / 2;
     /// How often to retransmit TX fragments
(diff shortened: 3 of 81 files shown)


llarp/service/endpoint.hpp
```diff
@@ -59,16 +59,16 @@ namespace llarp
     };
     using ConvoEventListener_ptr = std::shared_ptr< IConvoEventListener >;
 
+    /// minimum interval for publishing introsets
+    static constexpr auto INTROSET_PUBLISH_INTERVAL =
+        path::default_lifetime / 4;
```
notlesh: Is integer truncation going to behave differently / less predictably with `std::chrono`?

```cpp
int seconds = 15 / 4
```
is very different from
```cpp
int ms = 1500 / 4
```

`/ 4` just may not be a very safe way to do this in general...

llarp/service/endpoint_state.hpp
```diff
@@ -76,9 +76,9 @@ namespace llarp
       PendingRouters m_PendingRouters;
 
       uint64_t m_CurrentPublishTX       = 0;
-      llarp_time_t m_LastPublish        = 0;
-      llarp_time_t m_LastPublishAttempt = 0;
-      llarp_time_t m_MinPathLatency     = (5 * 1000);
+      llarp_time_t m_LastPublish        = 0s;
+      llarp_time_t m_LastPublishAttempt = 0s;
+      llarp_time_t m_MinPathLatency     = 1s;
       /// our introset
```
notlesh: The actual time value looks to have changed here (`5` vs `1`)

llarp/util/codel.hpp
```diff
@@ -132,8 +131,9 @@ namespace llarp
             if(lowest > dropMs)
             {
               item->~T();
-              nextTickInterval += initialIntervalMs / std::sqrt(++dropNum);
-              firstPut   = 0;
+              nextTickInterval +=
+                  initialIntervalMs / uint64_t(std::sqrt(++dropNum));
```
notlesh: is this an intentional behavior change?

llarp/service/endpoint.hpp
majestrate: it shouldn't be that bad right?

llarp/service/endpoint_state.hpp
majestrate: 5 seconds is really high latency.

llarp/util/codel.hpp
majestrate: is what?

llarp/service/endpoint.hpp
notlesh: as long as it's a multiple of 4 :laughing:

llarp/service/endpoint_state.hpp
notlesh: Fair enough, it's just a subtle change inside such a big diff.

llarp/util/codel.hpp
notlesh: casting to `uint64_t`

majestrate: otherwise it's a float.

llarp/service/endpoint.hpp
jagerman: `default_lifetime` is 600s, isn't it?

jagerman: No, it's 10min.  so we probably want something like `std::chrono::seconds{path::default_lifetime} / 4`

notlesh:
```c++
#include <chrono>
#include <iostream>

using namespace std::chrono_literals;

int main(int argc, char** argv) {
    auto d1 = 10min / 4;
    std::cout << "d1 = " << d1.count() << std::endl;

    auto d2 = std::chrono::duration_cast<std::chrono::milliseconds>(d1);
    std::cout << "d2 = " << d2.count() << std::endl;

    return 0;
}
```
output:
```
d1 = 2
d2 = 120000
```

So there's definitely a caveat here with the time unit used for `time_point` or `duration`. Probably to my earlier point: division may best be avoided, and when it's not, consider carefully how truncation will play out.

notlesh: It might be a good idea to standardize on a time unit (milliseconds or microseconds) and use those when defining types. Obviously this would still allow for passing other types, but would cut down on the chances of division and/or `count()` giving unintended results.

Ex:

```c++
struct Foo {
  std::chrono::milliseconds m_timeout;
  void setTimeout(std::chrono::milliseconds timeout);
};

Foo foo;
foo.setTimeout(5s); // still able to use literals or durations with arbitrary time units
```

majestrate: as i understand it we chose milliseconds long ago as the time unit. has that changed or no?

llarp/constants/path.hpp
```diff
@@ -16,15 +17,14 @@ namespace llarp
...
+    static constexpr auto build_timeout = 30s;
 
     /// measure latency every this interval ms
-    constexpr llarp_time_t latency_interval = 5000;
-
+    static constexpr auto latency_interval = 5s;
     /// if a path is inactive for this amount of time it's dead
-    constexpr llarp_time_t alive_timeout = 60000;
+    static constexpr auto alive_timeout = 30s;
   }  // namespace path
```
notlesh: Another logic change here. (I'm just taking notes at this point, this seems like a reasonable change)

llarp/exit/session.cpp
```diff
@@ -55,7 +55,7 @@ namespace llarp
     {
       const size_t expect = (1 + (numPaths / 2));
       // check 30 seconds into the future and see if we need more paths
-      const llarp_time_t future = now + (30 * 1000) + buildIntervalLimit;
+      const llarp_time_t future = now + 30s + buildIntervalLimit;
```
notlesh: FYI:

```c++
#include <chrono>
#include <iostream>

using namespace std::chrono_literals;

int main(int argc, char** argv) {
    auto mixed = 10min + 30s;
    std::cout << "mixed = " << mixed.count() << std::endl;

    auto mixed2 = 40s + 10min;
    std::cout << "mixed2 = " << mixed2.count() << std::endl;

    return 0;
}
```
Output:
```
mixed = 630
mixed2 = 640
```

llarp/iwp/session.hpp
```diff
@@ -17,24 +17,24 @@ namespace llarp
...
+    /// How long to keep a replay window for
+    static constexpr auto ReplayWindow = (ReceivalTimeout * 3) / 2;
+    /// How often to acks RX messages
+    static constexpr auto ACKResendInterval = DeliveryTimeout / 2;
+    /// How often to retransmit TX fragments
+    static constexpr auto TXFlushInterval = (DeliveryTimeout / 5) * 4;
+    /// How often we send a keepalive
+    static constexpr auto PingInterval = 5s;
+    /// How long we wait for a session to die with no tx from them
+    static constexpr auto SessionAliveTimeout = PingInterval * 5;
```
notlesh: Should we use `llarp_time_t` (aka `std::chrono::milliseconds`) here to make sure things like `count()` and `/` don't have surprises? Especially in the case of `PingInterval`...

majestrate: yes

llarp/path/path.cpp
```diff
@@ -793,7 +798,7 @@ namespace llarp
         }
         LogInfo(Name(), " ", Endpoint(), " Rejected exit");
         MarkActive(r->Now());
-        return InformExitResult(msg.B);
+        return InformExitResult(Time_t(msg.B));
```
notlesh: `Time_t` ? do we want to get rid of `Time_t`?

llarp/constants/path.hpp
majestrate: 60s is very long

llarp/util/time.cpp
```diff
@@ -10,8 +10,7 @@ namespace llarp
   static llarp_time_t
   time_since_epoch()
   {
-    return std::chrono::duration_cast< Res >(Clock::now().time_since_epoch())
-        .count();
+    return std::chrono::duration_cast< Res >(Clock::now().time_since_epoch());
```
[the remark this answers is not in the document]
majestrate: yes

llarp/service/intro_set.cpp
```diff
@@ -214,7 +214,7 @@ namespace llarp
           return false;
       }
       // Timestamp published
-      if(!BEncodeWriteDictInt("t", T, buf))
+      if(!BEncodeWriteDictInt("t", T.count(), buf))
         return false;
```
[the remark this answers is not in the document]
majestrate: it's serializing

llarp/router/router.cpp
```diff
@@ -300,8 +300,8 @@ namespace llarp
   Router::handle_router_ticker()
   {
     ticker_job_id = 0;
-    LogicCall(logic(), std::bind(&Router::Tick, this));
-    ScheduleTicker(ROUTER_TICK_INTERVAL_MS);
+    Tick();
```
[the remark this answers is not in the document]
majestrate: yes

llarp/router/router.cpp
```diff
@@ -32,7 +32,7 @@
 #include <unistd.h>
 #endif
 
-constexpr uint64_t ROUTER_TICK_INTERVAL_MS = 1000;
+static constexpr auto ROUTER_TICK_INTERVAL = 1s;
```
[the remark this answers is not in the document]
majestrate: i think i addressed this

llarp/path/path.cpp
majestrate: it's gone now

PR #1129, opened 2020-02-25
Fix string_view C++17 compatibility

jagerman:
string_view was implicitly convertible to std::string, but
std::string_view is only explicitly convertible.  This makes the
`operator std::string` explicit to be more compatible, and re-adds a
bunch of explicit string casts to the code where needed.

(This also fixes the build if changing the standard to c++17)


jagerman:
CC @despair86 - should fix the c++17 windows build

PR #1140, opened 2020-02-26
some light cmake cleanup for win32

despair86:


despair86:
also might become part of a full sweep of the build system itself

despair86:
merge this one first

cmake/DownloadLibSodium.cmake
```diff
@@ -12,12 +12,24 @@ file(MAKE_DIRECTORY ${LIBSODIUM_PREFIX}/include)
 
 include(ExternalProject)
 include(ProcessorCount)
+
+if (CMAKE_C_COMPILER_ARG1)
```
jagerman: What?

despair86: cmake doesn't pass through `-m[32|64]` to autotools if the user specified it in `CMAKE_C_COMPILER`/`ENV{CC}`

PR #1146, opened 2020-02-27
remove dht explore timer

majestrate:


llarp/dht/context.cpp
```diff
@@ -487,19 +472,13 @@ namespace llarp
 
     void
     Context::Init(const Key_t& us, AbstractRouter* r,
-                  llarp_time_t exploreInterval)
+                  llarp_time_t)
```
jagerman: Should just drop this unused parameter from the header.

PR #1149, opened 2020-03-01
fix dht propagation

majestrate:
serialized messages were not de-serialized right.


notlesh:
Also, do see my comments on `discord` if you haven't already.

PR #1151, opened 2020-03-02
fix up introset lookups

majestrate:


despair86:
this broke windows and SysV (why are we giving up lookups after 500ms?)

despair86:
rev. 79746e371991060d6f7ebf4e50d7505c852c9b79

llarp/service/endpoint.cpp
```diff
@@ -1003,6 +1008,10 @@ namespace llarp
         }
       }
 
+      // filter check for address
+      if(m_state->m_RemoteLookupFilter.Insert(remote))
```
despair86: found the bug

PR #1158, opened 2020-03-04
Prune non routers from nodedb

majestrate:


majestrate:
`disk->addJob(std::bind(&KillRCJobs, files));`

PR #1160, opened 2020-03-04
add llarp.admin.die rpc call to gracefully kill router

majestrate:
also patches possible dns rebinding vuln. moving to lokimq makes this irrelevant.

### DIFF
--- a/libabyss/src/server.cpp
+++ b/libabyss/src/server.cpp
@@ -96,7 +96,8 @@ namespace abyss
       {
         // TODO: header whitelist
         return name == string_view("content-type")
-            || name == string_view("content-length");
+            || name == string_view("content-length")
+            || name == string_view("host");
       }
 
       bool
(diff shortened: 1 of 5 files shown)


libabyss/src/server.cpp
```diff
@@ -162,6 +163,17 @@ namespace abyss
           {
             m_BodyParser.reset(json::MakeParser(contentLength));
           }
+          itr = Header.Headers.find("host");
+          if(itr == Header.Headers.end())
```
notlesh: Should this be `Host` instead of `host`?

notlesh: ...and, in theory, the HTTP responses should be consistent with this...

llarp/rpc/rpc.cpp
```diff
@@ -304,6 +308,12 @@ namespace llarp
 
       ~Handler() override = default;
 
+      bool
+      ValidateHost(const std::string& host) const override
+      {
+        return host == "localhost" || host == expectedHostname;
```
notlesh: Should this include `127.0.0.1`, etc.?

libabyss/src/server.cpp
notlesh: The headers appear to be normalized to lowercase, so this seems to work. :+1:

PR #1164, opened 2020-03-05
Don't ignore .exe, .png, .gif

jagerman:
These are part of the source code (in the documentation, and used in the
Windows build) so we really should still be including them.  (We can aim
at eliminating them, but should eliminate them from git, not just from
exported source).


despair86:
there weren't any binaries left by this point

PR #1168, opened 2020-03-07
Dont rely on operator[] side effects

majestrate:


llarp/messages/relay_commit.cpp
```diff
@@ -332,9 +333,10 @@ namespace llarp
       }
       else
       {
+        const auto now = time_now_ms();
         // persist session to downstream until path expiration
         self->context->Router()->PersistSessionUntil(
-            self->hop->info.downstream, self->hop->ExpireTime() + 10s);
+            self->hop->info.downstream, now + path::default_lifetime);
         // put hop
```
notlesh: `now + path::default_lifetime` may way overshoot the actual remaining life of a path, although this is probably harmless

PR #1171, opened 2020-03-08
better nodedb pruning logic

majestrate:
this also prunes client nodedbs so we don't have rcs from old 0.6.x net that are invalid.

### DIFF
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -747,7 +747,7 @@ namespace llarp
       // if we have no whitelist enabled or we have
       // the whitelist enabled and we got the whitelist
       // check against the whitelist and remove if it's not
-      // in the whitelist OR if there is no whitelist don't remoev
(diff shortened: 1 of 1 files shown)


llarp/router/router.cpp
```diff
@@ -737,6 +737,14 @@ namespace llarp
...
       });
     }
+    else
+    {
+      nodedb()->RemoveIf([&](const RouterContact &rc) -> bool {
+        if(IsBootstrapNode(rc.pubkey))
+          return false;
+        return not rc.IsPublicRouter();
+      });
+    }
```
notlesh: This logic will run if we haven't received the whitelist yet. Will that cause everything to be removed in this case? That would essentially mean we wipe our nodedb at startup.

majestrate: good point, adding check.

majestrate:
this is pretty important we should get this merged before rc3

PR #1173, opened 2020-03-09
Fix IPv6 lookups to return empty response

jagerman:
The logic here wasn't quite right and was returning an A record in response to an AAAA lookup.  DNS resolvers didn't accept that, so `ping` would delay 5 seconds (waiting for the AAAA lookup to return something valid) before giving up and trying an A lookup.

This returns nothing, which is better, but not quite enough: this gives empty responses, which produces warnings in host/dig.

### DIFF
--- a/llarp/dns/message.cpp
+++ b/llarp/dns/message.cpp
@@ -194,6 +194,31 @@ namespace llarp
     }
 
     void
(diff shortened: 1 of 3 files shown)


majestrate:
waiting for ci

PR #1175, opened 2020-03-09
limit path builds on snode sessions

majestrate:
respect path builder cooldown.

### DIFF
--- a/llarp/exit/session.cpp
+++ b/llarp/exit/session.cpp
@@ -53,10 +53,12 @@ namespace llarp
     bool
     BaseSession::ShouldBuildMore(llarp_time_t now) const
     {
+      if(BuildCooldownHit(now))
+        return false;
       const size_t expect = (1 + (numPaths / 2));
       // check 30 seconds into the future and see if we need more paths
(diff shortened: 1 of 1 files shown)


llarp/exit/session.cpp
```diff
@@ -53,10 +53,12 @@ namespace llarp
...
     BaseSession::ShouldBuildMore(llarp_time_t now) const
     {
+      if(BuildCooldownHit(now))
+        return false;
       const size_t expect = (1 + (numPaths / 2));
       // check 30 seconds into the future and see if we need more paths
       const llarp_time_t future = now + 30s + buildIntervalLimit;
-      return NumPathsExistingAt(future) < expect && !BuildCooldownHit(now);
+      return NumPathsExistingAt(future) < expect;
     }
```
notlesh: This is functionally equivalent, other than `BuildCooldownHit()` being the last checked condition instead of the first.

majestrate: yea, the real check is the next hunk as i understand it.

majestrate: (this was some cleanup)

PR #1176, opened 2020-03-10
fully randomize hop selection

majestrate:
before we were only using routers who we are directly connected to or recently looked up the RC for. now it's uniform random, this should help with spreading out traffic and anonymity related attributes.


majestrate:
god damnit travis

PR #1184, opened 2020-03-12
Path build rejected event 2020 03 12

notlesh:
This builds on Tom's Tooling :tm: branch/PR.

Note that this creates a segfault in the unit test code. Apparently this is the first time we hit a codepath inside testing where we want to use (dereference) `MockContext::getRouter()`.


notlesh:
18b8a38 was cherry-picked (and amended) into the official `tooling` branch.

Issue #932, opened 2019-11-26
Logging to file results in lost log statements


notlesh:
When lokinet is configured to log to file, it initially starts up printing log statements to `STDOUT`. After parsing the config file, it transitions to logging to file as appropriate, but it seems that there is a brief period where log statements are sent nowhere.

Example statements in `lokinet.ini`:

```ini
[logging]
type=file
file=/home/stephen/.lokinet/log
```

Here is an example output when logging to file:

STDOUT:
```log
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+1 ms] main.cpp:64   Using config file: /home/stephen/.lokinet/lokinet.ini
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+2 ms] config/config.cpp:64  setting netid to 'bugs'
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+2 ms] config/config.cpp:180 setting profiles to /home/stephen/.lokinet/profiles.dat
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+2 ms] config/config.cpp:211 set local dns to 127.3.2.1:53
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+2 ms] config/config.cpp:206 add upstream resolver 1.1.1.1
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+2 ms] config/config.cpp:391 open log file: /home/stephen/.lokinet/log
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+2 ms] config/config.cpp:397 will log to file /home/stephen/.lokinet/log
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+3 ms] context.cpp:193       lokinet-0.5.2-e8e2e21f (dev build)
[NFO] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+3 ms] context.cpp:194       starting up
[WRN] [lokinet](990) Tue Nov 26 15:07:59 2019 MST [+3 ms] router/router.cpp:386 !!!! you have manually set netid to be 'bugs' which does not equal 'beta' you will run as a different network, good luck and don't forget: something something MUH traffic shape correlation !!!!
[...]

PR #1186, opened 2020-03-13
Config cleanup 

notlesh:
This is an overhaul of parts of the config system. The goals are to make the code more understandable, handle errors more gracefully and explicitly, and to create a better definition of what options our config file supports.

There are a few remaining TODOs, mostly around 3 areas:
1) SNApp config -- needs its own cleanup
2) Many config options are still loose containers of strings, this should be fixed where possible
3) Some things aren't documented still, either because they may change or it's not obvious how to documente them (in part because of (2))

A note about style: I haven't adhered to `make format` since we intend to change the rules. There are probably hundreds of violations of our current format rules in this PR, so I suggest we nail that down before merging this.


notlesh:
Example of a generated relay conf file:

```ini
# Configuration for routing activity.
[router]

#job-queue-size=8192

#netid=lokinet

# Minimum number of routers lokinet will attempt to maintain connections to.
#min-connections=6

# Maximum number (hard limit) of routers lokinet will be connected to at any time.
#max-connections=60

# Router nickname. Kee wanted it.
#nickname=

# Optional directory for containing lokinet runtime data. This includes generated
# private keys.
#data-dir=/home/stephen/.lokinet/

#public-address=

#public-port=1090

#worker-threads=1

#net-threads=1

#block-bogons=1


# Network settings
[network]

#profiling=1

# File to contain router profiles.
#profiles=profiles.dat

# Public key of a router which will act as sole first-hop. This may be used to
# provide a trusted router (consider that you are not fully anonymous with your
# first hop).
#strict-connect=

#upstream-dns=

#local-dns=


# Configuration for lokinet's database of service nodes
[netdb]

# Root directory of netdb.
#dir=


# JSON API settings
[api]

# Determines whether or not the JSON API is enabled.
#enabled=1

# IP address and port to bind to.
# Recommend localhost-only for security purposes.
#bind=127.0.0.1:1190


# Lokid configuration (settings for talking to lokid
[lokid]

# File containing service node's seed.
#service-node-seed=

# Whether or not we should talk to lokid. Must be enabled for staked routers.
#enabled=0

# Host and port of running lokid that we should talk to.
#jsonrpc=127.0.0.1:22023

#username=

#password=


# Configure nodes that will bootstrap us onto the network
[bootstrap]

# Specify a bootstrap file containing a signed RouterContact of a service node
# which can act as a bootstrap. Accepts multiple values.
#add-node=


# logging settings
[logging]

# Log type (format). Valid options are:
#   file - plaintext formatting
#   json - json-formatted log statements
#   syslog - logs directed to syslog
#type=file

# Minimum log level to print. Logging below this level will be ignored.
# Valid log levels, in ascending order, are:
#   trace
#   debug
#   info
#   warn
#   error
#level=info

#file=stdout
```

llarp/config/key_manager.cpp
```diff
@@ -31,10 +31,10 @@ namespace llarp
     if(m_initialized)
       return false;
 
-    m_rcPath           = config.router.ourRcFile();
-    m_idKeyPath        = config.router.identKeyfile();
-    m_encKeyPath       = config.router.encryptionKeyfile();
-    m_transportKeyPath = config.router.transportKeyfile();
+    m_rcPath           = config.router.m_dataDir + "/rc.signed";
```
majestrate, line 34: these should use `fs::path` because windows

notlesh: Good point... fixed in db6abd8.

llarp/router/router.cpp
```diff
@@ -386,34 +385,23 @@ namespace llarp
...
-    addrInfo           = conf->router.addrInfo();
-    publicOverride     = conf->router.publicOverride();
-    ip4addr            = conf->router.ip4addr();
-
-    if(!conf->router.blockBogons().value_or(true))
-    {
-      RouterContact::BlockBogons = false;
-    }
+        conf->router.m_minConnectedRouters;
+    encryption_keyfile = conf->router.m_dataDir + "/encryption.key";
```
majestrate, line 397: `fs::path` for here because windows.

notlesh: Fixed in 5f6a23d

notlesh:
I believe 0029c10 fixes #932

notlesh:
Reverted back to ca24f25, which is a fairly stable state for this PR. The continued work can be opened as a new PR and will further address SNApp / Endpoint config related issues.

notlesh:
Note to future self: if we need the IWP factory code, the commit that pulled it out was 6fd270b

PR #1187, opened 2020-03-14
discard unknown fields in introset for future use

majestrate:
this allows us to put more things into introset in the future without breaking compat

### DIFF
--- a/llarp/service/intro_set.cpp
+++ b/llarp/service/intro_set.cpp
@@ -186,7 +186,10 @@ namespace llarp
       if(!BEncodeMaybeReadDictEntry("z", Z, read, key, buf))
         return false;
 
-      return read;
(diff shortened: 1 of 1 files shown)


majestrate:
`BEncodeMaybeReadDictEntry` is a nop if the key does not match the value given as a parameter.

jagerman:
> `BEncodeMaybeReadDictEntry` is a nop if the key does not match the value given as a parameter.

Which is going to be a problem for backwards compatibility.  If a future version adds a key `"j"` into the introset then older clients are going to choke when they find `"j"` but were expecting `"k"`.

majestrate:
the function in question is called per key/value in a dict.

jagerman:
Oh, I see.

PR #1194, opened 2020-03-23
disable accedental fec in iwp outbound messages

majestrate:


llarp/iwp/message_buffer.cpp
```diff
@@ -13,6 +13,7 @@ namespace llarp
         : m_Data{std::move(msg)}
         , m_MsgID{msgid}
         , m_Completed{handler}
+        , m_LastFlush{now}
```
notlesh: The net effect of this seems to be that we delay transmission. IIRC, this was even reflected in measurable latency in testing. Should we revisit this?

majestrate:
On Mon, Mar 30, 2020 at 07:18:08AM -0700, Stephen Shelton wrote:
> @notlesh commented on this pull request.
> 
> 
> 
> > @@ -13,6 +13,7 @@ namespace llarp
>          : m_Data{std::move(msg)}
>          , m_MsgID{msgid}
>          , m_Completed{handler}
> +        , m_LastFlush{now}
> 
> The net effect of this seems to be that we delay transmission. IIRC, this was even reflected in measurable latency in testing. Should we revisit this?

I believe that the later commits fixes this delay

PR #1205, opened 2020-03-31
[client side protocol change]: remove inner protocol bloat.

majestrate:


majestrate:
this is bitrotten and breaks protocol, closing.

PR #1208, opened 2020-04-01
[ASAN] copy don't move

majestrate:


llarp/iwp/session.cpp
```diff
@@ -254,20 +254,17 @@ namespace llarp
...
       }
-      auto self = shared_from_this();
       if(m_EncryptNext && !m_EncryptNext->empty())
       {
-        m_Parent->QueueWork([self, data = std::move(m_EncryptNext)] {
-          self->EncryptWorker(data);
-        });
+        m_Parent->QueueWork(std::bind(&Session::EncryptWorker,
+                                      shared_from_this(), m_EncryptNext));
         m_EncryptNext = nullptr;
```
jagerman, line 261: This assignment seems suspect to me, but also unnecessary with the previous code: the `data = std::move(m_EncryptNext)` is required to leave `m_EncryptNext` empty.

jagerman: Do any other threads ever touch `m_EncryptNext`?

majestrate: no other threads *should be* touching `m_EncryptNext`

PR #1215, opened 2020-04-03
Update clang format rules 

notlesh:
This updates `clang-format` rules. Unfortunately, since `clang-format` aggressively enforces its rules (it is never tolerant; that is, for any given rule, it always enforces some style), this change alone would require an enormous diff to the codebase (a "git blame wall", as Jeff called it).

So this also introduces `git-clang-format` (part of the typical `clang-format` distribution) and a `git pre-commit` hook script found in the following submodule:

`external/clang-format-hooks -> https://github.com/barisione/clang-format-hooks/`

This (currently) requires a one-time setup per clone:

```bash
# from repo root
# also, after e.g.
# git submodule update --init --recursive

./external/clang-format-hooks/git-pre-commit-format install

# this seems to use whatever `clang-format` it finds, so if `clang-format` resolves to
# the wrong version (e.g. probably version 9 on recent Ubuntu), adjust accordingly:
sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-8 100

# or perhaps add a symlink in your $PATH to the desired version, or an alias...

# the environment variable $CLANG_FORMAT can also be set to the desired binary
```

Once this is done, any `git commit` will pause with an interactive question about applying the `clang-format` rules. Example:

```diff
git commit -va
--- daemon/main.cpp	(before formatting)
+++ daemon/main.cpp	(after formatting)
@@ -27,8 +27,10 @@
 handle_signal(int sig)
 {
   // omg no more explosion of template braces
-  std::set< std::string > > its2020afterAll;
-  its2020afterAll.insert("Some really, really long string but this time without the profanity. Keep your stick, though.");
+  std::set<std::string>> its2020afterAll;
+  its2020afterAll.insert(
+      "Some really, really long string but this time without the profanity. Keep your stick, "
+      "though.");

   if(ctx)
   {

The staged content is not formatted correctly.
The fix shown above can be applied automatically to the commit.

You can:
 [a]: Apply the fix
 [f]: Force and commit anyway (not recommended!)
 [c]: Cancel the commit
 [?]: Show help

What would you like to do? [a/f/c/?]
```

Remaining tasks:
1) Avoid the one-off `install` mentioned above. Worst case: invoke from `Makefile`?
2) Modify `CI` to use `git-clang-format` so that it only enforces relevant changes.


.clang-format
```diff
@@ -19,14 +19,17 @@ BreakBeforeTernaryOperators: 'true'
...
 SpacesInSquareBrackets: 'false'
 Standard: Cpp11
 UseTab: Never
-SortIncludes: false
\ No newline at end of file
+SortIncludes: false
+ColumnLimit: 100
+
+# when wrapping function calls/declarations, force each parameter to have its own line
+BinPackParameters: 'true'
```
notlesh, line 35: Example:
```c++
void foo(int arg1,
         int arg2,
         int arg3);
```

Notice that this isn't allowed with this rule:
```c++
void foo(int arg1,
         int arg2, int arg3);
```

If such a declaration/function call wraps, this rule enforces that each param is on its own line. Feel free to object.

notlesh:
This is a good resource for anyone who wants to play with style rules:

https://zed0.co.uk/clang-format-configurator/

notlesh:
Note: the pre-commit hook mentioned above will use whatever version `clang-format` happens to resolve to -- see the original message for workarounds (since `lokinet` uses version `8`, which is superceded by version `9` and `11` now)

notlesh:
TODO:

`clang-format-8` and `clang-format-9` seem to disagree on some corner cases where a function declaration has both a template declaration and a return type. I haven't found the magic incantation to get the hook to use a specific version (it may involve both `clang-format` and `clang-format-diff`).

However, it seems that version 9 has correct behavior -- so perhaps we want to move to `clang-format-9` anyway.

PR #1216, opened 2020-04-03
The git blame wall 

notlesh:
This is what a codebase-wide change to `clang-format` rules would do...


notlesh:
Closing this, as it was absorbed by #1215.

PR #1226, opened 2020-04-15
link layer re-transmission reduction

majestrate:
reduce number of acks in a multi ack to reduce the damage done by a dropped multi ack to mitigate excessive retransmissions.


tewinget:
Seems like a reasonable idea.  I do wonder if there's a better way to choose a number though, or if it will just be trial and error.

PR #1227, opened 2020-04-16
hotfix: make user install VC2015 runtime

despair86:
this will be patched away before a 0.7.1+ release tho as i can now build everything with g++


despair86:
requested by @KeeJef

KeeJef:
Basically just checks for VC2015 Runtime and if its not found tells the user to install it from X link and start the installer again.

PR #1228, opened 2020-04-16
Win32 cleanup

despair86:
remove a ton of ded code from early win32 stuff


despair86:
also fixes the pybind build: if we're cross-compiling we may not have a python interpreter available for the target machine so don't even build pybind itself if we've already excluded the pybind code on the lokinet side

despair86:
holy hell this sped up windows quite a bit
getting 4 Mbps down, ~1 Mbps up now

majestrate:
i liek the number of deleted lines

despair86:
coming soon: `llarp/win32/wintun.c`

KeeJef:
Can we get this reviewed and merged, this gives a significant perf increase in upload speed for windows which would be good to get in in time for the 0.7.0 push

PR #1240, opened 2020-04-22
fix ci for win32

majestrate:
this line removed was causing a compile error


jagerman:
Should cherry-pick this into a PR for `dev` as well.

PR #1246, opened 2020-04-27
Snapp config cleanup 2020 04 24

notlesh:
Last piece of config overhaul.

I also plan to enforce that only one service is allowed per lokinet instance.


llarp/router/router.cpp
```diff
@@ -421,36 +424,25 @@ namespace llarp
       m_isServiceNode = true;
     }
 
+    /// build a set of  strictConnectPubkeys (
+    /// TODO: make this consistent with config -- do we support multiple strict connections
```
majestrate, line 428: yes we do support multiple strict connections

notlesh: We weren't in this codepath. I suspect it wasn't actually possible, although code downstream supported it (e.g. taking a `std::set<>` of strict connect keys).

majestrate:
can we not gut `llarp::Addr` yet?

majestrate:
(this PR is already really big and i am blocked on this being merged so it can unbreak dev)

PR #1248, opened 2020-04-29
Config fixes 2020 04 29

notlesh:
Fixes an issue where `[bind]` section wasn't accepting arbitrary interface names


despair86:
> Fixes an issue where `[bind]` section wasn't accepting arbitrary interface names

correct, on windows this takes a UUID wrapped in curly braces as the RH argument instead of a interface name

PR #1251, opened 2020-05-02
windows fix wchar/char path conversions

despair86:
adds the cmake detection logic for win32 and fixes those pesky path to string conversions which are implicit on every other platform but must be explicit on win32 since it exposes the underlying filesystem character encoding (`wchar_t`/`char16_t` vs `char`)

### DIFF
--- a/cmake/win32.cmake
+++ b/cmake/win32.cmake
@@ -2,6 +2,8 @@ if(NOT WIN32)
   return()
 endif()
 
+include(CheckCXXSourceCompiles)
+
 enable_language(RC)
 
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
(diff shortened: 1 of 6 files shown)


llarp/context.cpp
```diff
@@ -51,7 +51,7 @@ namespace llarp
       jobQueueSize = 1024;
     logic = std::make_shared<Logic>(jobQueueSize);
 
-    nodedb_dir = config->router.m_dataDir / nodedb_dirname;
+    nodedb_dir = fs::path(config->router.m_dataDir / nodedb_dirname).string();
```
despair86, line 54: this is the weirdest conversion

daemon/main.cpp
```diff
@@ -216,7 +216,7 @@ main(int argc, char* argv[])
   {
     llarp::ensureConfig(
         llarp::GetDefaultDataDir(), llarp::GetDefaultConfigPath(), overwrite, opts.isRelay);
-    conffname = llarp::GetDefaultConfigPath().c_str();
+    conffname = llarp::GetDefaultConfigPath().string().c_str();
```
majestrate, line 219: this string will fall out of scope

cmake/win32.cmake
```diff
@@ -47,6 +49,39 @@ if (STATIC_LINK_RUNTIME)
...
+    check_cxx_source_compiles("${filesystem_code}" filesystem_compiled_${fslib})
+    if (filesystem_compiled_${fslib})
+      message(STATUS "Using -l${fslib} for std::filesystem support")
+      target_link_libraries(filesystem INTERFACE ${fslib})
+      break()
+    endif()
+  endforeach()
+endif()
+unset(CMAKE_REQUIRED_LIBRARIES)
+
```
jagerman, line 84: This looks identical to the code in `unix.cmake` and needs to be DRY'ed out.

daemon/main.cpp
jagerman: `conffname` is a `std::string` so this is better as just `.string();` anyway (i.e. just drop the `.c_str()`).

llarp/config/key_manager.cpp
```diff
@@ -61,7 +61,7 @@ namespace llarp
     m_lokidRPCPassword = config.lokid.lokidRPCPassword;
 
     RouterContact rc;
-    bool exists = rc.Read(m_rcPath.c_str());
+    bool exists = rc.Read(m_rcPath.string().c_str());
```
jagerman, line 64: I think this would be nicer to just have `RouterContact::Read` take a `fs::path` instead of a c-str instead (and then have it do the path -> c-string conversion if needed).

llarp/config/key_manager.cpp
```diff
@@ -186,7 +186,7 @@ namespace llarp
   bool
   KeyManager::backupKeyFilesByMoving() const
   {
-    std::vector<std::string> files = {m_rcPath, m_idKeyPath, m_encKeyPath, m_transportKeyPath};
+    std::vector<std::string> files = {m_rcPath.string(), m_idKeyPath.string(), m_encKeyPath.string(), m_transportKeyPath.string()};
```
jagerman, line 189: Here, too, makes a lot more sense as a vector of `fs::path` rather than string.  These feel like we are doing the path-to-string conversion way too early by forcing to caller to do it rather than making the callee take a proper `path` object.

jagerman:
@despair86 - I'll hold off on this; we're likely going to refactor these config settings to get loaded into a `fs::path` and keep it in it as long as possible (i.e. not convert to string), which should eliminate a lot of the issues here.

PR #1255, opened 2020-05-04
Add [network]:enabled to ignored settings list

jagerman:
No one knows what this option does or did, but it is in some generated
config files.


jagerman:
Closing (included in #1256)

PR #1257, opened 2020-05-04
Gut Addr, replace with stubbed SockAddr and IpAddress

notlesh:
This guts the plagued `Addr` class and separates it into a couple different classes:

* `IpAddress` for representing an ip address
* `SockAddr` for providing a wrapper around `sockaddr_in`


llarp/net/net_addr.cpp
```diff
@@ -1,505 +0,0 @@
...
-// for addrinfo
-#ifndef _WIN32
-#include <sys/types.h>
-#include <sys/socket.h>
-#include <netdb.h>
-#else
-#include <winsock2.h>
-#include <ws2tcpip.h>
-#include <wspiapi.h>
-#define inet_aton(x, y) inet_pton(AF_INET, x, y)
```
despair86, line 14: @notlesh you forgot to retain these

llarp/net/sock_addr.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <netinet/in.h>
+
```
despair86, line 4: over here

jagerman: Do they need to be in the .hpp?

llarp/net/net_addr.cpp
notlesh: Thanks for pointing that out, it's hard to catch with `#ifdef`. I intend to migrate some of this stuff to `SockAddr`, so I'll keep this in mind.

llarp/net/sock_addr.hpp
despair86: we declared a field of `sockaddr_in` right in the class declaration

llarp/net/net.cpp
```diff
@@ -533,36 +537,45 @@ namespace llarp
...
+  std::optional<IpAddress>
   GetIFAddr(const std::string& ifname, int af)
   {
     sockaddr_storage s;
     sockaddr* sptr = (sockaddr*)&s;
     if (!llarp_getifaddr(ifname.c_str(), af, sptr))
       return std::nullopt;
-    return llarp::Addr{*sptr};
+    // TODO: why the fuck does this not compile?
+    // llarp::SockAddr saddr = SockAddr(*sptr);
```
notlesh, line 548: FYI this might be a compiler bug. I get errors about missing definition for constructor that takes `sockaddr&`, even when the definition was provided. I couldn't reproduce in a test case. If I provided the definition with the declaration (e.g. in the `.hpp` file) the compiler was happy.

llarp/net/sock_addr.hpp
```diff
@@ -0,0 +1,60 @@
...
+    SockAddr(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint16_t port);
+    SockAddr(std::string_view addr);
+
+    SockAddr(const SockAddr&);
+    SockAddr&
+    operator=(const SockAddr&) const;
+
+    SockAddr(const sockaddr& addr);
+    SockAddr(const sockaddr_in& addr);
+    operator const sockaddr*() const;
```
jagerman, line 25: I think this should be `operator sockaddr const*() const`, i.e. return a pointer to a const value.

jagerman: Err, `operator sockaddr* const() const`

jagerman: Wait, no, I am completely wrong about this.  Need more coffee.

notlesh: Heh, I struggled with that a bit yesterday. Boo pointers....

notlesh:
#1261 tries to get this up and running with some fairly ugly implementation

PR #1258, opened 2020-05-06
unbreak router hive

majestrate:


majestrate:
llarp/config/config.cpp:
respect [network]:type option

llarp/handlers/exit.cpp:
when [network]:type is null dont init tun interface

llarp/service/context.cpp:
respect [network]:type option
change endpoint name back to "default"

llarp/tooling/router_hive.cpp:
dont use LogicCall for obtaining RCs from underlying relays, it crashes the mainloop and it's probably safe to readonly access RCs.

pybind/common.hpp:
remove typecasters as we use C++17 now

pybind/llarp/config.cpp:
remove SnappConfig
wire up NetworkConfig

pybind/llarp/handlers/pyhandler.hpp:
remove SnappConfig from constructor

pybind/llarp/handlers/pyhandler.cpp:
update constructor implementation to match header

test/hive/hive.py:
remove broke endpoint related code
wire up null endpoint option using NetworkConfig
use index at 0 for relays and clients instead of 1
dont add a python endpoint to all clients

notlesh:
Mmmmm, I love a PR that removes more code than it adds...

PR #1260, opened 2020-05-11
perf improvements

majestrate:
* increase path testing interval, reduces idle bandwidth use.
* stop using libuv tick based event loop, replace with blocking run which speeds up throughput and reduces cpu/power use on idle.

### DIFF
--- a/llarp/path/pathset.cpp
+++ b/llarp/path/pathset.cpp
@@ -180,9 +180,14 @@ namespace llarp
         }
         ++itr;
       }
-      std::random_device rd;
-      std::mt19937 g(rd());
-      std::shuffle(chosen.begin(), chosen.end(), g);
+      if (chosen.empty())
+        return nullptr;
+      if (chosen.size() >= 2)
+      {
+        std::random_device rd;
+        std::mt19937 g(rd());
+        std::shuffle(chosen.begin(), chosen.end(), g);
+      }
       return chosen[0];
     }
 
--- a/llarp/path/pathset.cpp
+++ b/llarp/path/pathset.cpp
@@ -5,6 +5,8 @@
 #include <routing/dht_message.hpp>
 #include <router/abstractrouter.hpp>
 
+#include <random>
+
 namespace llarp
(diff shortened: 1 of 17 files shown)


llarp/router/i_outbound_message_handler.hpp
```diff
@@ -24,9 +24,9 @@ namespace llarp
 
   using SendStatusHandler = std::function<void(SendStatus)>;
 
-  static const size_t MAX_PATH_QUEUE_SIZE = 50;
-  static const size_t MAX_OUTBOUND_QUEUE_SIZE = 200;
-  static const size_t MAX_OUTBOUND_MESSAGES_PER_TICK = 100;
+  static const size_t MAX_PATH_QUEUE_SIZE = 100;
+  static const size_t MAX_OUTBOUND_QUEUE_SIZE = 1000;
+  static const size_t MAX_OUTBOUND_MESSAGES_PER_TICK = 500;
```
notlesh, line 30: If these queue sizes are really improving things, then we should probably take a closer look at them. An easy start would be to print out any time we hit a new ceiling on size, then make sure we aren't filling the queue.

Also, I'd be interested to see if there's a direct correlation between queue size and bandwidth being pushed. Assuming this is the case, we don't want to unconditionally tune our queue size for someone with a large pipe (1Gbit, let's say) as  this would be inappropriate for someone with a small pipe (5Mbit, for example).

llarp/path/pathset.cpp
```diff
@@ -180,9 +180,14 @@ namespace llarp
         }
         ++itr;
       }
-      std::random_device rd;
-      std::mt19937 g(rd());
-      std::shuffle(chosen.begin(), chosen.end(), g);
+      if (chosen.empty())
+        return nullptr;
+      if (chosen.size() >= 2)
```
tewinget, line 185: Just get a random int in range and return chosen[randint]

majestrate: ok

PR #1266, opened 2020-05-14
default to port 53 in dns section entries

### LINKED ISSUE #1264: Setting upstream to a local IP does not work
I use this in my dns settings:
```
[dns]
upstream=10.0.0.1
bind=127.3.2.1:53
```
but this fails: all DNS requests via 127.3.2.1 time out (but DNS requests directly to 10.0.0.1 work fine).  This setup also works fine on master.  If I use some public IP like 1.1.1.1 it works fine (but then I can't resolve local network hostnames, which my 10.0.0.1 router provides).

At a guess: I think perhaps the configuration DNS settings is disallowing private addresses for the upstream DNS setting?  If so, it shouldn't: they are valid.

majestrate:
IpAddresses in dns section config were not defaulting to port 53 when no port was provided.

Fixes #1264

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -228,22 +228,20 @@ namespace llarp
   DnsConfig::defineConfigOptions(ConfigDefinition& conf, const ConfigGenParameters& params)
   {
     (void)params;
-
-    conf.defineOption<std::string>(
-        "dns", "upstream", false, true, std::nullopt, [=](std::string arg) {
-          IpAddress addr{arg};
-          const auto maybePort = addr.getPort();
-          if (not maybePort.has_value())
-            addr.setPort(53);
-          m_upstreamDNS.push_back(std::move(addr));
-        });
-
-    conf.defineOption<std::string>("dns", "bind", false, std::nullopt, [=](std::string arg) {
+    static auto parseDNSAddr = [](auto arg) {
       IpAddress addr{arg};
       const auto maybePort = addr.getPort();
       if (not maybePort.has_value())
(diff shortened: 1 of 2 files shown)


llarp/config/config.cpp
```diff
@@ -231,11 +231,20 @@ namespace llarp
...
+        });
+
+    conf.defineOption<std::string>("dns", "bind", false, std::nullopt, [=](std::string arg) {
+      IpAddress addr{arg};
+      const auto maybePort = addr.getPort();
+      if (not maybePort.has_value())
+        addr.setPort(53);
+      m_bind = std::move(addr);
+    });
   }
```
notlesh, line 248:
```diff
+    auto parseDnsAddr = [](std::string arg) {
+      IpAddress addr{arg};
+      if (not addr.getPort())
+        addr.setPort(53);
+      return addr;
+    };
+
     conf.defineOption<std::string>(
         "dns", "upstream", false, true, std::nullopt, [=](std::string arg) {
-          m_upstreamDNS.push_back(IpAddress(arg));
+          m_upstreamDNS.push_back(parseDnsAddr(arg));
         });

     conf.defineOption<std::string>(
-        "dns", "bind", false, std::nullopt, [=](std::string arg) { m_bind = IpAddress(arg); });
+        "dns", "bind", false, std::nullopt, [=](std::string arg) { m_bind = parseDnsAddr(arg); });
   }
```

notlesh: Apparently suggestions don't work across deleted sections...

PR #1269, opened 2020-05-15
Make unit tests and CI pass

majestrate:


cmake/check_for_std_optional.cmake
```diff
@@ -0,0 +1,18 @@
...
+   return maybe.value() == 1;
+}
+]])
+
+check_cxx_source_compiles("${std_optional_code}" was_compiled)
+if(was_compiled)
+  message(STATUS "we have std::optional")
+else()
+  message(FATAL_ERROR "we dont have std::optional your compiler is broke af")
+endif()
```
jagerman, line 18: Is this really needed?  We can just let the compilation break instead.

majestrate: i added this so it short circuits travis mac ci until we unbreak that so they don't take as long.

PR #1270, opened 2020-05-15
relicense to GPL3

majestrate:
this officially makes lokinet GPL3


LICENSE.txt
```diff
@@ -1,23 +1,683 @@
-LokiNET is the reference implementation of LLARP (Low Latency Anonymous 
-Routing Protocol) and is licensed under ZLIB license, however the protocol 
-specifications are placed into the public domian using the CC0 License.
+LokiNET is the reference implementation of LLARP (Low Latency Anonymous
+Routing Protocol); this distribution of LokiNET and the Lokinet control
+panel is licensed under the GNU General Public License, version 3.
```
jagerman, line 3: You don't want this control panel stuff; that was there just for the control panel itself.  Instead:

`is the reference implementation of LLARP (Low Latency Anonymous
Routing Protocol).  LokiNET is licensed under the GNU General Public License, version 3.`

jagerman: Actually just using the GNU-recommended preamble might be better:

    LokiNET is the reference implementation of LLARP (Low Latency Anonymous 
Routing Protocol).

    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.

(Do we want the "or later" part?)

majestrate:
something to think about, LGPL3 allows linking to non gpl code, may want to consider that instead of flat gpl3

jagerman:
> something to think about, LGPL3 allows linking to non gpl code, may want to consider that instead of flat gpl3

For lokinet as it currently exists that won't make any difference.  We probably *do* want to separate the full application from the core logic though, though, and when we do so we can consider whether to license the library parts as LGPL.

PR #1273, opened 2020-05-18
Drone CI + cmake + fixes

jagerman:
This PR converts the common linux tests for AMD64 and ARM from travis-ci to a self-hosted Drone CI.

The tests are running fine on debian sid and ubuntu focal, but on bionic and debian buster I get:
```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] KeyManagerTest.TestInitialize_MakesKeyfiles
[  FAILED  ] KeyManagerTest.TestInitialize_DetectsBadRcFile
```
(Sometimes it's just one, or the other).  Cc @notlesh to take a look at that.

Along the way I found a bunch of warnings and small issues that needed fixing, and then got sucked into some cmake cleanup, which ended up being the biggest part of this PR (that commit message repeated here for convenience):

---
Refactors many things in cmake to improve and simplify:

- don't use variable indirection for target names; target names are
*already* a variable of sorts.  (e.g. ${UTIL_LIB} is now just
lokinet-util).  cmake/basic_definitions.cmake is now gone.

- fix LTO enabling to use the standard cmake (3.9+) LTO mechanism rather
than shoving a bunch of flag hacks through link_libraries and
add_compile_options.  This also now enables LTO when building a shared
library (because previously the -flto hacks were only turned on in the
static code for some reason).

- build liblokinet as *either* shared library or static library, but not
both.  Building both makes things more complicated because they had
different names (lokinet-shared or lokinet-static) and seems pointless:
you generally want one or the other.  Now there is just the liblokinet
target, which will be shared or static depending on the value of
BUILD_SHARED_LIBS.

- Simplify lokinet-cryptography AVX2 code: just build *one* library, and
add in the additional AVX2 files when possible, rather than building two
and needing to merge them.

- Compress STATIC_LINK and STATIC_LINK_RUNTIME into just STATIC_LINK.
It makes no sense to use one of these (_RUNTIME) on Windows and the
other on non-Windows when they appear to try to do the same thing.

- remove a bunch of annotations from `endif(FOO)` -> `endif()`.

- move all the tuntap compilation code (including OS-specific source
file selection) into vendor/CMakeLists.txt and build tuntap as an
intermediate OBJECT library rather than keeping a global variable in 5
different files.

- move release motto define to root cmake; it made no sense being
duplicated in both unix.cmake and win32.cmake

- fix add_log_tag to not stomp on any existing source compile flags with
its definition.  Also use proper compile definition property instead of
cramming it into compile flags.

- make optimization/linker flags less hacky.  There's no reason for us
to force particular optimization flags because the cmake build type
already does that (e.g. -DCMAKE_BUILD_TYPE=Release does -O3).  Not doing
that also silences a bunch of cmake warnings because it thinks "-O0 -g3"
etc.  are link libraries (which is reasonable: that's what the code was
telling cmake they are).

- sets the default build type to RelWithDebInfo which gives us `-O2 -g`
if you don't specify a build type.

- Move PIC up (so that the things loaded in unix.cmake, notably libuv,
have it set).

- Add a custom `curl` interface library that carries the correct link
target and include paths for curl (system or bundled).
---


cmake/enable_lto.cmake
```diff
@@ -0,0 +1,14 @@
+# -flto
+include(CheckIPOSupported)
+check_ipo_supported(RESULT IPO_ENABLED OUTPUT ipo_error)
+if(IPO_ENABLED)
```
majestrate, line 4: i didn't know we were going to have an initial public offering .... :^)

PR #1276, opened 2020-05-18
fix sockaddr on win32

despair86:


despair86:
also took the liberty of collapsing the unix header includes to the same place where all downstream modules can see it

PR #1272, opened 2020-05-16
exit traffic via snapps

majestrate:
this makes the codepath for exit traffic via snapp work, adds new protocol type for exit traffic.

PR #1277, opened 2020-05-18
force cmake to run test in C++17 mode

despair86:
on my sun cmake kept running the compiler in g++1y mode

### DIFF
--- a/cmake/check_for_std_filesystem.cmake
+++ b/cmake/check_for_std_filesystem.cmake
@@ -6,6 +6,7 @@
 
 add_library(filesystem INTERFACE)
 
+# why does cmake insist on using C++14 for this
 set(filesystem_code [[
 #include <filesystem>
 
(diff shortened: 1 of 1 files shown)


cmake/check_for_std_filesystem.cmake
```diff
@@ -15,6 +16,8 @@ int main() {
 }
 ]])
 
+set(CMAKE_CXX_STANDARD 17)
+
```
jagerman, line 20: Is this code running before we set the variable in the top-level CMakeLists?  Otherwise I can't see how this helps.

despair86: i have no idea how cmake orders this
the `CMAKE_CXX_STANDARD` is indeed set to C++1z before this module is included

despair86:
leaving this as a draft in case @majestrate has something else in the pipline

majestrate:
it's in https://github.com/loki-project/loki-network/pull/1272/commits/f2999d1e74a9ede445265fbd6a92afe92fa64a4b with #1272

PR #1279, opened 2020-05-19
Fix cmake curl definition

jagerman:


despair86:
I think this forced me to recompile my local libcurl without SSL

I used to specify the link line inside `CURL_LIBRARY_*` which broke when it turned into an interface library, so `CURL_LIBRARY_RELEASE='-L/usr/local/lib -lcurl -lssl -lcrypto'` for example

jagerman:
Why not just build it shared?

PR #1281, opened 2020-05-20
fix wstring path conversions

despair86:
aw crap i lost the other half of this where i started doing the refactoring in #1251

### DIFF
--- a/llarp/context.cpp
+++ b/llarp/context.cpp
@@ -50,7 +50,7 @@ namespace llarp
       jobQueueSize = 1024;
     logic = std::make_shared<Logic>(jobQueueSize);
 
-    nodedb_dir = config->router.m_dataDir / nodedb_dirname;
+    nodedb_dir = fs::path(config->router.m_dataDir / nodedb_dirname).string();
 
     return true;
   }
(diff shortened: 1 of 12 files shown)


despair86:
this one actually fixes the windows build altogether

majestrate:
this ready?

llarp/context.cpp
```diff
@@ -50,7 +50,7 @@ namespace llarp
       jobQueueSize = 1024;
     logic = std::make_shared<Logic>(jobQueueSize);
 
-    nodedb_dir = config->router.m_dataDir / nodedb_dirname;
+    nodedb_dir = fs::path(config->router.m_dataDir / nodedb_dirname).string();
```
majestrate, line 53: nodedb_dir can probably also be an `fs::path`

despair86: k will do

despair86:
ok this is ready @majestrate

majestrate:
@despair86 revert last commit, i think it was fine before that.

despair86:
done

llarp/util/logging/logger_internal.cpp
```diff
@@ -9,6 +9,6 @@ namespace llarp
   {
     std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds> now{
         std::chrono::milliseconds{ts.now}};
-    return date::operator<<(out, now) << " UTC [+" << ts.delta << "]";
+    return date::operator<<(out, now) << " GMT [+" << ts.delta << "]";
```
jagerman, line 12: Why this change?

notlesh: :thinking:

PR #1297, opened 2020-06-04
Dont touch self.signed if client

notlesh:
This PR causes `router.cpp` to ignore our `self.signed` file if we're a client.

This was causing issues because `KeyManager` tries to do backups if `self.signed` is present and invalid. **We also need to figure out why these RCs are not passing validation.**

### DIFF
--- a/llarp/config/key_manager.cpp
+++ b/llarp/config/key_manager.cpp
@@ -61,7 +61,7 @@ namespace llarp
     m_lokidRPCPassword = config.lokid.lokidRPCPassword;
 
     RouterContact rc;
-    bool exists = rc.Read(m_rcPath.c_str());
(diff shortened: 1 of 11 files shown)


llarp/config/key_manager.cpp
```diff
@@ -61,7 +61,7 @@ namespace llarp
     m_lokidRPCPassword = config.lokid.lokidRPCPassword;
 
     RouterContact rc;
-    bool exists = rc.Read(m_rcPath);
+    bool exists = rc.Read(m_rcPath.c_str());
```
majestrate, line 64: why?

notlesh: Derp.

PR #1301, opened 2020-06-08
add regression test for key backup bug

majestrate:
make sure the bug is squashed with a regression test.

### DIFF
--- a/llarp/config/config.hpp
+++ b/llarp/config/config.hpp
@@ -72,11 +72,11 @@ namespace llarp
     std::string m_ifname;
     std::string m_ifaddr;
 
-    std::string m_keyfile;
+    std::optional<fs::path> m_keyfile;
     std::string m_endpointType;
     bool m_reachable = false;
-    int m_hops = -1;
-    int m_paths = -1;
+    std::optional<int> m_Hops;
+    std::optional<int> m_Paths;
     bool m_AllowExit = false;
     std::set<RouterID> m_snodeBlacklist;
     std::optional<service::Address> m_exitNode;
@@ -152,7 +152,9 @@ namespace llarp
 
   struct BootstrapConfig
   {
-    std::vector<std::string> routers;
+    std::vector<fs::path> routers;
+    /// for unit tests
+    bool skipBootstrap = false;
     void
     defineConfigOptions(ConfigDefinition& conf, const ConfigGenParameters& params);
   };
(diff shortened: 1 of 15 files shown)


llarp/config/config.hpp
```diff
@@ -72,11 +72,11 @@ namespace llarp
...
 
-    std::string m_keyfile;
+    std::optional<fs::path> m_keyfile;
     std::string m_endpointType;
     bool m_reachable = false;
-    int m_hops = -1;
-    int m_paths = -1;
+    std::optional<int> m_Hops;
+    std::optional<int> m_Paths;
     bool m_AllowExit = false;
```
notlesh, line 80: Did this fix something, or is it more of a clean up thing?

llarp/util/logging/logger.hpp
```diff
@@ -60,9 +60,25 @@ namespace llarp
...
+  {
+    LogSilencer();
+    ~LogSilencer();
+    explicit LogSilencer(LogContext& ctx);
+
+   private:
+    LogContext& parent;
+    ILogStream_ptr stream;
+  };
+
```
notlesh, line 75: this is a cool use of RAII :)

test/regress/2020-06-08-key-backup-bug.cpp
```diff
@@ -0,0 +1,62 @@
...
+  config->impl.bootstrap.skipBootstrap = true;
+  auto ptr = llarp_main_init_from_config(config, false);
+  llarp_config_free(config);
+  return ptr;
+}
+
+TEST_CASE("key backup bug regression test", "[regress]")
+{
+  llarp::LogSilencer shutup;
+  for (const fs::path& path : {"regress-1.private", "regress-2.private", ""})
```
notlesh, line 23: Some comments about what these nested loops are doing would help

llarp/config/config.hpp
majestrate: it's a cleanup

majestrate:
> It's not clear from the regression test what the bug is or what the test is trying to do. I'm guessing the bug had something to do with an invalid path or empty string being used as the keyfile?

this was the missing/invalid self.signed reseting the hidden service keys bug, i wanted to add a regression test to make sure it's gone.

test/regress/2020-06-08-key-backup-bug.cpp
majestrate: added comments

PR #1302, opened 2020-06-08
add pybind to CI

majestrate:
make pybind compile again.
add pybind and pytest to run in CI so it doesn't bitrot as much.
depends on #1301


tewinget:
looks alright, though some of the config-related changes feel like they're a bit out of scope for this PR.

majestrate:
rebased

PR #1303, opened 2020-06-08
Added runit init and log script

wratc:
Runs as user loki, needs lokinet-bootstrap run as loki before and logs to /var/log/lokinet.

### DIFF
--- a/contrib/runit/run
+++ b/contrib/runit/run
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec chpst -u loki:loki /usr/local/bin/lokinet ~loki/.lokinet/lokinet.ini 2>&1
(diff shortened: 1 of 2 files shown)


contrib/runit/run
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec chpst -u loki:loki /usr/local/bin/lokinet ~loki/.lokinet/lokinet.ini 2>&1
```
majestrate, line 2: shouldn't it be `~/.lokinet/lokinet.ini` instead of `~loki/.lokinet/lokinet.ini` or is this some runit-ism?

majestrate:
if you can please also provide a readme on how to use this. thnx.

contrib/runit/run
wratc: It's a chpst-ism because home dir is still /root I think.

majestrate: i see

wratc:
Do I open a different mr for that? And should I put the two run files in it again?

majestrate:
you can add it in a new commit to your branch

majestrate:
do we have any updates on documentation on how to use this?

PR #1306, opened 2020-06-11
LokiMQ

majestrate:
* fully replace `libabyss` with lokimq
* fully replace `llarp::thread::ThreadPool` with lokimq


Makefile
```diff
@@ -311,7 +310,7 @@ mac: mac-release
 	$(MAKE) -C '$(BUILD_ROOT)' package
 
 format:
-	$(FORMAT) -i $$(find jni daemon llarp include libabyss pybind | grep -E '\.[hc](pp)?$$')
+	$(FORMAT) -i $$(find jni daemon llarp include pybind | grep -E '\.[h,c](pp)?$$')
```
jagerman, line 313: This `,` seems wrong unless we have some files named `foo.,pp`

llarp/router/router.cpp
```diff
@@ -811,29 +832,30 @@ namespace llarp
...
       {
         rpcBindAddr = DefaultRPCBindAddr;
       }
-      rpcServer = std::make_unique<rpc::Server>(this);
-      if (not rpcServer->Start(rpcBindAddr))
+
+      // older configs just specify an ip/port tuple so check for that and prepend tcp:// in that
+      // case
+      if (rpcBindAddr.find("tcp://") == std::string::npos
+          and rpcBindAddr.find("ipc://") == std::string::npos)
```
jagerman, line 853: I would make this a little more general by just looking for the `://`: `if (rpcBindAddr.find("://") == std::string::npos)` because there are some other address types like `curve://` or `ipc+curve://` which can be used with LMQ if you want to connect to a curve-encrypted listener.

llarp/router/router.hpp
```diff
@@ -151,31 +165,30 @@ namespace llarp
       return _netloop;
     }
 
-    std::shared_ptr<llarp::thread::ThreadPool>
-    threadpool() override
+    void
+    QueueWork(std::function<void(void)> func) override
```
jagerman, line 169: The inner `void` isn't needed.

llarp/router/router.hpp
```diff
@@ -151,31 +165,30 @@ namespace llarp
...
     {
-      return cryptoworker;
+      m_lmq->job(std::move(func));
     }
 
-    std::shared_ptr<llarp::thread::ThreadPool>
-    diskworker() override
+    void
+    QueueDiskIO(std::function<void(void)> func) override
```
jagerman, line 175: Likewise

llarp/router/router.hpp
```diff
@@ -242,15 +255,16 @@ namespace llarp
     NetworkConfig networkConfig;
     DnsConfig dnsConfig;
 
-    const IpAddress DefaultRPCBindAddr = IpAddress("127.0.0.1:1190");
+    const std::string DefaultRPCBindAddr = "tcp://127.0.0.1:1190";
```
jagerman, line 258: You could perhaps switch this to a `lokimq::address`.

llarp/router/router.hpp
```diff
@@ -242,15 +255,16 @@ namespace llarp
...
     const llarp_time_t _randomStartDelay;
 
-    /// lokid caller
-    std::unique_ptr<rpc::Caller> rpcCaller;
-    IpAddress lokidRPCAddr = IpAddress("127.0.0.1:22023");
+    std::shared_ptr<rpc::LokidRpcClient> m_lokidRpcClient;
+
+    std::string lokidRPCAddr = "ipc://loki.sock";
     std::string lokidRPCUser;
     std::string lokidRPCPassword;
```
jagerman, line 269: These later two aren't needed anymore.  (If you needed to do authentication you still can via a pubkey embedded in the lokidRPCAddr).

llarp/router/router.hpp
```diff
@@ -242,15 +255,16 @@ namespace llarp
...
+    std::unique_ptr<rpc::RpcServer> m_RPCServer;
+
     const llarp_time_t _randomStartDelay;
 
-    /// lokid caller
-    std::unique_ptr<rpc::Caller> rpcCaller;
-    IpAddress lokidRPCAddr = IpAddress("127.0.0.1:22023");
+    std::shared_ptr<rpc::LokidRpcClient> m_lokidRpcClient;
+
+    std::string lokidRPCAddr = "ipc://loki.sock";
```
jagerman, line 267: Can be a lokimq::address.  This is going to open lock.sock in pwd, is that the right behaviour? (i.e. does the process chdir there?)

jagerman: Also lokimq::address has a factory function designed for this case:
```C++
auto lokidRPCAddr = lokimq::address::ipc("loki.sock");
```

llarp/router/router.hpp
jagerman: Can also be written as:
```C++
const auto DefaultRPCBindAddr = lokimq::address::tcp("127.0.0.1", 1190);
```

llarp/rpc/endpoint_rpc.hpp
```diff
@@ -0,0 +1,45 @@
...
+    Start();
+
+    void
+    AuthenticateAsync(
+        llarp::service::Address from,
+        service::ConvoTag tag,
+        std::function<void(service::AuthResult)> hook) override;
+
+   private:
+    const std::string m_AuthURL;
```
jagerman, line 38: `lokimq::address`

llarp/router/router.hpp
majestrate: i want to support unix sockets too

llarp/config/config.cpp
```diff
@@ -852,6 +871,46 @@ namespace llarp
             "Permanently map a `.loki` address to an IP owned by the snapp. Example:",
             "mapaddr=whatever.loki:10.0.10.10 # maps `whatever.loki` to `10.0.10.10`.",
         });
+    // extra [network] options
+    // TODO: probably better to create an [exit] section and only allow it for routers
```
tewinget, line 875: should we start using relay/client as terminology, or did we wanna stick with "router"?

llarp/router/router.hpp
jagerman: Right, but this is just for the default value -- a `lokimq::address` can hold either tcp/ipc or the +curve variants.

llarp/router/router.cpp
tewinget: +1

llarp/router/router.hpp
```diff
@@ -295,10 +311,7 @@ namespace llarp
     void
     GossipRCIfNeeded(const RouterContact rc) override;
 
-    Router(
-        std::shared_ptr<llarp::thread::ThreadPool> worker,
-        llarp_ev_loop_ptr __netloop,
-        std::shared_ptr<Logic> logic);
+    explicit Router(llarp_ev_loop_ptr __netloop, std::shared_ptr<Logic> logic);
```
notlesh, line 314: double underscore?

notlesh: oh, that was from previous code... but yeah

llarp/rpc/rpc_server.cpp
```diff
@@ -0,0 +1,41 @@
...
+            "version", [](lokimq::Message& msg) { msg.send_reply(llarp::VERSION_FULL); })
+        .add_request_command("status", [&](lokimq::Message& msg) {
+          std::promise<std::string> result;
+          LogicCall(m_Router->logic(), [&result, r = m_Router]() {
+            const auto state = r->ExtractStatus();
+            result.set_value(state.dump());
+          });
+          auto ftr = result.get_future();
+          msg.send_reply(ftr.get());
+        });
```
notlesh, line 38: this is really nice... :+1:

llarp/service/auth.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <string>
+#include <functional>
+#include "address.hpp"
+#include "handler.hpp"
+#include <crypto/types.hpp>
+
+namespace llarp::service
+{
+  /// authentication status
+  enum AuthResult
```
notlesh, line 12: enum class?

llarp/service/protocol.cpp
```diff
@@ -92,6 +96,32 @@ namespace llarp
       return bencode_end(buf);
     }
 
+    std::optional<std::vector<byte_t>>
+    ProtocolMessage::MaybeEncodeAuthInfo() const
```
notlesh, line 100: would throwing be better here? returning `nullopt` on all errors seems like it will make it impossible for the caller to know what went wrong (which might matter in debugging if nothing else).

llarp/router/router.hpp
jagerman: You should go epee style and use 5 underscores.

llarp/service/protocol.cpp
jagerman: Agree on this - `std::optional` is conceptually not meant for errors, but rather for optional values where empty means you can still go ahead.

include/llarp.h
```diff
@@ -32,9 +32,17 @@ extern "C"
   /// llarp_application config
   struct llarp_config;
 
-  /// get default config for current platform
   struct llarp_config*
-  llarp_default_config();
+  llarp_default_client_config();
+
+  struct llarp_config*
+  llarp_default_relay_config();
```
jagerman, line 39: relay -> router

llarp/config/config.cpp
```diff
@@ -175,23 +194,37 @@ namespace llarp
     });
 
     conf.defineOption<int>("network", "paths", false, PathsDefault, [this](int arg) {
-      if (arg < 1 or arg > 8)
-        throw std::invalid_argument("[endpoint]:paths must be >= 1 and <= 8");
+      if (arg < 2 or arg > 8)
```
jagerman, line 197: Could use a comment here to explain why these 2/8 values.

jagerman: (Or alternatively self-describing constants)

llarp/config/config.cpp
```diff
@@ -175,23 +194,37 @@ namespace llarp
...
+        throw std::invalid_argument("[endpoint]:paths must be >= 2 and <= 8");
       m_Paths = arg;
     });
 
     conf.defineOption<std::string>("network", "exit-node", false, "", [this](std::string arg) {
       if (arg.empty())
         return;
       service::Address exit;
+      IPRange range;
+      const auto pos = arg.find(":");
```
jagerman, line 207: This `:`-separated range support could use some user documentation somewhere.

llarp/config/config.cpp
```diff
@@ -175,23 +194,37 @@ namespace llarp
...
+      {
+        range.FromString("0.0.0.0/0");
+      }
+      else if (not range.FromString(arg.substr(pos + 1)))
+      {
+        throw std::invalid_argument("[network]:exit-node invalid ip range for exit provided");
+      }
+      if (pos != std::string::npos)
+      {
+        arg = arg.substr(0, pos);
```
jagerman, line 218: `arg.resize(pos);` avoids a copy and replace and is marginally simpler.

llarp/config/config.cpp
```diff
@@ -175,23 +194,37 @@ namespace llarp
...
+      IPRange range;
+      const auto pos = arg.find(":");
+      if (pos == std::string::npos)
+      {
+        range.FromString("0.0.0.0/0");
+      }
+      else if (not range.FromString(arg.substr(pos + 1)))
+      {
+        throw std::invalid_argument("[network]:exit-node invalid ip range for exit provided");
+      }
```
jagerman, line 215: So this lets me write something like:

```
exit-node=abc.loki:8.0.0.0/8
exit-node=def.loki:14.2.0.0/16
```
but a couple questions:
1. What happens if the ranges overlap, either partially such as:
```
exit-node=abc.loki:8.0.0.0/8
exit-node=def.loki   # (implied :0.0.0.0/0 default)
```
or entirely:
```
exit-node=abc.loki:8.0.0.0/8
exit-node=def.loki:8.0.0.0/8
```
2. Does this support using the same exit multiple times, so that I can map multiple subnets to an exit?
```
exit-node=abc.loki:8.0.0.0/8
exit-node=abc.loki:123.0.0.0/8
```

(This could use some comment, at least in the code, explaining how this is supposed to work, which exits get priority, etc.)

llarp/config/config.cpp
```diff
@@ -852,6 +887,46 @@ namespace llarp
...
+            "lmq endpoint to talk to for authenticating new sessions",
+            "ipc:///var/lib/lokinet/auth.socket",
+        });
+
+    def.addOptionComments(
+        "network",
+        "auth-method",
+        {
+            "lmq function to call for authenticating new sessions",
+            "llarp.auth",
```
jagerman, line 921: `llarp.auth` feels a little too generic as an endpoint name; how about `exit.auth`, or `llarp-exit.auth`?  I can imagine that this API gets extended at some point in the future, and having just plain `auth` may come to seem a little too generic.

llarp/config/config.cpp
```diff
@@ -852,6 +887,46 @@ namespace llarp
...
+        {
+            "Whether or not we should act as an exit node. Beware that this increases demand",
+            "on the server and may pose liability concerns. Enable at your own risk.",
+        });
+
+    def.addOptionComments(
+        "network",
+        "auth",
+        {
+            "authenticate remote sessions against a whitelist or an external lmq server",
```
jagerman, line 904: \+ "Must additionally specify either 'auth-url' or 'auth-whitelist' settings"

llarp/config/config.cpp
```diff
@@ -852,6 +887,46 @@ namespace llarp
...
+            "true/false",
+        });
+
+    def.addOptionComments(
+        "network",
+        "auth-url",
```
jagerman, line 910: `auth-url` -> `auth-address`.  `url` implies things that ZMQ address strings don't really have.

jagerman: Or maybe `auth-connect`?

llarp/config/config.cpp
```diff
@@ -852,6 +869,46 @@ namespace llarp
...
+        {
+            "lmq endpoint to talk to for authenticating new sessions",
+            "ipc:///var/lib/lokinet/auth.socket",
+        });
+
+    def.addOptionComments(
+        "network",
+        "auth-method",
```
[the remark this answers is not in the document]
jagerman: LMQ docs are probably too generic for this -- we need some documentation somewhere in lokinet as to how that endpoint is supposed to work (i.e. endpoint type [probably request], data it expects, expected return).

llarp/config/config.cpp
```diff
@@ -852,6 +887,46 @@ namespace llarp
...
+            "lmq function to call for authenticating new sessions",
+            "llarp.auth",
+        });
+
+    def.addOptionComments(
+        "network",
+        "auth-whitelist",
+        {
+            "manually add a remote endpoint by .loki address to the access whitelist",
+        });
```
jagerman, line 929: Should also mention: "Cannot be used with `network.auth-url`

llarp/config/config.cpp
```diff
@@ -852,6 +887,46 @@ namespace llarp
...
+        {
+            "authenticate remote sessions against a whitelist or an external lmq server",
+            "true/false",
+        });
+
+    def.addOptionComments(
+        "network",
+        "auth-url",
+        {
+            "lmq endpoint to talk to for authenticating new sessions",
```
jagerman, line 912: \+ cannot be used with auth-whitelist

jagerman: Perhaps reword to: "lmq endpoint to talk to for doing advanced exit session authentication"

llarp/config/config.cpp
```diff
@@ -852,6 +887,46 @@ namespace llarp
...
             "mapaddr=whatever.loki:10.0.10.10 # maps `whatever.loki` to `10.0.10.10`.",
         });
+    // extra [network] options
+    // TODO: probably better to create an [exit] section and only allow it for routers
+    def.addOptionComments(
+        "network",
+        "exit",
+        {
+            "Whether or not we should act as an exit node. Beware that this increases demand",
+            "on the server and may pose liability concerns. Enable at your own risk.",
```
jagerman, line 897: Add: "You will also need to configure the auth-related settings, below."

llarp/config/config.cpp
```diff
@@ -852,6 +887,46 @@ namespace llarp
...
+        });
+
+    def.addOptionComments(
+        "network",
+        "auth",
+        {
+            "authenticate remote sessions against a whitelist or an external lmq server",
+            "true/false",
+        });
```
jagerman, line 906: This feels a little bit more complicated that necessary because we support *both* an LMQ auth server and whitelists.  Simpler is better here: we should support just whitelists for a "simple mode", and for anything else you set up an LMQ auth server (and if it wants to do whitelists plus other stuff, great, go ahead).

So this option, instead of true/false, can be "whitelist/lmq/none".  If whitelist then we use the `auth-whitelist` setting; if `lmq` we use lmq, if none we allow anyone.

Description could be:
"Authentication mode for remote sessions: can use a simple hard-coded whitelist, a separate lmq authentication server, or none to allow anyone."

I think this option either should not have a default at all (and so must be set explicitly whenever `network.exit` is set), or else should default to whitelist with an empty default whitelist (i.e. so it won't allow anything by default, even though exit mode is technically enabled).

llarp/config/config.cpp
jagerman: Also: "can be specified multiple times"

llarp/config/config.cpp
```diff
@@ -896,33 +953,6 @@ namespace llarp
...
-        });
-
-    def.addOptionComments(
-        "network",
-        "exit-blacklist",
-        {
-            "Blacklist of destinations (same format as whitelist).",
-        });
-
     return def.generateINIConfig(true);
```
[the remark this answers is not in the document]
jagerman: It's in generated configs, but commented out:
```ini
exit=false
#exit-blacklist=tcp:25
#exit-whitelist=tcp:*
#exit-whitelist=udp:*
```
I think it's likely that no one uncommented it, but if they *did* it seems reasonable to break on this given that exit nodes are not currently a supported feature.

daemon/lokinet-vpn
```diff
@@ -1,59 +1,141 @@
 #!/usr/bin/env python3
```
jagerman, line 1: This basic working example script is fine, but I think we should have a separate document describing input and output from the LMQ endpoint.  (Can come in a later PR).

llarp/constants/files.hpp
```diff
@@ -19,7 +23,19 @@ namespace llarp
...
+    auto pw = getpwuid(getuid());
+    if (pw and pw->pw_dir)
+    {
+      homedir = pw->pw_dir;
+    }
+    else
+    {
+      // no home dir specified yea idk
+      homedir = "/var/lib/lokinet";
+      return homedir;
```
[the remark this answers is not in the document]
jagerman: Or just throw an exception to crash.

llarp/router/router.cpp
```diff
@@ -266,7 +288,7 @@ namespace llarp
       LogError("RC is invalid, not saving");
       return false;
     }
-    diskworker()->addJob(std::bind(&Router::HandleSaveRC, this));
+    QueueDiskIO(std::bind(&Router::HandleSaveRC, this));
```
jagerman, line 291: std::bind is sort of obsolete now because you can do pretty much everything it does with a lambda, and the lambda often generates more efficient code, especially when the size/number of captures is small:

```C++
QueueDiskIO([this] { HandleSaveRC(); });
```
(it's also shorter).

llarp/context.cpp
```diff
@@ -88,9 +84,11 @@ namespace llarp
     crypto = std::make_unique<sodium::CryptoLibSodium>();
     cryptoManager = std::make_unique<CryptoManager>(crypto.get());
 
-    router = std::make_unique<Router>(worker, mainloop, logic);
+    router = std::make_unique<Router>(mainloop, logic);
 
-    nodedb = std::make_unique<llarp_nodedb>(router->diskworker(), nodedb_dir);
+    nodedb = std::make_unique<llarp_nodedb>(
+        nodedb_dir,
+        [r = router.get()](std::function<void(void)> call) { r->QueueDiskIO(std::move(call)); });
```
jagerman, line 91: `auto` simplification:
```C++
[r = router.get()](auto call) { r->QueueDiskIO(std::move(call)); });
```

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -0,0 +1,202 @@
...
+      std::promise<std::optional<SecretKey>> promise;
+
+      Request(
+          "admin.get_service_privkeys",
+          [self = shared_from_this(), &promise](bool success, std::vector<std::string> data) {
+            if (not success)
+            {
+              LogError("failed to get private key");
+              promise.set_value(std::nullopt);
+              return;
```
jagerman, line 173: Perhaps we can change this to throw on error; right now we fail to set a value, and that trickles back with empty/false returns to `Router::Configure` which then throws.  If we throw via the promise from here we can accomplish the same thing, but provide some more context.

This ought to work:
```C++
      std::promise<SecretKey> promise;

      Request(
          "admin.get_service_privkeys",
          [self = shared_from_this(), &promise](bool success, std::vector<std::string> data) {
            try {
              // all the current code, except the `LogError/promise.set_value(std::nullopt)/return`
              // bits replaced replaced with `throw std::runtime_error{"failed to get private key because [Jesus]"}`;
              // ...
            } catch (...) {
              promise.set_exception(std::current_exception());
            }
          }
```

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -0,0 +1,202 @@
...
+        case eLogNone:
+        default:
+          return lokimq::LogLevel::trace;
+      }
+    }
+
+    LokidRpcClient::LokidRpcClient(LMQ_ptr lmq, AbstractRouter* r)
+        : m_lokiMQ(std::move(lmq)), m_Router(r)
+    {
+      // m_lokiMQ->log_level(toLokiMQLogLevel(LogLevel::Instance().curLevel));
```
jagerman, line 38: I assume this is commented out because the log level gets set elsewhere, but if so the above `toLokiMQLogLevel` is unused and can get removed, too.

llarp/crypto/encrypted_frame.hpp
```diff
@@ -78,12 +77,14 @@ namespace llarp
...
+    using WorkerFunction_t = std::function<void(WorkFunc_t)>;
+
     void
-    AsyncDecrypt(
-        const std::shared_ptr<thread::ThreadPool>& worker, const EncryptedFrame& frame, User_ptr u)
+    AsyncDecrypt(const EncryptedFrame& frame, User_ptr u, WorkerFunction_t worker)
     {
       target = frame;
-      worker->addJob(std::bind(&AsyncFrameDecrypter<User>::Decrypt, this, std::move(u)));
+      worker(std::bind(&AsyncFrameDecrypter<User>::Decrypt, this, std::move(u)));
```
jagerman, line 87: Since this is only called once, we can avoid the extra shared_ptr ref count increment in Decrypt via a mutable lambda:
```C++
worker([this, u=std::move(u)] () mutable { Decrypt(std::move(u)); });
```

llarp/iwp/session.hpp
```diff
@@ -25,7 +26,7 @@ namespace llarp
     /// How long to keep a replay window for
     static constexpr auto ReplayWindow = (ReceivalTimeout * 3) / 2;
     /// How often to acks RX messages
-    static constexpr auto ACKResendInterval = DeliveryTimeout / 2;
+    static constexpr auto ACKResendInterval = DeliveryTimeout / 4;
```
jagerman, line 29: Um?

llarp/net/ip_packet.cpp
```diff
@@ -12,6 +12,22 @@
 #include <algorithm>
 #include <map>
 
+constexpr uint32_t ipv6_flowlabel_mask = 0b00000000000011111111111111111111;
```
jagerman, line 15: Readability improvement:
`0b00000000'00001111'11111111'11111111;`

llarp/net/ip_range_map.hpp
```diff
@@ -30,13 +31,32 @@ namespace llarp
...
         for (const auto& entry : m_Entries)
           functor(entry.second);
       }
 
+      template <typename Visit_t>
+      void
+      ForEachEntry(Visit_t visit) const
+      {
+        for (const auto& entry : m_Entries)
+          visit(entry.first, entry.second);
```
jagerman, line 57: C++17:
```C++
        for (const auto& [range, value] : m_Entries)
          visit(range, value);
```

llarp/net/net.cpp
```diff
@@ -469,49 +495,33 @@ namespace llarp
...
-      const huint32_t hiaddr = ipaddr_ipv4_bits(172, 16, oct, 255);
-      bool hit = false;
-      for (const auto& range : currentRanges)
-      {
-        hit = hit || range.ContainsV4(loaddr) || range.ContainsV4(hiaddr);
-      }
-      if (!hit)
-        return loaddr.ToString() + "/24";
-      ++oct;
+      possibleRanges.emplace_back(IPRange::FromIPv4(172, 16, oct, 1, 24));
```
jagerman, line 518: The private range here is quite a bit bigger - 172.16.x.x through 172.32.x.x.  This might be more usefully used for /16's:
```C++
    for (byte_t oct = 16; oct < 32; ++oct)
    {
      possibleRanges.emplace_back(IPRange::FromIPv4(172, oct, 0, 1, 16));
    }
```

jagerman: Also we should put this above 192.168.x.0/24, and possibly even before 10.x.0.0/16 checks (since 10.x.x.x is pretty commonly used, and thus more likely to run into conflicts).

llarp/nodedb.cpp
```diff
@@ -84,8 +83,7 @@ llarp_nodedb::RemoveIf(std::function<bool(const llarp::RouterContact& rc)> filte
         ++itr;
     }
   }
-
-  disk->addJob(std::bind(&KillRCJobs, files));
+  disk(std::bind(&KillRCJobs, files));
```
jagerman, line 86: `KillRCJobs` is a two line function and is only used here, maybe worth just inlining it here:
```c++
disk([files=std::move(files)] {
  for (const auto& file : files)
    fs::remove(file);
});
```

llarp/nodedb.cpp
```diff
@@ -277,7 +275,7 @@ llarp_nodedb::ShouldSaveToDisk(llarp_time_t now) const
 void
 llarp_nodedb::AsyncFlushToDisk()
 {
-  disk->addJob(std::bind(&llarp_nodedb::SaveAll, this));
+  disk(std::bind(&llarp_nodedb::SaveAll, this));
```
jagerman, line 278: Lambda-ize:
```C++
disk([this] { SaveAll(); });
```

llarp/path/transit_hop.cpp
```diff
@@ -260,9 +260,9 @@ namespace llarp
     {
       if (m_DownstreamQueue && not m_DownstreamQueue->empty())
       {
-        r->threadpool()->addJob([self = shared_from_this(),
-                                 data = std::move(m_DownstreamQueue),
-                                 r]() { self->DownstreamWork(data, r); });
+        r->QueueWork([self = shared_from_this(), data = std::move(m_DownstreamQueue), r]() {
+          self->DownstreamWork(data, r);
```
jagerman, line 264: Add a `mutable` to the lambda and then `std::move(data)` here to avoid the atomic ref increment

llarp/router/abstractrouter.hpp
```diff
@@ -239,7 +261,7 @@ namespace llarp
 
     /// set router's service node whitelist
     virtual void
-    SetRouterWhitelist(const std::vector<RouterID>& routers) = 0;
+    SetRouterWhitelist(const std::vector<RouterID> routers) = 0;
```
jagerman, line 264: Why this change to copy?

llarp/router/abstractrouter.hpp
```diff
@@ -268,14 +290,10 @@ namespace llarp
...
-    NotifyRouterEvent(Params&&... args) const
+    NotifyRouterEvent([[maybe_unused]] Params&&... args) const
     {
-      // TODO: no-op when appropriate
-      auto event = std::make_unique<EventType>(args...);
 #ifdef LOKINET_HIVE
-      hive->NotifyEvent(std::move(event));
-#elif LOKINET_DEBUG
-      LogDebug(event->ToString());
+      hive->NotifyEvent(std::make_unique<EventType>(args...));
```
jagerman, line 296: Perfect forwarding of `args...`:
```C++
hive->NotifyEvent(std::make_unique<EventType>(std::forward<Params>(args)...));
```

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -0,0 +1,202 @@
...
+        }
+      }
+
+      if (nodeList.empty())
+      {
+        LogWarn("got empty service node list from lokid");
+        return;
+      }
+      // inform router about the new list
+      LogicCall(m_Router->logic(), [r = m_Router, nodeList]() { r->SetRouterWhitelist(nodeList); });
```
jagerman, line 158: Can move to avoid at least one copy (and maybe two) here:
```C++
[r=m_Router, nodeList=std::move(nodeList)]() mutable { r->SetRouterWhitelist(std::move(nodeList)); });
```

llarp/rpc/rpc_server.cpp
```diff
@@ -0,0 +1,187 @@
...
+              std::promise<util::StatusObject> result;
+              LogicCall(m_Router->logic(), [&result, r = m_Router]() {
+                const auto state = r->ExtractStatus();
+                result.set_value(state);
+              });
+              auto ftr = result.get_future();
+              msg.send_reply(CreateJSONResponse(ftr.get()));
+            })
+        .add_request_command("exit", [&](lokimq::Message& msg) {
+          HandleJSONRequest(msg, [r = m_Router](nlohmann::json obj, ReplyFunction_t reply) {
```
jagerman, line 113: This is a pretty big lambda; maybe move this code into a method?

llarp/service/auth.cpp
```diff
@@ -0,0 +1,20 @@
...
+{
+  /// maybe get auth result from string
+  std::optional<AuthResult>
+  ParseAuthResult(std::string data)
+  {
+    static thread_local std::unordered_map<std::string, AuthResult> values = {
+        {"OKAY", AuthResult::eAuthAccepted},
+        {"REJECT", AuthResult::eAuthRejected},
+        {"PAYME", AuthResult::eAuthPaymentRequired},
+        {"LIMITED", AuthResult::eAuthRateLimit}};
```
jagerman, line 14: Why is this thread_local?  It looks like it could simply be a static initialized global const variable.

llarp/service/endpoint.cpp
```diff
@@ -853,6 +856,39 @@ namespace llarp
...
+    void
+    Endpoint::SendAuthReject(
+        path::Path_ptr path, PathID_t replyPath, ConvoTag tag, AuthResult result)
+    {
+      if (result == AuthResult::eAuthAccepted)
+        return;
+      ProtocolFrame f;
+      f.R = result;
+      f.T = tag;
+      f.F = path->intro.pathID;
```
jagerman, line 882: w t f s l v n

jagerman: (what the fuck single letter variable names)

llarp/util/encode.hpp
```diff
@@ -150,6 +150,20 @@ namespace llarp
...
+    while (begin != end && sz)
+    {
+      const auto first = *begin++;
+      const auto second = *begin++;
+      *(target++) = char2int(first) * 16 + char2int(second);
+      --sz;
+    }
+    return sz == 0;
+  }
+
```
jagerman, line 166: The vast majority of this file can be replaced with lokimq's hex.h/base32z.h/base64.h (which use constexpr lookup tables and probably end up faster -- especially for base32z decoding which is currently using an unordered_map here).

jagerman: The size-limited one here isn't directly replaceable, but can be wrapped pretty easily:
```C++
template <typename InputIt, typename OutputIt>
HexDecode(InputIt begin, InputIt end, OutputIt target, size_t sz)
{
  if (std::distance(begin, end) != 2*sz) return false;
  lokimq::to_hex(begin, end, target);
  return true;
}
```

readme.md
```diff
@@ -57,61 +57,44 @@ to run as relay:
...
 build:
 
     $ sudo apt install build-essential cmake git libcap-dev curl libuv1-dev libsodium-dev libcurl4-openssl-dev pkg-config
-    $ git clone https://github.com/loki-project/loki-network
+    $ git clone --recursive https://github.com/loki-project/loki-network
     $ cd loki-network
+    $ mkdir build 
+    $ cd build
+    $ cmake .. 
     $ make
```
jagerman, line 80: `make -j$(nproc)`

readme.md
```diff
@@ -57,61 +57,44 @@ to run as relay:
...
     
-    $ git clone https://github.com/loki-project/loki-network
+    $ git clone --recursive https://github.com/loki-project/loki-network
     $ cd loki-network
-    $ make -j8
-
+    $ mkdir build 
+    $ cd build
+    $ cmake .. 
+    $ make
```
jagerman, line 96: I'd suggest `make -j$(nproc)` here but apparently mac doesn't ship with it.  You probably have to ask Siri or something how many cores you have.

llarp/util/encode.hpp
jagerman: Done (via sub-PR into this PR).

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -0,0 +1,200 @@
...
+    {
+      // m_lokiMQ->log_level(toLokiMQLogLevel(LogLevel::Instance().curLevel));
+    }
+
+    void
+    LokidRpcClient::ConnectAsync(lokimq::address url)
+    {
+      LogInfo("connecting to lokid via LMQ at ", url);
+      m_lokiMQ->connect_remote(
+          url.zmq_address(),
```
jagerman, line 46: If you pull the latest lokimq master I fixed this API oversight by letting you pass in the `address` here.  (Which also potentially carries more info than just the zmq address -- for example, someone can use an address that specified curve encryption + server pubkey).

llarp/net/ip_packet.cpp
notlesh: :+1:

llarp/router/abstractrouter.hpp
majestrate: lifetime concerns

llarp/ev/ev_libuv.cpp
```diff
@@ -1052,7 +1051,27 @@ namespace libuv
...
+
+    while (m_LogicCalls.full() and inEventLoop)
+    {
+      FlushLogic();
+    }
+    if (inEventLoop)
+    {
+      if (m_LogicCalls.tryPushBack(f) != llarp::thread::QueueReturn::Success)
+      {
+        f();
```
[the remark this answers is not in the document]
notlesh: Jeff addressed this by log warning, for the record

PR #1307, opened 2020-06-11
Implement upstream DNS using libunbound

tewinget:


llarp/dns/unbound_resolver.cpp
```diff
@@ -0,0 +1,142 @@
...
+    llarp_buffer_t buf;
+    buf.base = buf.cur = static_cast<byte_t*>(result->answer_packet);
+    buf.sz = result->answer_len;
+
+    MessageHeader hdr;
+    hdr.Decode(&buf);
+    hdr.id = lookup->msg.hdr_id;
+
+    Message msg(hdr);
+    msg.Decode(&buf);
```
majestrate, line 73: we don't implement all of the parsing of RR so it's probably better to only rewrite the message id and send the body back unmodified.

PR #1311, opened 2020-06-17
update ci to use zip for windows

majestrate:


majestrate:
looks like it works for real now

jagerman:
Pushed a commit to DRY the upload script by merging the windows-specific bits into the static upload script.

PR #1283, opened 2020-05-20
Peer stats 2020 05 20

notlesh:
This PR adds a framework for collecting peer stats for use in analyzing performance of Service Node peers.

PR #1312, opened 2020-06-24
Router hive router access

notlesh:
This builds on #1283. It contains some mods that are probably useful to RouterHive, but also has some code that should be handled with caution, specifically 95b49ed which adds a function call to `Router` to disable gossiping of RCs (for testing purposes).

That might not be so bad itself, but in general, the concept of adding code that causes routers to misbehave for testing purposes should be done in such a way that it can't make its way into a production router.


notlesh:
This PR can be closed, the work done on `HiveContext` and `HiveRouter` as well as ripping out the C-API make it irrelevant.

PR #1314, opened 2020-06-29
Remove [most of] C API

notlesh:
This removes most of the C API, which was acting as a mostly redundant.

Hive is broken on this branch, but can be addressed in #1312.

### DIFF
--- a/include/llarp.hpp
+++ b/include/llarp.hpp
@@ -60,7 +60,7 @@ namespace llarp
     LoadDatabase();
 
     void
-    Setup(bool isRelay);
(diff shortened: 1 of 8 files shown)


include/llarp.hpp
```diff
@@ -54,20 +53,17 @@ namespace llarp
...
-
     void
     Close();
 
     int
     LoadDatabase();
 
-    int
+    void
     Setup(bool isRelay);
```
majestrate, line 63: this should take a `RuntimeOptions` instead

PR #1318, opened 2020-07-24
Peer stats follow up 2020 07 09

notlesh:
This builds upon the previous peer stats work to integrate with `lokid`. This adds a bt-encoded API endpoint for loki-core to hit and serializes data with `loki-mq`'s bt-encoding framework.

At the time of this writing, the only outstanding work (other than lots of testing) is to parse `lokid`'s list of RouterIds.

The corresponding `loki-core` PR is: https://github.com/loki-project/loki-core/pull/1198


llarp/router/router.cpp
```diff
@@ -222,7 +222,31 @@ namespace llarp
...
+              numTries,
+              " of ",
+              maxTries,
+              " to get lokid identity keys because: ",
+              e.what());
+
+          if (numTries == maxTries)
+            throw;
+        }
+      }
```
notlesh, line 249: This deserves a closer look. The problem is that `lokinet` will try to immediately connect to `lokid` if configured to do so, and bails out ungracefully if it can't connect.

PR #1319, opened 2020-07-29
exit tool rewrite in C++ (without macos support yet)

majestrate:
rewrite the lokinet-vpn tool in native code so we dont need to ship python on windows/macos.
windows/macos support comes later.
clean up unit tests and correct some incorrectness with bencoding of peer stats.


daemon/lokinet-vpn.cpp
```diff
@@ -0,0 +1,419 @@
...
+#include <array>
+
+#ifdef _WIN32
+// add the unholy windows headers for iphlpapi
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#include <iphlpapi.h>
+#include <stdio.h>
+
+#pragma comment(lib, "iphlpapi.lib")
```
despair86, line 15: who uses MSVC

majestrate: i copy pasted this from the windows documentation

majestrate:
rebased on dev

PR #1327, opened 2020-08-12
Win32 scm daemon

despair86:


despair86:
@majestrate look how cursed it is!

majestrate:
run `contrib/format.sh` please

despair86:
have clang-format 10 send help

majestrate:
does clang-format 10 not work right?

despair86:
coming soon: an update to the daemon startup parameters to keep keys, RCs, and profiles in `$SystemRoot/lokinet` same way the systemd unit starts up in `/var/run/lokinet`

also a cleanup of `daemon/main.cpp` to get rid of 90% of the platform-specific fuzz

majestrate:
looks like clang format 10 does indeed have different rules yet again...

PR #1330, opened 2020-08-31
proper route poking

majestrate:
* update finding default gateway to read from procfs instead of calling popen
* add dynamic route adding/removal for first hops


llarp/net/route.cpp
```diff
@@ -338,28 +340,28 @@ namespace llarp::net
...
+        if (ch != '0')
+          return false;
       }
-      if (words[0] == "default" and words[1] == "via" and words[3] == "dev" and words[4] != ifname)
+      return true;
+    };
+    std::ifstream inf("/proc/net/route");
+    for (std::string line; std::getline(inf, line);)
+    {
+      const auto parts = split(std::string_view{line}, '\t');
```
jagerman, line 355: The explicit construction here isn't needed, `std::string`s are always implicitly convertible to `std::string_view`s (which is part of what makes string_view arguments nice).

llarp/net/route.cpp
```diff
@@ -338,28 +340,28 @@ namespace llarp::net
...
+      if (is_zero(parts[1]) and parts[0] != ifname)
       {
-        gateways.emplace_back(std::move(words[2]));
+        nuint32_t x{};
+        const auto& ip = parts[2];
+        std::from_chars(ip.begin(), ip.end(), x.n, 16);
+        gateways.emplace_back(x.ToString());
```
jagerman, line 361: This is weird and it was not at all obvious to me why it works.

The hex value (e.g. `0100000A` for 10.0.0.1) is the little endian byte representation (*not* a hex-encoded 32-byte value) but `from_chars` always interprets values as a hex-encoded values (which are implicitly big-endian), so it ends up doing the byte swap -- but doing a byte swap implicitly inside a text-to-string converter is *really* unintuitive.  I think this would be clearer:
```C++
if (ip.size() == 2*sizeof(uint32_t) && lokimq::is_hex(ip))
{
  huint32_t x;
  lokimq::from_hex(ip.begin(), ip.end(), reinterpret_cast<char*>(&x.h));
  gateways.emplace_back(x.ToString());
}
```

llarp/net/route.cpp
```diff
@@ -338,28 +340,31 @@ namespace llarp::net
...
+          return false;
       }
-      if (words[0] == "default" and words[1] == "via" and words[3] == "dev" and words[4] != ifname)
+      return true;
+    };
+    std::ifstream inf("/proc/net/route");
+    for (std::string line; std::getline(inf, line);)
+    {
+      const auto parts = split(line, '\t');
+      if (is_zero(parts[1]) and parts[0] != ifname)
```
jagerman, line 356: The lambda could be ditched and replaced with:
```C++
if (parts[1].find_first_not_of('0') != std::string::npos and ...
```

jagerman: Oops, should be `==`

llarp/net/route.cpp
majestrate: done

llarp/net/route.cpp
majestrate: done

llarp/net/route.cpp
majestrate: done

tewinget: waiting for push before approval

majestrate: pushed

PR #1333, opened 2020-09-02
macos route poking

majestrate:
this makes route poking work on macos

### DIFF
--- a/llarp/net/route.cpp
+++ b/llarp/net/route.cpp
@@ -412,8 +412,8 @@ namespace llarp::net
     ssize_t read = 0;
     while ((read = getline(&line, &len, p)) != -1)
     {
-      std::string line_str(line, len);
-      const auto parts = llarp::split(line_str, sv " ");
+      std::string_view line_str(line, len);
+      const auto parts = llarp::split(line_str, " "sv);
       std::vector<std::string_view> parts_nonempty;
       for (auto part : parts)
       {
--- a/llarp/net/route.cpp
+++ b/llarp/net/route.cpp
@@ -243,7 +243,7 @@ namespace llarp::net
 #if _WIN32
     ss << "route ADD " << ip << " MASK 255.255.255.255 " << gateway << " METRIC 2";
 #elif __APPLE__
-    ss << "route -n add -host " << ip << " " << gateway;
+    ss << "/sbin/route -n add -host " << ip << " " << gateway;
 #else
 #error unsupported platform
 #endif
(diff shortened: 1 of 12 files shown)


llarp/net/route.cpp
```diff
@@ -399,35 +399,36 @@ namespace llarp::net
...
+      for (auto part : parts)
       {
-        line_str = line_str.substr(7);
-        while (line_str[0] == ' ')
-        {
-          line_str = line_str.substr(1);
-        }
-        const auto pos = line_str.find(" ");
-        if (pos != std::string::npos)
+        if (not part.empty())
```
tewinget, line 423: It seems that if you use `split_any(input_str, delimiter(s), bool trim)` with trim=true then you will get this behavior without having to check here for empty splits.

majestrate:
The documentation is unclear if that really does yield the
desired result.

llarp/router/router.cpp
```diff
@@ -770,6 +770,47 @@ namespace llarp
...
+        for (const auto& [ip, gw] : m_PokedRoutes)
+        {
+          // save hole
+          holes.emplace(ip);
+          // unpoke route
+          net::DelRoute(ip, gw);
+        }
+        m_PokedRoutes.clear();
+
+        if (not gateway.empty())
```
tewinget, line 803: Seems this is basically "our gateway changed, so check if that means we've just lost internet connectivity entirely".  In that case, will the routes we removed above be recreated properly when connectivity is restored?  Or will new first hops be selected rendering that moot?

llarp/service/endpoint.cpp
```diff
@@ -914,7 +915,24 @@ namespace llarp
         RemoveConvoTag(frame.T);
         return true;
       }
-      return frame.AsyncDecryptAndVerify(EndpointLogic(), p, m_Identity, this);
+      if (not frame.AsyncDecryptAndVerify(EndpointLogic(), p, m_Identity, this))
+      {
```
tewinget, line 919: Is this (and should it be) doing crypto in the logic thread?

llarp/service/endpoint.cpp
```diff
@@ -914,7 +915,24 @@ namespace llarp
...
+        // send reset convo tag message
+        ProtocolFrame f;
+        f.R = 1;
+        f.T = frame.T;
+        f.F = p->intro.pathID;
+
+        f.Sign(m_Identity);
```
tewinget, line 926: Is this (and should it be) doing crypto in the logic thread?

majestrate: not frequent enough to warrant it imo

llarp/router/router.cpp
majestrate: yes they will be recreated later on demand.

llarp/service/endpoint.cpp
tewinget: ok, just checking

llarp/net/route.cpp
majestrate: meh

majestrate: split_any used

tewinget (approved):

PR #1336, opened 2020-09-04
remove locks

majestrate:
remove locks and use lockless queues to remove lock contention.
this yields a massive speedup in mainloop thread. before it was eating 100% cpu at 3200KBps, now it is at about 54% at same speeds.


tewinget (approved):
LGTM

PR #1342, opened 2020-09-09
LNS

majestrate:
relay and client side LNS.


llarp/service/tag_lookup_job.hpp
```diff
@@ -1,75 +1,4 @@
 #ifndef LLARP_SERVICE_TAG_LOOKUP_JOB_HPP
 #define LLARP_SERVICE_TAG_LOOKUP_JOB_HPP
```
tewinget, line 2: delete it harder!

llarp/service/endpoint_types.hpp
```diff
@@ -54,6 +54,8 @@ namespace llarp
 
     using PathEnsureHook = std::function<void(Address, OutboundContext*)>;
 
+    using LNSNameCache = std::unordered_map<std::string, std::pair<Address, llarp_time_t>>;
```
tewinget, line 57: Should probably either make decayed hashset a templated type with an arbitrary container, or make a decayed map class.  Likely other things will come along (or are already implemented) that would benefit from such a class.

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -215,6 +215,36 @@ namespace llarp
...
+              try
+              {
+                const auto j = nlohmann::json::parse(data[1]);
+                const auto itr = j.find("encrypted_value");
+                if (itr != j.end())
+                {
+                  result = lokimq::from_hex(itr->get<std::string>());
+                }
+              }
+              catch (...)
```
tewinget, line 239: Should have a comment with a `catch (...) do nothing` block, to save others and future-you from figuring out what you're ignoring and why.

llarp/handlers/tun.cpp
```diff
@@ -468,6 +469,21 @@ namespace llarp
                 addr.as_array(), std::make_shared<dns::Message>(msg), isV6);
           }
         }
+        else if (ends_with(qname, ".loki"))
+        {
+          return LookupNameAsync(
```
tewinget, line 474: maybe length check here and NX if the name is too long?

jagerman: Validity rules for registration:
- not `localhost.loki`, `loki.loki`, or `snode.loki`
- if it contains no dash then max 32 characters long (assuming a leading subdomain has been stripped)
These are from general DNS requirements, and also enforced in registrations:
- Must be all [A-Za-z0-9-].  (A-Z will be lower-cased by the RPC call).
- cannot start or end with a `-`
- max 63 characters long if it does contain a dash
- cannot contain `--` in the third and fourth characters *unless* it starts with `xn--`

llarp/dht/messages/findname.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+    if (key == "T")
+    {
+      return bencode_read_integer(val, &TxID);
+    }
+    return bencode_discard(val);
+  }
+
+  bool
+  FindNameMessage::HandleMessage(struct llarp_dht_context* dht, std::vector<Ptr_t>& replies) const
```
tewinget, line 42: a bit of whitespace between sections in here would help with readability imo

llarp/dht/messages/gotname.cpp
```diff
@@ -0,0 +1,49 @@
+#include <dht/messages/gotname.hpp>
+#include <lokimq/bt_serialize.h>
+
+namespace llarp::dht
+{
+  constexpr size_t NameSizeLimit = 512;
```
tewinget, line 6: as per conversation about this, seems checking against this limit doesn't make sense here.

llarp/service/tag_lookup_job.hpp
majestrate: done

llarp/handlers/tun.cpp
majestrate: added this check

llarp/service/endpoint_types.hpp
majestrate: refactored

llarp/dht/messages/gotname.cpp
```diff
@@ -0,0 +1,55 @@
...
+      : IMessage(from), result(std::move(data)), TxID(txid)
+  {
+    if (result.ciphertext.size() > NameSizeLimit)
+      throw std::invalid_argument("name data too big");
+  }
+
+  bool
+  GotNameMessage::BEncode(llarp_buffer_t* buf) const
+  {
+    const std::string nonce((const char*)result.nonce.data(), result.nonce.size());
```
tewinget, line 18: Encoding this as a string rather than an integer won't run into endian-ness issues?

tewinget: Nevermind, I'm dumb.

tewinget (approved):
LGTM

PR #1343, opened 2020-09-09
Update requirements and other README.md clean up

neuroscr:
git is now required due to submodules
libcurl/libcap: not sure if the minimum version is correct, The number I used is based on the bionic package version
add recursive option to all cloning
other tweaks to make concise and  precise

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -18,20 +18,25 @@ See the [documentation](https://loki-project.github.io/loki-docs/Lokinet/Lokinet
 
 Also read the [Public Testing Guide](https://lokidocs.com/Lokinet/Guides/PublicTestingGuide/#1-lokinet-installation) for installation and other helpful information.
 
+### Create default config
+
+to configure as client:
+
+    $ lokinet -g
+    $ lokinet-bootstrap
+
+to configure as relay:
+
+    $ lokinet -r -g
+    $ lokinet-bootstrap
+
+
 ## Running on Linux
 
 **DO NOT RUN AS ROOT**, run as normal user. This requires the binary to have the proper setcaps set by `make install` on the binary.
 
-to run as client:
+to run, after you create default config:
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -77,7 +80,7 @@ build:
     $ mkdir build 
     $ cd build
     $ cmake .. 
-    $ make -j$(nproc)
+    $ make -j
```
jagerman, line 83: This is basically a fork bomb

readme.md
```diff
@@ -61,14 +57,21 @@ to run as relay:
 
 Build requirements:
 
+* Git
 * CMake
 * C++ 17 capable C++ compiler
-* gcovr (if generating test coverage with gcc)
 * libuv >= 1.27.0
 * libsodium >= 1.0.18
+* libcurl >= 7.58.0
```
majestrate, line 65: curl is no longer used in the codebase.

readme.md
```diff
@@ -61,14 +57,21 @@ to run as relay:
...
+* gcovr (if generating test coverage with gcc)
 
 ### Linux
 
+additional build requirements:
+
+* libcap >= 2.25 (Linux only)
```
majestrate, line 72: this is not a build dependency. setcap is only for source based installs and it's provided by `libcap2-bin` on debian.

readme.md
neuroscr: How so?

jagerman: `make -j` without a number spawns as many compilation processes as it can.

readme.md
```diff
@@ -102,6 +102,7 @@ windows builds are cross compiled from ubuntu linux
 
 additional build requirements:
 
+* libcurl
```
majestrate, line 105: but why?

majestrate (approved):
looks good now.

PR #1345, opened 2020-09-15
Win32 exclude tun from route poking

majestrate:
* exclude tun interfaces from route poking on windows
* use explicit path for route command on windows
* add IGNORE command for verify format script for use as pre-commit hook with `git add -p`


despair86 (approved):
looks kewl

majestrate:
this broke, i need to fix it.

majestrate:
re opening with different branch name

PR #1352, opened 2020-09-19
dont poke routes if we have no exit

majestrate:
if we have no exit enabled disable the route poking.

### DIFF
--- a/llarp/rpc/rpc_server.cpp
+++ b/llarp/rpc/rpc_server.cpp
@@ -167,7 +167,7 @@ namespace llarp::rpc
                 {
                   endpoint = endpoint_itr->get<std::string>();
                 }
-                LogicCall(r->logic(), [map, exit, range, token, endpoint, r, reply]() {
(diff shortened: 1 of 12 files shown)


majestrate:
rebased

llarp/rpc/rpc_server.cpp
```diff
@@ -1,5 +1,5 @@
 #include "rpc_server.hpp"
-#include <router/abstractrouter.hpp>
+#include <router/router.hpp>
```
tewinget, line 2: If things need to be added to a class' interface, that should be done via the interface class.  If that feels cumbersome, probably a sign that kitchen sinking is happening.

majestrate:
squashed and redid commits so they are cleaner.

llarp/rpc/rpc_server.cpp
majestrate: done, squashed and rebased

tewinget (approved):
Looks fine.

Issue #840, opened 2019-09-21
Failure to save self.signed (and related files) should be fatal


jagerman:
I had a snapp running but the configured directory wasn't owned by the lokinet user so I got:

Sep 21 02:23:15 dev lokinet-snapps[1758261]: [ERR] [lokinet](515) Sat Sep 21 05:23:15 2019 UTC [+10 ms] util/fs.cpp:119	failed to ensure /etc/loki/snapp-dev.private, Permission denied
Sep 21 02:23:15 dev lokinet-snapps[1758261]: [ERR] [lokinet](515) Sat Sep 21 05:23:15 2019 UTC [+10 ms] service/endpoint.cpp:455	Can't ensure keyfile [/etc/loki/snapp-dev.private]

but lokinet started up anyway and I didn't notice the error.  I was using the snapp address (this was the block explorer snapp -- published in some of our docs), but on a lokinet restart because it hadn't saved the file, I got a totally new address and the old one was permanently lost.

That should be a fatal error instead, at the very least for snapp addresses (which are meant to be preserved).

majestrate:
sounds like a place where exceptions would be appropriate.

majestrate:
this should be fixed.

PR #1358, opened 2020-09-22
make errors with service keyfile throw

majestrate:
addresses #840


tewinget (approved):
Looks like it should work.

PR #1360, opened 2020-09-22
Remove gcc-5 workaround

jagerman:


jagerman:
Seems good.  The build failure is some unrelated windows wchar path crap.

majestrate (approved):

PR #1368, opened 2020-09-25
New RC format

tewinget:
Implements the new RouterContact format.  This PR does not cause the code to create the new RC format, but it is able to read and write it.

### DIFF
--- a/llarp/router_contact.cpp
+++ b/llarp/router_contact.cpp
@@ -100,9 +100,6 @@ namespace llarp
       if (not buf->writef("e"))
         return false;
 
-      buf->sz = buf->cur - buf->base;
-      buf->cur = buf->base;
-
       return true;
     }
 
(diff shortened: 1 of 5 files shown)


llarp/router_contact.cpp
```diff
@@ -189,6 +222,70 @@ namespace llarp
     return obj;
   }
 
+  bool
+  RouterContact::BDecode(llarp_buffer_t* buf)
+  {
+    Clear();
+
+    if ((*buf)[0] == 'd') // old format
```
majestrate, line 230: this should be `*buf->cur` instead of `(*buf)[0]`

melyalex (approved):

melyalex (approved):

melyalex (approved):

llarp/router_contact.cpp
```diff
@@ -189,6 +219,81 @@ namespace llarp
...
+  RouterContact::DecodeVersion_0(llarp_buffer_t* buf)
+  {
+    signed_bt_dict = std::string(reinterpret_cast<char *>(buf->cur), buf->sz);
+    return bencode_decode_dict(*this, buf);
+  }
+
+  bool
+  RouterContact::DecodeVersion_1(lokimq::bt_list_consumer& btlist)
+  {
+    auto signature_string = btlist.consume_string_view();
```
majestrate, line 277: `const auto`

majestrate (approved):

PR #1369, opened 2020-09-25
serialize auth codes in rpc

majestrate:
serialize auth codes for endpoints in rpc so the gui can see them

### DIFF
--- a/llarp/router_contact.hpp
+++ b/llarp/router_contact.hpp
@@ -19,7 +19,7 @@
 
 namespace lokimq
 {
-  struct bt_list_consumer;
+  class bt_list_consumer;
 }  // namespace lokimq
 
 namespace llarp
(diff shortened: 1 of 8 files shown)


majestrate:
this needs to go in before we tag

llarp/router_contact.hpp
```diff
@@ -19,7 +19,7 @@
 
 namespace lokimq
 {
-  struct bt_list_consumer;
+  class bt_list_consumer;
```
tewinget, line 22: for some reason I conflated this with the list producer I was working on when you mentioned it, so I was confused.  ty for fixing.

tewinget (approved):
looks fine

tewinget:
Linter is unhappy, should clang-format before merge.

majestrate:
rebased

llarp/router_contact.hpp
majestrate: yw

PR #1370, opened 2020-09-27
dont serialize ephemeral members of service identity

majestrate:
* don't serialize ephemeral members of service identity so that generating keyfiles is easier
* generate ephemeral members of service identity
* add a keygen python script that generates hidden service keyfiles

### DIFF
--- a/contrib/py/keygen/keygen.py
+++ b/contrib/py/keygen/keygen.py
@@ -6,7 +6,7 @@ keygen tool for lokinet
 from argparse import ArgumentParser as AP
 from base64 import b32encode
 
-import pysodium
+from nacl.signing import SigningKey
 
 def base32z(data):
     """ base32 z encode """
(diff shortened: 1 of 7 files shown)


contrib/py/keygen/keygen.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+"""
+keygen tool for lokinet
+"""
+
+from argparse import ArgumentParser as AP
+from base64 import b32encode
+
+import pysodium
```
jagerman, line 9: I think that the pynacl module would be a better choice here (despite the name, it is a libsodium wrapper) as it is far more commonly used than pysodium.

majestrate: it's easier with pysodium. pynacl's interface is too high level for what i need, it gets waaaay to verbose.

tewinget (approved):
LGTM

PR #1372, opened 2020-09-30
order packets when writing to network interface

majestrate:
this may or may not help with tcp throughput.


tewinget (approved):
Looks fine, but save for after 0.8 for further testing.

PR #1395, opened 2020-10-05
respect configured worker threads count

majestrate:
set lmq thread count to the configured amount


tewinget (approved):
Yeah, that makes sense.

PR #1399, opened 2020-10-08
add route poking for solaris and bsd

despair86:


despair86:
performance is disappointing cause i literally cannot run the optimiser on libzmq or else lokinet blows up in my face (waiting on clang-11 and g++ 11 lul)

llarp/net/route.cpp
```diff
@@ -449,14 +461,18 @@ namespace llarp::net
...
     LogDebug("get gateways not on ", ifname);
     const auto maybe = GetIFAddr(ifname);
     if (not maybe.has_value())
       return gateways;
     const auto interface = maybe->toString();
     // mac os is so godawful man
+#if __APPLE__
     FILE* p = popen("/usr/sbin/netstat -rn -f inet", "r");
+#else
+    FILE* p = popen("/usr/bin/netstat -rn -f inet", "r");
```
despair86, line 491: being a mashup of BSD and SysV works well here: macOS and Solaris have the same flavour of `netstat(1M)`

llarp/net/route.cpp
```diff
@@ -408,6 +416,10 @@ namespace llarp::net
     Execute("/sbin/route -n delete -cloning -net 0.0.0.0 -netmask 128.0.0.0 -interface " + ifname);
     Execute(
         "/sbin/route -n delete -cloning -net 128.0.0.0 -netmask 128.0.0.0 -interface " + ifname);
+#elif __sun
+    const auto iface = GetIFAddr(ifname);
+    Execute("/usr/sbin/route delete 0.0.0.0/1 -iface " + iface->toHost());
```
despair86, line 436: i could have also recycled the route commands but `-cloning` does nothing on SysV

majestrate: this returns a std::optional and de-referencing a std::optional is undefined behavior

despair86: then what the hell is line 378
```
 read_addr(maybe->toHost().c_str(), &gw_addr);
```

despair86:
this is ready now since the SIIT fixes were stuffed into #1421

despair86:
route poker runs twice when exit is enabled: lokimq via lokinet-vpn triggers the original route poker which works correctly
then the new route poker is triggered and fails with error `[route|file] exists` which is harmless

new route poker leaks routes on BSD and possibly Macintosh (they use the same command to add and delet individual routes). removing the /1 routes works as expected however.

output of old route poker at exit:
```
delete route 223.223.223.223 via 223.223.223.223
```
behaviour of new route poker is inconsistent on BSD, and mostly steps out of the way on Solaris oddly enough (routes are not leaked on exit, at least)

last but not least I still haven't figured out how to get udp sockets from timing out so quickly and causing tunnel layer tcp timeouts (bsd only)

majestrate:
i would really rather use a routing socket at this point.

despair86:
i dropped clang-9 a long time ago so all lints fail for now

despair86:
finally got rid of windows hot loop allocation :DDD

llarp/ev/ev_win32.cpp
```diff
@@ -12,6 +12,12 @@ static HANDLE tun_event_queue = INVALID_HANDLE_VALUE;
 static HANDLE* kThreadPool;
 static int poolSize;
 
+// ring buffer to replace dynamic loop allocation
+// we only care about the first dimension of the array
+static byte_t pkt_ring[1000][1500];
```
majestrate: can we put this into another PR?

majestrate:
we are gutting this code soon so i think i'll be closing this.

PR #1400, opened 2020-10-09
fix empty config case

majestrate:


tewinget (approved):

llarp/config/config.cpp
```diff
@@ -480,16 +480,7 @@ namespace llarp
...
-          if (arg.empty())
-          {
-            const auto maybe = llarp::FindFreeTun();
-            if (not maybe)
-              throw std::invalid_argument("cannot determine free interface name");
-            arg = *maybe;
-          }
-          m_ifname = arg;
-        });
+        [this](std::string arg) { m_ifname = arg; });
```
jagerman: `AssignmentAcceptor(m_ifname)`

Issue #1402, opened 2020-10-10
cmake can't find libuv files


jewel-snake:
Got an error while compiling lokinet v8.0.0
`Make Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LIBUV_LIBRARY (ADVANCED)
    linked by target "lokinet-platform" in directory /home/kun/Downloads/loki-network/llarp`
despite that I have libuv installed on my PC. Got this issue on three different machines and two different OS(Arch and Void).

majestrate:
On Sat, 2020-10-10 at 09:22 -0700, jewel-snake wrote:
> Got an error while compiling lokinet v8.0.0
> Make Error: The following variables are used in this project, but
> they are set to NOTFOUND. Please set them or make sure they are set
> and tested correctly in the CMake files: LIBUV_LIBRARY (ADVANCED)
> linked by target "lokinet-platform" in directory
> /home/kun/Downloads/loki-network/llarp
> despite that I have libuv installed on my PC. Got this issue on three
> different machines and two different OS(Arch and Void).
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.

on debian the package is libuv1-dev, you may also need pkgconfig.

jewel-snake:
Reffering to [this](https://bbs.archlinux.org/viewtopic.php?id=41231) arch linux forum thread I assume that  I have that package installed.

majestrate:
On Sat, 2020-10-10 at 11:22 -0700, jewel-snake wrote:
> Reffering to this arch linux forum thread I assume that I have that
> package installed.
> 
> —
> 
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or unsubscribe.

what version of libuv do you have install?
do you have pkgconfig as well?

jewel-snake:
>libuv
1.40.0-1
>pkgconfig
1.7.3-1

jagerman:
What do you get if you run:
```
pkg-config --libs --cflags libuv
```
?

majestrate:
On Sat, 2020-10-10 at 11:45 -0700, jewel-snake wrote:
> > libuv
> > 1.40.0-1
> > pkgconfig
> > 1.7.3-1
> > 
> 
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or unsubscribe.

[...]

PR #1403, opened 2020-10-10
handle libuv case when we don't have it

majestrate:


jagerman:
This seems fine, however can we fix the detection issue from #1402?  The FindLibuv cmake code looks a little hairy, and I'm not really sure why it needs to be as complicated as it is (as opposed to being a much simpler pkg-config call).

majestrate:
> This seems fine, however can we fix the detection issue from #1402? 

yes we should

> The FindLibuv cmake code looks a little hairy, and I'm not really sure why it needs to be as complicated as it is (as opposed to being a much simpler pkg-config call).

i agree, it should be binned and redone. I just don't want to do it on the weekend. I'll leave this open and finish it on monday as it only affects builds on archlinux.

tewinget (approved):
sufficient for now, but 1402 will still need a proper solution.

PR #1404, opened 2020-10-12
dns features

majestrate:
* fixes for LNS
* add txt records for status

### DIFF
--- a/llarp/dns/question.cpp
+++ b/llarp/dns/question.cpp
@@ -64,6 +64,12 @@ namespace llarp
       return (qname == "localhost.loki." or llarp::ends_with(qname, ".localhost.loki."));
     }
 
+    bool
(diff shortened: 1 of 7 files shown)


llarp/dns/question.cpp
```diff
@@ -64,6 +64,12 @@ namespace llarp
       return (qname == "localhost.loki." or llarp::ends_with(qname, ".localhost.loki."));
     }
 
+    bool
+    Question::HasSubdomains() const
+    {
+      return qname.size() >= 3;
```
tewinget: I don't understand this.

majestrate: sub.main.loki has 3 parts, sub, main and loki

tewinget (approved):
Looks good.

PR #1407, opened 2020-10-12
Lns exits

majestrate:


llarp/service/endpoint.cpp
```diff
@@ -227,6 +235,30 @@ namespace llarp
           now, m_state->m_RemoteSessions, m_state->m_DeadSessions, Sessions());
       // expire convotags
       EndpointUtil::ExpireConvoSessions(now, Sessions());
+
+      if (NumInStatus(path::ePathEstablished) > 1)
+      {
+        for (const auto& [lnsName, info] : m_StartupLNSMappings)
```
tewinget: since clang doesn't like using this style of for loop's symbols in lambda captures, probably should just iterate the old way and change references below to `x.first` and `x.second` as necessary.

tewinget (approved):
Looks fine, pending CI.

PR #1408, opened 2020-10-13
once the network gets big enough 3s isn't enough

despair86:
to scan the entire nodedb @KeeJef


despair86:
i think the GUI mangles line endings inside lokinet.ini and this causes lokinet to die

despair86:
either that or lokimq

PR #1410, opened 2020-10-13
add conf.d directory for config overrides

majestrate:


tewinget:
Config::LoadDefault function name is a bit ambiguous, made more so by the changes herein.

llarp/config/config.cpp
```diff
@@ -960,13 +960,31 @@ namespace llarp
...
   Config::Override(std::string section, std::string key, std::string value)
   {
-    m_Parser.AddOverride(std::move(section), std::move(key), std::move(value));
+    m_Parser.AddOverride(m_overridesDir / "overrides.ini", section, key, value);
+  }
+
+  void
+  Config::LoadOverrides(fs::path defaultDataDir)
+  {
+    m_overridesDir = defaultDataDir / "conf.d";
```
tewinget: If `m_overridesDir` is only being used in this function, it shouldn't be a member variable.  If it should be a member variable, it should maybe be set elsewhere?

tewinget (approved):
Seems fine, pending CI

PR #1413, opened 2020-10-15
redo windows ci to use debian

majestrate:


tewinget (approved):
Is fine, but would prefer to not have a commit and its revert commit in a PR

aaaaaaand that was fixed literally as I submitted this comment.  ha.

PR #1418, opened 2020-10-21
use correct path for windows service

majestrate:
the gui uses `lokinet` not `.lokinet`, using `.lokinet` on windows was a bug.


despair86:
that was an oversight, see https://github.com/loki-project/loki-network-control-panel/pull/62

tewinget (approved):

PR #1419, opened 2020-10-21
dont bind inbound links as client

majestrate:


majestrate (commented):
cc @tewinget

llarp/ev/ev_libuv.cpp
```diff
@@ -1081,12 +1081,11 @@ namespace libuv
     }
     const auto inEventLoop = *m_EventLoopThreadID == std::this_thread::get_id();
 
-    if (inEventLoop)
+    if (inEventLoop and m_LogicCalls.full())
     {
-      f();
+      FlushLogic();
```
majestrate, line 1086: LogicCall **MUST** defer the call.

tewinget (approved):

PR #1420, opened 2020-10-22
win32 installer updates

majestrate:
* bundle gui
* bundle bootstrap.signed
* install/remove lokinet win32 service


tewinget (approved):
Would prefer to have someone familiar with these things give it a once over, but if it works then it's probably fine.

PR #1421, opened 2020-10-23
enable profiling on clients by default

majestrate:
this may or may not help with path build success

### DIFF
--- a/llarp/dns/server.cpp
+++ b/llarp/dns/server.cpp
@@ -66,11 +66,8 @@ namespace llarp
     Proxy::HandleUDPRecv_server(llarp_udp_io* u, const SockAddr& from, ManagedBuffer buf)
     {
       Buffer_t msgbuf = CopyBuffer(buf.underlying);
-      auto self = static_cast<Proxy*>(u->user)->shared_from_this();
-      // yes we use the server loop here because if the server loop is not the
-      // client loop we'll crash again
-      LogicCall(
-          self->m_ServerLogic, [self, from, msgbuf]() { self->HandlePktServer(from, msgbuf); });
+      auto self = static_cast<Proxy*>(u->user);
+      self->HandlePktServer(from, msgbuf);
     }
 
     void
@@ -140,7 +137,8 @@ namespace llarp
     void
     Proxy::SendServerMessageBufferTo(const SockAddr& to, const llarp_buffer_t& buf)
     {
-      llarp_ev_udp_sendto(&m_Server, to, buf);
+      if (llarp_ev_udp_sendto(&m_Server, to, buf) < 0)
+        llarp::LogError("dns reply failed");
     }
 
     void
@@ -244,7 +242,6 @@ namespace llarp
         return;
       }
 
-      TX tx = {hdr.id, from};
       Message msg(hdr);
       if (!msg.Decode(&pkt))
       {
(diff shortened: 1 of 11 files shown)


llarp/dns/server.cpp
```diff
@@ -96,20 +93,13 @@ namespace llarp
     bool
     Proxy::SetupUnboundResolver(const std::vector<IpAddress>& resolvers)
     {
-      auto failFunc = [self = weak_from_this()](SockAddr to, Message msg) {
-        auto this_ptr = self.lock();
-        if (this_ptr)
-        {
-          this_ptr->SendServerMessageTo(to, std::move(msg));
-        }
+      auto failFunc = [self = shared_from_this()](SockAddr to, Message msg) {
```
tewinget: Any particular reason for this change?

llarp/ev/ev_libuv.cpp
```diff
@@ -474,13 +474,19 @@ namespace libuv
...
+        llarp::LogError("udp sendto failed: ", uv_strerror(ret));
+      }
+      return ret;
     }
 
     bool
     Bind()
     {
-      auto ret = uv_udp_bind(&m_Handle, m_Addr, 0);
+      auto ret = uv_udp_bind(&m_Handle, (sockaddr*)static_cast<const sockaddr_in*>(m_Addr), 0);
```
tewinget: Is...is that a C-cast of a static_cast?  That feels wrong...

llarp/dns/server.cpp
tewinget: Ah, I see the commit message for this now.  I initially used weak_ptr to avoid complexity/leaks when re-initializing upstream resolvers, but if this works it's fine by me.

tewinget:
Seems more or less fine, apart from a test crashing on one of the CI builds.

llarp/ev/ev_libuv.cpp
majestrate: it is very wrong

llarp/net/sock_addr.cpp
```diff
@@ -113,7 +115,13 @@ namespace llarp
     init();
 
     memcpy(&m_addr, &other, sizeof(sockaddr_in6));
-
+    if (other.sin6_addr.s6_addr[11] == 0xff and other.sin6_addr.s6_addr[10] == 0xff)
```
tewinget: for code clarity, probably should comment the significance of checking these two bytes...and of changing the others if the conditional passes.

tewinget: also it looks like this and the other `operator=` are almost identical.  some DRYness would go well here.

llarp/profiling.cpp
```diff
@@ -92,7 +96,8 @@ namespace llarp
   bool
   RouterProfile::IsGoodForPath(uint64_t chances) const
   {
-    return checkIsGood(pathFailCount, pathSuccessCount, chances);
+    return checkIsGood(pathFailCount, pathSuccessCount, chances)
```
tewinget, line 99: any reason `checkIsGood` isn't a member function?  Also, its name is very generic...

majestrate: `checkIsGood` is a constexpr because it's used across multiple different stats member variables.

llarp/net/sock_addr.cpp
majestrate: dry'd out

llarp/ev/ev_libuv.cpp
majestrate: this is better than the old implicit conversions done before, this is more explicit about what is going on.

tewinget (approved):
Approved, pending CI.

PR #1422, opened 2020-10-23
Crank clang version to 11

jagerman:


majestrate:
so it DOES go up to 11....

majestrate (approved):

PR #1423, opened 2020-10-27
more aggressive path building.

majestrate:
do more than one in parallel path builds at a time


llarp/path/pathbuilder.cpp
```diff
@@ -191,7 +191,7 @@ namespace llarp
     {
       util::StatusObject obj{{"buildStats", m_BuildStats.ExtractStatus()},
                              {"numHops", uint64_t(numHops)},
-                             {"numPaths", uint64_t(numPaths)}};
```
majestrate, line 194: no bueno

tewinget (approved):
Should fix the status object at some later time, but ok to merge for now.

Issue #1430, opened 2020-10-28
Specifying relatve lokinet.ini without `./` prefix fails


cathugger:
**Describe the bug**
When starting lokinet as `lokinet lokinet.ini`, it fails with
```
[NFO] [lokinet](763) 2020-10-28 20:33:39.864 GMT [+0.001s] main.cpp:261 Using config file: "lokinet.ini"
[ERR] [lokinet](763) 2020-10-28 20:33:39.890 GMT [+0.027s] config/config.cpp:1054       Error trying to init and parse config from file: Specified [router]:data-dir "" d
oes not exist
[ERR] [lokinet](763) 2020-10-28 20:33:39.890 GMT [+0.027s] main.cpp:289 Fatal: caught exception while running: Config file parsing failed
main thread threw exception: Config file parsing failed
```
Starting as `lokinet ./lokinet.ini` works. If `data-dir "."` in config file gets specified, that works too, since it overrides option which causes failure.

**To Reproduce**
`lokinet lokinet.ini` (with data-dir commented out in config file).

**Screenshots Or Logs**
(specified above)

**Device and Operating system (please complete the following information):**
 - OS: aarch64 gentoo (but shouldn't matter)
-  Device: espressobin, doesn't matter
 - Lokinet Version number or Git commit hash: 0.8.1-rc1 but was problem in 0.8.0 too

majestrate:
fixed by #1432

PR #1432, opened 2020-10-28
allow relative path for configuration file

majestrate:


majestrate:
fixes #1430

llarp/config/config.cpp
```diff
@@ -1132,10 +1132,12 @@ namespace llarp
     if (ec)
       throw std::runtime_error(stringify("filesystem error: ", ec));
 
+    const auto parent = confFile.parent_path();
+
     // create parent dir if it doesn't exist
-    if (not fs::exists(confFile.parent_path(), ec))
+    if ((not parent.empty()) and (not fs::exists(parent, ec)))
```
tewinget: Since we're passing `basedir` in here as `defaultDataDir`, it's weird to use `parent_path` again here.

tewinget: Also the parameter `defaultDataDir` here should probably be renamed; suggestion: `dataDir`.

llarp/config/config.cpp
```diff
@@ -107,6 +107,8 @@ namespace llarp
             "private keys.",
         },
         [this](fs::path arg) {
+          if (arg.empty())
```
tewinget, line 110: imo this should be a fatal error.  The case where the `data-dir` setting is present in the config file but is an empty string is almost certainly user error and should be treated as such.

majestrate: good point

llarp/config/config.cpp
```diff
@@ -1143,9 +1145,13 @@ namespace llarp
     if (ec)
       throw std::runtime_error(stringify("filesystem error: ", ec));
 
-    llarp::LogInfo("Attempting to create config file, asRouter: ", asRouter, " path: ", confFile);
+    llarp::LogInfo(
+        "Attempting to create config file for ",
+        (asRouter ? "router" : "client"),
```
tewinget, line 1142: good call.

llarp/config/config.cpp
```diff
@@ -1132,18 +1134,24 @@ namespace llarp
...
 
+    const auto parent = confFile.parent_path();
+
     // create parent dir if it doesn't exist
-    if (not fs::exists(confFile.parent_path(), ec))
+    if ((not parent.empty()) and (not fs::exists(parent, ec)))
     {
-      if (not fs::create_directory(confFile.parent_path()))
+      if (not fs::create_directory(parent))
         throw std::runtime_error(stringify("Failed to create parent directory for ", confFile));
```
tewinget: this should be something more like "Unable to create config file: unable to create directory \\"directory_path\\" and it does not exist."

tewinget (approved):

PR #1434, opened 2020-10-29
dont do native builds in static linux ci

majestrate:


tewinget (approved):
in future, please leave reasoning in commit message.  If nothing else it helps with wondering "why did we change that?" later using `git blame`.

PR #1438, opened 2020-10-30
make service node dns bind work again

majestrate:


tewinget (approved):
github will let me force push over a PR's branch (if the PR submitter keeps the opt-out box checked) and still I'm a valid approval for merge requirement...lolwut

PR #1440, opened 2020-10-31
add exit.localhost.loki cname record

majestrate:
adds a CNAME record for `exit.localhost.loki` that points to the address of the exit used.

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -521,21 +521,31 @@ namespace llarp
         }
         else if (is_localhost_loki(msg))
         {
-          size_t counter = 0;
-          context->ForEachService(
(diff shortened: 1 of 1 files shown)


llarp/handlers/tun.cpp
```diff
@@ -461,6 +461,26 @@ namespace llarp
...
+            }
+            else
+            {
+              msg.AddNXReply();
+            }
+          }
+          else
+          {
+            msg.AddNXReply();
+          }
```
jagerman, line 475: The two else's can be simplified a bit:

```C++
if (subdomain == "exit" and HasExit())
{
    m_ExitMap.....
}
else
{
  msg.addNXReply();
}

jagerman (approved):

PR #1441, opened 2020-11-01
update information in exit-setup.md

majestrate:
add 1 hops config and increase number of paths and open connections


tewinget (approved):
Looks correct.

PR #1442, opened 2020-11-01
try fixing memory leak in path builder

majestrate:


tewinget (approved):
TODO: replace a shared_ptr with weak_ptr to avoid the ouroboros pointer issue.

Issue #1443, opened 2020-11-01
Clients publishing introsets to expired relays


jagerman:
A couple times in the last day I've been unable to reach blocks.loki.  Lokinet reports, over and over:

```
[+1m03.915s] service/hidden_service_address_lookup.cpp:41	found result for kcpyawm9se7trdbzncimdi5t7st4p5mh9i1mg7gkpuubi4k4ku1y.loki
[+1m04.091s] service/hidden_service_address_lookup.cpp:41	found result for kcpyawm9se7trdbzncimdi5t7st4p5mh9i1mg7gkpuubi4k4ku1y.loki
[+1m04.404s] router/rc_lookup_handler.cpp:97	Lookup tnifuu5joqkthmbmaz8u63kr4twybeajmqya1ptkicbrfdbycbso.snode anonymously
[+1m04.404s] router/rc_lookup_handler.cpp:106	cannot lookup tnifuu5joqkthmbmaz8u63kr4twybeajmqya1ptkicbrfdbycbso.snode anonymously
[+1m04.404s] service/outbound_context.cpp:191	OBContext:endpoint:3xjjgti9sxob1h4oinwfkgdbzq96nwwerf31somfh7su1w316tbo.loki-kcpyawm9se7trdbzncimdi5t7st4p5mh9i1mg7gkpuubi4k4ku1y.loki dropping intro frame, no path to 88aa59cf6983951e2c2bc5cf3f6544d46800a3095b8189362aab02428c20606d
```
then after a while the existing introset expires, my lokinet pulls a new one, and can connect to that new one.  (But later the problem reappeared when another newer introset got published that wanted to align on tnifu again).

This `tnifu` node doesn't appear to be registered; I cannot look it up locally, but I *can* look it up on blocks.loki (and can even ping it), so I suspect this is a service node that expired or got deregistered but blocks.loki and some other lokinet routers along the path aren't properly handling the deactivation of that node.

Separately but perhaps related, once I initiated a lookup locally, my log gets one of these every second:

```
[+20m37.983s] path/pathbuilder.cpp:411	SNode::tnifuu5joqkthmbmaz8u63kr4twybeajmqya1ptkicbrfdbycbso.snode failed to select hop 3
```
and every 30 seconds a timeout message:
```
[+21m32.022s] service/endpoint_util.cpp:65	lookup for tnifuu5joqkthmbmaz8u63kr4twybeajmqya1ptkicbrfdbycbso.snode timed out
[...]

PR #1444, opened 2020-11-01
deny traffic to non whitelisted routers.

majestrate:
deny message delivery from non registered nodes when whitelist is enabled and applicable


majestrate:
fixes #1443

jagerman (approved):

PR #1480, opened 2020-11-16
do /1 route poker split on linux

### LINKED ISSUE #1479: do the /1 split for route poker on linux
i have been getting reports that the default route adding is broke on linux so we're going to have to do that hacky /1 route poking like we do everywhere else.

majestrate:
i hate netlink so god damn much
fixes #1479


tewinget (approved):
Looks fine, pending testing.

PR #1496, opened 2020-11-28
add llarp.endpoint rpc command 

majestrate:
add llarp.endpoint rpc command with a parameter "kill" to kill all active convotags given a list of or a single remote service address.
this is useful for endpoints that use lmq authentication.

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -345,6 +345,25 @@ namespace llarp
       itr->second.lastUsed = Now();
     }
 
+    size_t
+    Endpoint::RemoveAllConvoTagsFor(service::Address remote)
+    {
+      size_t removed = 0;
+      auto& sessions = Sessions();
+      auto itr = sessions.begin();
+      while (itr != sessions.end())
+      {
+        if (itr->second.remote.Addr() == remote)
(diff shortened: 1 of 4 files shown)


llarp/service/endpoint.cpp
```diff
@@ -345,6 +345,25 @@ namespace llarp
       itr->second.lastUsed = Now();
     }
 
+    size_t
+    Endpoint::RemoveAllConvoTagsFor(service::Address remote)
+    {
+      size_t removed = 0;
+      auto& sessions = Sessions();
+      auto itr = sessions.begin();
+      while (itr != sessions.end())
```
tewinget, line 354: This section could use std::find_if, I think?

tewinget: could do `for (auto itr = sessions.find_if(...); itr != sessions.end(); itr = sessions.find_if(...))` I think

tewinget: not sure if that's better/more clear though

tewinget: probably not worth the effort to change, not clearly better and possibly worse.

tewinget (approved):

PR #1511, opened 2021-01-14
libuv cmake fixups

### LINKED ISSUE #1510: libuv libs/headers/etc. shouldn't get installed with a static build
Currently doing a static build ends up installing libuv headers, libs, pkgconfig, etc.  Those shouldn't get installed.

Looking at how it is used, it should be pretty straightforward to build this in `cmake/StaticBuild.cmake` along with all our other static deps rather than as a submodule.

majestrate:
* fix up cmake parts for libuv
* make libuv use external project via StaticBuild.cmake
* remove libuv submodule
this fixes #1510


CMakeLists.txt
```diff
@@ -124,6 +124,19 @@ endif()
...
+
+if(NOT BUILD_STATIC_DEPS)
+  pkg_check_modules(LIBUV libuv>=1.28.0 IMPORTED_TARGET)
+endif()
+if(LIBUV_FOUND AND NOT BUILD_STATIC_DEPS)
+  add_library(libuv INTERFACE)
+  target_link_libraries(libuv INTERFACE PkgConfig::LIBUV)
+else()
+  if(NOT BUILD_STATIC_DEPS)
+    message(FATAL_ERROR "Could not find libu >= 1.28.0; install it on your system or use -DBUILD_STATIC_DEPS=ON")
```
jagerman, line 136: Missing "v" in libuv

jagerman (approved):
Looks good, but needs the libuv version check lowered to be >= 1.18 to be installable on bionic (hopefully that version works, if not we'll have to muck around making a backport).

PR #1512, opened 2021-01-15
Update readme.md

phr0stb1t3:
Added missing instruction for FreeBSD install


majestrate:
this reminded me that i also have to re-add freebsd support for the tun code since we totally redid that part.

majestrate:
otherwise this looks good.

majestrate (approved):

PR #1515, opened 2021-01-19
prevent crash by removing duplicate logic

majestrate:
when a path expires or fails or something causes its lifecycle to end we have cleanup
logic that handles clean up for it. this removes a code path that was crashing that
is duplicated elsewhere and is thus probably safe to bin. yolo.


tewinget (approved):
LGTM, pending testing

PR #1526, opened 2021-01-29
feedback auth result via lmq rpc

### LINKED ISSUE #1497: lokinet exit auth invalidates session before competing auth
**Describe the bug**
lokinet exit with auth resets session before auth can complete

**To Reproduce**
turn on exit with an auth code with lots of traffic on your end

**Screenshots Or Logs**
none

majestrate:
fixes #1497

### DIFF
--- a/llarp/rpc/endpoint_rpc.cpp
+++ b/llarp/rpc/endpoint_rpc.cpp
@@ -33,15 +33,27 @@ namespace llarp::rpc
         });
   }
 
+  bool
+  EndpointAuthRPC::AsyncAuthPending(service::ConvoTag tag) const
+  {
+    return m_PendingAuths.count(tag) > 0;
+  }
+
   void
   EndpointAuthRPC::AuthenticateAsync(
       std::shared_ptr<llarp::service::ProtocolMessage> msg,
       std::function<void(service::AuthResult)> hook)
   {
+    service::ConvoTag tag = msg->tag;
+    m_PendingAuths.insert(tag);
     const auto from = msg->sender.Addr();
-    auto reply = [logic = m_Endpoint->RouterLogic(), hook](service::AuthResult result) {
-      logic->Call([hook, result]() { hook(result); });
-    };
+    auto reply =
+        [self = this, tag, logic = m_Endpoint->RouterLogic(), hook](service::AuthResult result) {
+          logic->Call([self, hook, result, tag]() {
+            self->m_PendingAuths.erase(tag);
+            hook(result);
+          });
+        };
     if (m_AuthWhitelist.count(from))
     {
       // explicitly whitelisted source
--- a/llarp/rpc/endpoint_rpc.cpp
+++ b/llarp/rpc/endpoint_rpc.cpp
@@ -57,20 +57,20 @@ namespace llarp::rpc
     if (m_AuthWhitelist.count(from))
     {
       // explicitly whitelisted source
-      reply(service::AuthResult::eAuthAccepted);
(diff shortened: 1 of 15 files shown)


llarp/rpc/endpoint_rpc.cpp
```diff
@@ -33,15 +33,27 @@ namespace llarp::rpc
...
       std::function<void(service::AuthResult)> hook)
   {
+    service::ConvoTag tag = msg->tag;
+    m_PendingAuths.insert(tag);
     const auto from = msg->sender.Addr();
-    auto reply = [logic = m_Endpoint->RouterLogic(), hook](service::AuthResult result) {
-      logic->Call([hook, result]() { hook(result); });
-    };
+    auto reply =
+        [self = this, tag, logic = m_Endpoint->RouterLogic(), hook](service::AuthResult result) {
```
jagerman, line 51: Why capture `self=this` rather than just `this`?

majestrate: a horrid decrepit bad habit of mine.

llarp/service/auth.cpp
```diff
@@ -18,6 +18,24 @@ namespace llarp::service
...
+      case AuthResult::eAuthRejected:
+        return "rejected";
+      case AuthResult::eAuthPaymentRequired:
+        return "payment required";
+      case AuthResult::eAuthRateLimit:
+        return "rate limit hit";
+      default:
+        return "unknown";
+    }
+  }
```
jagerman: I'm thinking that there needs to be an "other" option here which could be returned with a short explanatory message.  I can imagine a bunch of random things a paid exit might want to return, e.g. "daily bandwidth exceeded" or "account suspended" or ...

jagerman: Or perhaps: AuthResult has Accepted/Rejected, with an optional reason string that could carry the reason info.

majestrate:
rebased on dev

llarp/service/auth.cpp
majestrate: added reason info.

llarp/service/outbound_context.cpp
```diff
@@ -528,7 +528,90 @@ namespace llarp
...
+
+        m_Endpoint->RemoveConvoTag(frame.T);
+        if (authResultListener)
+        {
+          switch (frame.R)
+          {
+            case 1:
+              result.code = eAuthRejected;
+              break;
+            case 3:
```
tewinget: could use a comment noting why `case 2` is (apparently) omitted.

llarp/service/tag.hpp
```diff
@@ -6,10 +6,7 @@
 #include <util/status.hpp>
 
 #include <sodium/crypto_generichash.h>
-
-#ifndef WIN32_LEAN_AND_MEAN
-#define WIN32_LEAN_AND_MEAN
-#endif
+#include <functional>
```
tewinget: only change to this file, so likely unnecessary new include

tewinget (approved):
Couple of small notes, but looks good to me.

llarp/service/outbound_context.cpp
jagerman: Or better: change them to be constexpr values rather than magic numbers.

majestrate: now uses an enum class

llarp/service/tag.hpp
majestrate: done

PR #1529, opened 2021-02-02
NodeDB refactor

majestrate:
* bump zmq static dep
* lokimq -> oxenmq
* llarp_nodedb -> llarp::NodeDB
* remove all crufty api parts of NodeDB
* make NodeDB rc selection api not suck
* make path builder api not suck
* propagate all above changes so that unit tests work and it all compile


llarp/path/pathbuilder.cpp
```diff
@@ -205,56 +205,37 @@ namespace llarp
...
+          },
+          true);
+      return found;
+    }
 
-      do
+    std::optional<std::vector<RouterContact>>
+    Builder::GetHopsForBuild()
+    {
+      if (const auto maybe = m_router->nodedb()->GetIf([r = m_router](const auto& rc) -> bool {
```
tewinget: put the `maybe` assignment outside the `if` please, or remove the `; maybe.has_value()` as it is redundant in this context.

tewinget: also, is `GetIf` the correct function here?  Seems like we want a random/arbitrary RC, but `GetIf` sounds like it should give us an RC we specify *if* it meets the condition we specify.

llarp/router/router.cpp
```diff
@@ -150,19 +150,13 @@ namespace llarp
...
 
-    auto pick_router = [&](auto& collection) -> bool {
-      const auto sz = collection.size();
-      auto itr = collection.begin();
-      if (sz == 0)
-        return false;
-      if (sz > 1)
-        std::advance(itr, randint() % sz);
-      router = itr->first;
+    if (const auto maybe = nodedb()->GetIf([](const auto&) -> bool { return true; }, true);
```
tewinget: If this usage of `GetIf` is correct, perhaps it should be renamed?  Perhaps `GetRandomIf` or similar?

tewinget (approved):
Looks good.  Consider squash before merge.

PR #1532, opened 2021-02-08
thaw network when default gateway changes

majestrate:
thaw does the following when the default gateway changes or we come back from a lost network event.

1) close all link sessions
2) reset all upstream dns contexts

in that order

### DIFF
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -124,7 +124,6 @@ namespace llarp
   void
   Router::Thaw()
   {
-    LogInfo("We arise from a long sleep, probably need to reset the network state");
     // get pubkeys we are connected to
     std::unordered_set<RouterID> peerPubkeys;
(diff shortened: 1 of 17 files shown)


llarp/router/router.cpp
```diff
@@ -121,6 +121,30 @@ namespace llarp
...
+      for (const auto& remote : peerPubkeys)
+        link->CloseSessionTo(remote);
+    });
+    // thaw endpoints
+    hiddenServiceContext().ForEachService([](const auto& name, const auto& ep) -> bool {
+      LogInfo(name, " thawing...");
+      ep->Thaw();
+      return true;
+    });
+    LogInfo("We are ready to go bruh...  probably");
```
tewinget, line 145: Classic jeff log message.  >_>

tewinget (approved):
Looks good

PR #1533, opened 2021-02-08
wire up dns for android

majestrate:
(buillds off #1532)

wires up dns from tun interface while on android.
refactors dns parts so that it can handle such.


majestrate:
this PR will not compile on android yet.

PR #1535, opened 2021-02-13
only do single ack packets

majestrate:
this reduces jitter on the link layer in exchange for higher packet rate.


llarp/iwp/session.cpp
```diff
@@ -822,20 +817,33 @@ namespace llarp
...
         }
-        m_RXMsgs.erase(itr);
       }
     }
 
+    void
+    Session::HandleRecvMsgCompleted(const InboundMessage& msg)
+    {
+      /// should we send a multiack?
+      constexpr bool UseMACK = false;
```
tewinget: if getting rid of multi-ack, should probably get rid of conditionals on it.  can always revert if we change our minds.

majestrate: done

tewinget (approved):

PR #1537, opened 2021-02-15
Android

### LINKED ISSUE #61: Android App
i'd love to get someone to get the vpn for android working
Jeff already has an JNI lib for it

### LINKED ISSUE #386: RFC: take port 53 on all tun adapters?
We would not proxy DNS but ONLY respond to localhost.loki

1. to make it easy to detect configuration (loop on all adapters, talk to DNS server if available, ask about localhost.loki and now you have all the information about that potential tun adapter). The Lokinet registry socket would eliminate the need for this, however this is much easier to implement in the short term.

2. to block the DNS path building (.snode / .loki) from being exposed to the network which is ripe for abuse.

majestrate:
this makes us able to poop out an android apk build.

fixes #61
fixes #386

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -62,8 +62,30 @@ local debian_pipeline(name, image,
         }
     ],
 };
-
-// windows cross compile on alpine linux
+local apk_builder(name, image, extra_cmds=[], allow_fail=true) = {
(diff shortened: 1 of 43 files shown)


tewinget (approved):
Looks good

PR #1538, opened 2021-02-16
ip6 nat exits

majestrate:
enables ip6 exits using nat

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -742,7 +742,7 @@ namespace llarp
         return false;
       }
 
-      const auto maybe = GetInterfaceIP6(m_IfName);
+      const auto maybe = GetInterfaceIPv6Address(m_IfName);
       if (maybe.has_value())
       {
         m_OurIP6 = *maybe;
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -745,8 +745,8 @@ namespace llarp
       const auto maybe = GetInterfaceIPv6Address(m_IfName);
       if (maybe.has_value())
       {
-        m_OurIP6 = *maybe;
-        LogInfo(Name(), " has ipv6 address ", m_OurIP6);
+        m_OurIPv6 = *maybe;
+        LogInfo(Name(), " has ipv6 address ", m_OurIPv6);
       }
 
       netloop->add_ticker([&]() { Flush(); });
@@ -970,7 +970,7 @@ namespace llarp
         }
         else if (pkt.IsV6())
         {
-          dst = m_OurIP6;
+          dst = m_OurIPv6;
           src = pkt.srcv6();
         }
         // find what exit we think this should be for
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -182,8 +182,9 @@ namespace llarp
         }
         m_OurRange = *maybe;
       }
+
       m_OurIP = m_OurRange.addr;
-      m_UseV6 = not m_OurRange.IsV4();
+      m_UseV6 = false;
       return Endpoint::Configure(conf, dnsConf);
     }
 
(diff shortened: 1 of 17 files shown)


llarp/handlers/tun.cpp
```diff
@@ -934,7 +949,10 @@ namespace llarp
           if (pkt.IsV4())
```
tewinget, line 949: This conditional (starting at `if (m_state->m_ExitEnabled)`) feels poorly structured.

llarp/handlers/tun.hpp
```diff
@@ -235,6 +235,9 @@ namespace llarp
       std::unordered_map<huint128_t, llarp_time_t> m_IPActivity;
       /// our ip address (host byte order)
       huint128_t m_OurIP;
+      /// our ip for ip6
```
tewinget: should at least be consistent in comments (ip6 vs ipv6 vs v6)

llarp/net/net.hpp
```diff
@@ -77,6 +77,18 @@ namespace llarp
...
 
+  /// get an interface's ip6 address
+  std::optional<huint128_t>
+  GetInterfaceIP6(std::string ifname);
+
+#ifdef _WIN32
+  namespace net
+  {
+    std::optional<int>
+    GetIfIndex(std::string ip);
```
tewinget: Reads like a conditional get, rather than `If` meaning `interface`

llarp/net/route.cpp
```diff
@@ -265,6 +284,24 @@ namespace llarp::net
...
+  {
+    llarp::huint32_t h{};
+    h.FromString(ip);
+    std::optional<int> ret = std::nullopt;
+    ForEachWIN32Interface([&ret, n = ToNet(h)](auto* iface) {
+      if (ret.has_value())
+        return;
+      if (iface->dwForwardNextHop == n.n)
+      {
+        ret = iface->dwForwardIfIndex;
```
tewinget, line 297: As in `net/net.hpp`, the `If` here reads like a conditional rather than `interface`

llarp/net/route.cpp
```diff
@@ -335,7 +372,7 @@ namespace llarp::net
     int if_idx = if_nametoindex(ifname.c_str());
     _inet_addr to_addr{};
     _inet_addr gw_addr{};
-    const auto maybe = GetIFAddr(ifname);
+    auto maybe = GetIFAddr(ifname);
```
tewinget: contrasted with the above comments, this seems fine because `IF` is both capital letters, but spelling out `Interface` might still be better.

llarp/net/sock_addr.cpp
```diff
@@ -187,6 +188,12 @@ namespace llarp
             sizeof(m_addr.sin6_addr.s6_addr)));
   }
 
+  huint128_t
+  SockAddr::asIPv6() const
+  {
+    return net::In6ToHUInt(m_addr.sin6_addr);
```
tewinget, line 194: I understand this function name `In6ToHUInt` but wow is it ugly as hell.  Not sure how best to change it, if it needs changing.

llarp/vpn/win32.hpp
```diff
@@ -299,19 +302,18 @@ namespace llarp::vpn
...
       for (const auto& ifaddr : m_Info.addrs)
       {
         if (ifaddr.fam == AF_INET6)
         {
-          IPRange range = ifaddr.range;
-          range.netmask_bits = netmask_ipv6_bits(128);
-          NetSH(
-              "interface ipv6 set address " + std::to_string(ifindex) + " " + range.ToString()
-              + " store=active");
+          const auto maybe = net::GetIfIndex(ip.ToString());
```
tewinget: `If` == `interface` again

llarp/net/net.hpp
majestrate: resolved

llarp/handlers/tun.cpp
majestrate: it may but it's the most readable way to do the logic required.

llarp/handlers/tun.hpp
majestrate: updated

majestrate:
> if `GetInterfaceAddr` is for both IPv4 and IPv6, is this function necessary? If so, why is there no IPv4-specific equivalent to this function?

one returns a `IpAddress` and the other returns `huint128_t`.

tewinget (approved):
Looks good.

PR #1539, opened 2021-02-18
add option to enforce unique netblocks per path.

majestrate:
* add new section called `[paths]` for path building options
* add `[paths]:unique-range-size` option to enforce build paths over unique ranges of certain size
* add stub for geoip path selection for layer behind ifdef

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -1022,19 +1022,16 @@ namespace llarp
   bool
   PeerSelectionConfig::Acceptable(const std::set<RouterContact>& rcs) const
   {
-    if (m_UniqueHopsNetmaskSize)
+    const auto netmask = netmask_ipv6_bits(96 + m_UniqueHopsNetmaskSize);
+    std::set<IPRange> seenRanges;
+    for (const auto& hop : rcs)
     {
-      const auto netmask = netmask_ipv6_bits(96 + *m_UniqueHopsNetmaskSize);
-      std::set<IPRange> seenRanges;
-      for (const auto& hop : rcs)
+      for (const auto& addr : hop.addrs)
       {
-        for (const auto& addr : hop.addrs)
+        const auto network_addr = net::In6ToHUInt(addr.ip) & netmask;
+        if (auto [it, inserted] = seenRanges.emplace(network_addr, netmask); not inserted)
         {
-          const auto network_addr = net::In6ToHUInt(addr.ip) & netmask;
-          if (auto [it, inserted] = seenRanges.emplace(network_addr, netmask); not inserted)
-          {
-            return false;
-          }
+          return false;
         }
       }
     }
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -1028,7 +1028,7 @@ namespace llarp
   bool
   PeerSelectionConfig::Acceptable(const std::set<RouterContact>& rcs) const
   {
-    if (m_UniqueHopsNetMask == 0)
+    if (m_UniqueHopsNetmaskSize == 0)
       return true;
     const auto netmask = netmask_ipv6_bits(96 + m_UniqueHopsNetmaskSize);
     std::set<IPRange> seenRanges;
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -1000,10 +1000,10 @@ namespace llarp
             throw std::invalid_argument{"[paths]:unique-range-size must be between 4 and 32"};
           m_UniqueHopsNetmaskSize = arg;
         },
-        Comment{"In path hop selection use this as the default netmkask for testing if we should "
-                "include a relay in our hop list",
-                "i.e. 32 for uniuqe ip addresses for every hop, 24 for all hops are in a different "
-                "/24, 16 for all hops are in a different /16, etc..."});
+        Comment{"Netmask for router path selection; each router must be from a distinct IP subnet "
+                "of the given size.",
+                "E.g. 16 ensures that all routers are using distinct /16 IP addresses."});
+
 #ifdef WITH_GEOIP
     conf.defineOption<std::string>(
         "paths",
@@ -1014,32 +1014,27 @@ namespace llarp
           m_ExcludeCountries.emplace(lowercase_ascii_string(std::move(arg)));
         },
         Comment{"exclude a country given its 2 letter country code from being used in path builds",
-                "i.e. exlcude-country=DE",
-                "can be listed multiple times to exlcude multiple countries"});
+                "e.g. exclude-country=DE",
+                "can be listed multiple times to exclude multiple countries"});
 #endif
   }
 
   bool
-  PeerSelectionConfig::Acceptable(std::set<RouterContact> rcs) const
+  PeerSelectionConfig::Acceptable(const std::set<RouterContact>& rcs) const
   {
     if (m_UniqueHopsNetmaskSize)
     {
-      auto makeRange =
-          [netmask = netmask_ipv6_bits(96 + *m_UniqueHopsNetmaskSize)](IpAddress addr) -> IPRange {
-        return IPRange{net::ExpandV4(addr.toIP()) & netmask, netmask};
-      };
-
+      const auto netmask = netmask_ipv6_bits(96 + *m_UniqueHopsNetmaskSize);
       std::set<IPRange> seenRanges;
       for (const auto& hop : rcs)
       {
         for (const auto& addr : hop.addrs)
         {
-          const auto range = makeRange(addr.toIpAddress());
-          if (seenRanges.count(range))
+          const auto network_addr = net::In6ToHUInt(addr.ip) & netmask;
+          if (auto [it, inserted] = seenRanges.emplace(network_addr, netmask); not inserted)
           {
             return false;
           }
-          seenRanges.emplace(range);
         }
       }
     }
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -996,8 +996,14 @@ namespace llarp
         DefaultUniqueCIDR,
         ClientOnly,
         [=](int arg) {
-          if (arg > 32 or arg < 4)
+          if (arg == 0)
+          {
+            m_UniqueHopsNetmaskSize = arg;
+          }
+          else if (arg > 32 or arg < 4)
+          {
             throw std::invalid_argument{"[paths]:unique-range-size must be between 4 and 32"};
+          }
           m_UniqueHopsNetmaskSize = arg;
         },
         Comment{"Netmask for router path selection; each router must be from a distinct IP subnet "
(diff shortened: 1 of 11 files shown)


llarp/config/config.cpp
```diff
@@ -983,6 +983,69 @@ namespace llarp
...
+        "paths",
+        "unique-range-size",
+        DefaultUniqueCIDR,
+        ClientOnly,
+        [=](int arg) {
+          if (arg > 32 or arg < 4)
+            throw std::invalid_argument{"[paths]:unique-range-size must be between 4 and 32"};
+          m_UniqueHopsNetmaskSize = arg;
+        },
+        Comment{"In path hop selection use this as the default netmkask for testing if we should "
```
jagerman: "netmkask"

llarp/config/config.cpp
```diff
@@ -983,6 +983,69 @@ namespace llarp
...
+                "include a relay in our hop list",
+                "i.e. 32 for uniuqe ip addresses for every hop, 24 for all hops are in a different "
```
jagerman: "uniuqe"

llarp/config/config.cpp
```diff
@@ -983,6 +983,69 @@ namespace llarp
...
+    (void)params;
+
+    constexpr Default DefaultUniqueCIDR{32};
+    conf.defineOption<int>(
+        "paths",
+        "unique-range-size",
+        DefaultUniqueCIDR,
+        ClientOnly,
+        [=](int arg) {
+          if (arg > 32 or arg < 4)
```
jagerman: Should be weak inequalities

llarp/config/config.cpp
```diff
@@ -983,6 +983,69 @@ namespace llarp
...
+        "paths",
+        "exclude-country",
+        ClientOnly,
+        MultiValue,
+        [=](std::string arg) {
+          m_ExcludeCountries.emplace(lowercase_ascii_string(std::move(arg)));
+        },
+        Comment{"exclude a country given its 2 letter country code from being used in path builds",
+                "i.e. exlcude-country=DE",
+                "can be listed multiple times to exlcude multiple countries"});
```
jagerman: "exlcude" (twice)

llarp/config/config.cpp
jagerman: I think it should also accept a "do nothing" option (maybe 0?).  32 isn't the same because it excludes routers on different ports on the same IP.

I think the default be something larger than /32, perhaps /24?

llarp/config/config.cpp
```diff
@@ -983,6 +983,69 @@ namespace llarp
...
+        ClientOnly,
+        [=](int arg) {
+          if (arg > 32 or arg < 4)
+            throw std::invalid_argument{"[paths]:unique-range-size must be between 4 and 32"};
+          m_UniqueHopsNetmaskSize = arg;
+        },
+        Comment{"In path hop selection use this as the default netmkask for testing if we should "
+                "include a relay in our hop list",
+                "i.e. 32 for uniuqe ip addresses for every hop, 24 for all hops are in a different "
+                "/24, 16 for all hops are in a different /16, etc..."});
```
jagerman: This wording is a bit clunky.  Perhaps:

"Netmask for router path selection; each router must be from a distinct IP subnet of the given size. E.g. 16 ensures that all routers are using distinct /16 IP addresses."

llarp/config/config.cpp
```diff
@@ -983,6 +983,69 @@ namespace llarp
...
+    conf.defineOption<std::string>(
+        "paths",
+        "exclude-country",
+        ClientOnly,
+        MultiValue,
+        [=](std::string arg) {
+          m_ExcludeCountries.emplace(lowercase_ascii_string(std::move(arg)));
+        },
+        Comment{"exclude a country given its 2 letter country code from being used in path builds",
+                "i.e. exlcude-country=DE",
```
jagerman: "i.e." -> "e.g."  (i.e. means essentially "in other words", e.g. means essentially "for example")

llarp/config/config.cpp
```diff
@@ -983,6 +983,69 @@ namespace llarp
...
+      for (const auto& hop : rcs)
+      {
+        for (const auto& addr : hop.addrs)
+        {
+          const auto range = makeRange(addr.toIpAddress());
+          if (seenRanges.count(range))
+          {
+            return false;
+          }
+          seenRanges.emplace(range);
```
jagerman: Simplification (avoids a double-search):
```C++
if (auto [it, inserted] = seenRanges.insert(std::move(range)); inserted)
{
  return false;
}
```
Also changed the `emplace` to an `insert`: the intent of `emplace` is to explicitly construct while inserting by passing constructor arguments and deferring construction to the container code, e.g. `as.emplace(x, y, z)` is meant as a more efficient version of `as.insert(A(x, y, z))`.  Here the intent is only to insert a pre-existing element, not to construct one, and so `insert` offers clearer intentions.

jagerman: Actually a further simplification that *does* use emplace for deferred construction:
```C++
      auto netmask = netmask_ipv6_bits(96 + *m_UniqueHopsNetmaskSize);
      std::set<IPRange> seenRanges;
      for (const auto& hop : rcs)
      {
        for (const auto& addr : hop.addrs)
        {
          auto network_addr = net::ExpandV4(addr.toIP()) & netmask;
          if (auto [it, inserted] = seenRanges.emplace(network_addr, netmask);
              inserted)
          {
            return false;
          }
        }
      }
```

llarp/path/pathbuilder.hpp
```diff
@@ -93,17 +93,17 @@ namespace llarp
       bool
       BuildOneAlignedTo(const RouterID endpoint) override;
 
-      virtual std::optional<std::vector<RouterContact>>
-      GetHopsAlignedToForBuild(RouterID endpoint);
+      std::optional<std::vector<RouterContact>>
+      GetHopsAlignedToForBuild(RouterID endpoint, std::set<RouterID> exclude = {});
```
jagerman: These `set` arguments should be const lvalue references

llarp/path/pathbuilder.hpp
```diff
@@ -93,17 +93,17 @@ namespace llarp
...
 
       void
       Build(std::vector<RouterContact> hops, PathRole roles = ePathRoleAny) override;
 
       /// pick a first hop
-      virtual std::optional<RouterContact>
-      SelectFirstHop() const;
+      std::optional<RouterContact>
+      SelectFirstHop(std::set<RouterID> exclude = {}) const;
```
jagerman: and here

llarp/config/config.hpp
```diff
@@ -69,6 +71,25 @@ namespace llarp
...
+
+    /// set of countrys to exclude from path building (2 char country code)
+    std::unordered_set<std::string> m_ExcludeCountries;
+
+    void
+    defineConfigOptions(ConfigDefinition& conf, const ConfigGenParameters& params);
+
+    /// return true if this set of router contacts is acceptable against this config
+    bool
+    Acceptable(std::set<RouterContact> hops) const;
```
jagerman: const lvalue ref

llarp/path/pathbuilder.cpp
```diff
@@ -292,44 +295,63 @@ namespace llarp
...
-              });
-
-          if (not maybe.has_value())
+          auto filter =
+              [&hops, r = m_router, endpointRC, pathConfig, exclude](const auto& rc) -> bool {
+            if (exclude.count(rc.pubkey))
+              return false;
+
+            std::set<RouterContact> hopsSet;
+            hopsSet.emplace(endpointRC);
```
jagerman: `insert` (see comment above about insert vs emplace)

llarp/path/pathbuilder.cpp
```diff
@@ -292,44 +295,63 @@ namespace llarp
...
+            for (const auto& hop : hops)
+              hopsSet.emplace(hop);
```
jagerman: `hopsSet.insert(hops.begin(), hops.end());`

llarp/path/pathbuilder.cpp
```diff
@@ -292,44 +295,63 @@ namespace llarp
...
+            }
+
+            hopsSet.emplace(rc);
+            if (not pathConfig.Acceptable(hopsSet))
+              return false;
+
+            return rc.pubkey != endpointRC.pubkey;
+          };
+
+          if (const auto maybe = m_router->nodedb()->GetRandom(filter))
```
jagerman, line 350: This has some non-uniform selection problems; discussion off github.

jagerman: Problem: GetRandom()'s implementation isn't random enough: if the filter excludes a big chunk of nodes then we have a good chance of selecting into that chunk, in which case we will heavily favour selection of the node that immediately follows the bad chunk.

Potential solution: make GetRandom shuffle a set of indices (or std::ref's, if we don't have random access into the list) and walk through those rather than walking sequentially.

llarp/config/config.cpp
majestrate: if not provided the optional in the config will be empty so it won't be done if it's not provided explicitly.

majestrate:
should we default this to on?

llarp/config/config.cpp
majestrate: i was wrong, it defaults to /32

llarp/nodedb.hpp
```diff
@@ -96,8 +96,7 @@ namespace llarp
       for (const auto& entry : m_Entries)
         keys.emplace_back(entry.first);
 
-      std::shuffle(
-          keys.begin(), keys.end(), std::mt19937{static_cast<unsigned int>(llarp::randint())});
+      std::shuffle(keys.begin(), keys.end(), std::mt19937_64{llarp::randint()});
```
jagerman: `std::mt19937_64` construction is quite heavy (but drawing from it is very fast).  A better approach here would be to have a global, thread_local mt19937_64 around in some header:

```c++
inline thread_local std::mt19937_64 rng{std::random_device{}()};
```
or header/cpp pair:
```c++
extern thread_local std::mt19937_64 rng;
```
```c++
thread_local std::mt19937_64 rng{std::random_device{}()};
```
and then just use this directly wherever non-crypto psuedo-rng is needed (like here).  Most compilers will defer the construction until first use of the variable within a thread, *not* on every thread construction.

jagerman: `llarp::randint()` in the constructor might also work fine here, but I'm not sure how soon that is available.

jagerman: Actually, perhaps a better approach is just to implement a [URNG-implementing class](https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator) around llarp::randint() and drop mt19937_64.  I think this would do it:

```C++
struct RNG {
  using result_type = uint64_t;
  static constexpr uint64_t min() { return std::numeric_limits<uint64_t>::min(); }
  static constexpr uint64_t max() { return std::numeric_limits<uint64_t>::max(); }
  uint64_t operator() { return llarp::randint(); }
};
```

jagerman:
Would be nice to apply CSRNG to rc_lookup_handler.cpp and service/outbound_context.cpp instead of the MTs they are using now.

majestrate:
> Would be nice to apply CSRNG to rc_lookup_handler.cpp and service/outbound_context.cpp instead of the MTs they are using now.

would probably be best to do it in its own distinct PR after this is merged.

jagerman (approved):

PR #1541, opened 2021-02-19
lns lookup consensus

majestrate:
ask all service nodes we can for each lns lookup and only use the result if they all agree

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -802,7 +802,7 @@ namespace llarp
     void
     Endpoint::LookupNameAsync(
         std::string name,
-        std::function<void(std::optional<std::variant<Address, RouterID>>)> handler)
+        std::function<void(std::optional<var::variant<Address, RouterID>>)> handler)
     {
       auto& cache = m_state->nameCache;
       const auto maybe = cache.Get(name);
@@ -827,9 +827,16 @@ namespace llarp
       }
 
       auto maybeInvalidateCache = [handler, &cache, name](auto result) {
-        if (result and result->IsZero())
+        if (result)
         {
-          result = std::nullopt;
+          var::visit(
+              [&](auto&& value) {
+                if (value.IsZero())
+                {
+                  result = std::nullopt;
+                }
+              },
+              *result);
         }
         if (result)
         {
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -232,12 +232,13 @@ namespace llarp
                   const auto maybe_auth = info.second;
 
                   m_StartupLNSMappings.erase(name);
-
-                  if (maybe_range.has_value())
-                    m_ExitMap.Insert(*maybe_range, *maybe_addr);
-
-                  if (maybe_auth.has_value())
-                    SetAuthInfoForEndpoint(*maybe_addr, *maybe_auth);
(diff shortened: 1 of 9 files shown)


tewinget (changes requested):
the result callback has non-obvious and ideally unnecessary side-effects, and is generally cursed.

llarp/service/endpoint.cpp
```diff
@@ -808,12 +808,43 @@ namespace llarp
...
+        if (path->IsReady())
+        {
+          paths.insert(path);
+        }
+      });
+      // not enough paths
+      if (paths.size() < 3)
+      {
+        handler(std::nullopt);
+        return true;
```
tewinget: function now always returns true; should change to void

llarp/service/lns_tracker.hpp
```diff
@@ -0,0 +1,45 @@
...
+      std::unordered_set<Address> m_CurrentValues;
+      std::function<void(std::optional<Address>)> m_HandleResult;
+      std::size_t m_ResultsGotten = 0;
+      std::size_t m_ResultsNeeded;
+
+      LookupInfo(std::size_t wantResults, std::function<void(std::optional<Address>)> resultHandler)
+          : m_HandleResult{std::move(resultHandler)}, m_ResultsNeeded{wantResults}
+      {}
+
+      /// return true to remove self
```
tewinget: that's a very un-intuitive use of a boolean return.

perhaps add a function "Done" or similar, then above when using do

```
if (itr = find):
    itr->HandleOneResult;
    if itr->Done():
        erase(itr)
```

llarp/service/endpoint.cpp
```diff
@@ -808,12 +808,43 @@ namespace llarp
...
+        handler(result);
+      };
+
+      auto resultHandler =
+          m_state->lnsTracker.MakeResultHandler(name, paths.size(), maybeInvalidateCache);
+
+      for (const auto& path : paths)
+      {
+        LogInfo(Name(), " lookup ", name, " from ", path->Endpoint());
+        auto job = new LookupNameJob(this, GenTXID(), name, resultHandler);
```
tewinget, line 858: ownership very unclear, I smell cursed shenanigans

llarp/service/endpoint.cpp
majestrate: fixed

llarp/service/endpoint.cpp
majestrate: beware the abyss

llarp/service/lns_tracker.hpp
majestrate: fixed

llarp/service/endpoint.cpp
```diff
@@ -802,7 +802,7 @@ namespace llarp
     void
     Endpoint::LookupNameAsync(
         std::string name,
-        std::function<void(std::optional<std::variant<Address, RouterID>>)> handler)
+        std::function<void(std::optional<var::variant<Address, RouterID>>)> handler)
```
jagerman: This should stay `std::variant`: the `var` function provides just `var::get` and `var::visit` which work on `std::variant` which mac *does* support as long as you don't use `std::get` or `std::visit` on them.

majestrate: fixed

tewinget (approved):
should be fine, but still a bit ugly (re: handlers on handlers on handlers).

PR #1543, opened 2021-02-22
remove packet rate based cpu bottleneck

majestrate:
this removed the packet based cpu use bottleneck which came from primarily a lack of coalescing of udp reads.

this also removes use of `llarp::IpAddress` from hotpaths as it does a lot of unneeded allocations.

### DIFF
--- a/llarp/iwp/linklayer.hpp
+++ b/llarp/iwp/linklayer.hpp
@@ -51,7 +51,7 @@ namespace llarp::iwp
     MapAddr(const RouterID& pk, ILinkSession* s) override;
 
     void
-    UnmapAddr(const IpAddress& addr);
+    UnmapAddr(const SockAddr& addr);
 
     void
     WakeupPlaintext();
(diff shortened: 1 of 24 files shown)


llarp/iwp/linklayer.hpp
```diff
@@ -7,56 +7,69 @@
 #include <crypto/types.hpp>
 #include <link/server.hpp>
 #include <config/key_manager.hpp>
-
+#include <util/thread/queue.hpp>
```
tewinget: maybe I'm missing something but I don't see this new include being used in this file?

majestrate: you're right.

llarp/net/address_info.hpp
```diff
@@ -44,9 +44,9 @@ namespace llarp
     IpAddress
     toIpAddress() const;
 
-    /// Updates our ip and port to reflact that of the given IpAddress
+    /// Updates our ip and port to reflact that of the given SockAddr
```
tewinget: s/reflact/reflect

tewinget (approved):
couple minor things but otherwise LGTM

PR #1548, opened 2021-02-25
add mx records to lns names

majestrate:
this allows sending and receiving mail from human readable names like `probably.loki`

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -412,18 +412,17 @@ namespace llarp
         }
         else if (service::NameIsValid(lnsName))
         {
-          return LookupNameAsync(
-              lnsName, [msg = std::make_shared<dns::Message>(msg), lnsName, reply](auto maybe) {
-                if (maybe.has_value())
-                {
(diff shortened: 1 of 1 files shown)


llarp/handlers/tun.cpp
```diff
@@ -407,7 +407,24 @@ namespace llarp
         service::Address addr;
         if (addr.FromString(qname, ".loki") || addr.FromString(qname, ".snode")
             || is_random_snode(msg) || is_localhost_loki(msg))
+        {
           msg.AddMXReply(qname, 1);
+        }
+        else if (service::NameIsValid(lnsName))
+        {
+          return LookupNameAsync(
+              lnsName, [msg = std::make_shared<dns::Message>(msg), lnsName, reply](auto maybe) {
```
jagerman: I don't think this needs to be a shared_ptr, just capture `msg` (to copy it) should be enough?

majestrate: you're right. fixed, needed a mutable lambda

jagerman (approved):

PR #1560, opened 2021-03-03
Allow multiple gradles to run on the same box

jagerman:
/cache is shared among multiple runners that can potentiall run at
once, but gradle's cache is retarded and just completely blocks anything
from using the cache for the span of the entire build.  Work around it
by sticking the unique builder identifier in the cache path.


tewinget (approved):
yup.

PR #1575, opened 2021-03-10
let format.sh try system clang-format in case it is the correct version

tewinget:


contrib/format.sh
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env bash
 
+CLANG_FORMAT_DESIRED_VERSION=11
+
+binary=$(which clang-format-11 2>/dev/null)
```
jagerman: $CLANG_FORMAT_DESIRED_VERSION

jagerman (approved):

PR #1580, opened 2021-03-18
remove link layer throughput limit

majestrate:
I suspect this is what is causing most of the bottleneck on the network so we should remove this limit for now.


majestrate:
actually no this is a really bad idea

PR #1581, opened 2021-03-18
make ipv6 range used by exit traffic configurable.

majestrate:
* add ip6-range option to network section to control which range we want to use for ipv6 traffic
* make ip6-range able to disable ipv6 exits ( i dont like this but eh )


tewinget (approved):
LGTM

PR #1582, opened 2021-03-18
fix issue #1562

### LINKED ISSUE #1562: > Would be nice to apply CSRNG to rc_lookup_handler.cpp and service/outbound_context.cpp instead of the MTs they are using now.
> Would be nice to apply CSRNG to rc_lookup_handler.cpp and service/outbound_context.cpp instead of the MTs they are using now.

would probably be best to do it in its own distinct PR after this is merged.

_Originally posted by @majestrate in https://github.com/oxen-io/loki-network/issues/1539#issuecomment-789973162_

majestrate:
fixes #1562 
* use csrng for std::shuffle

### DIFF
--- a/llarp/router/rc_lookup_handler.cpp
+++ b/llarp/router/rc_lookup_handler.cpp
@@ -270,8 +270,7 @@ namespace llarp
 
       if (lookupRouters.size() > LookupPerTick)
       {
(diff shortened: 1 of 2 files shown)


tewinget (approved):
makes sense

PR #1583, opened 2021-03-20
allow failure of setting ipv6 address

majestrate:
this allows a soft fail when we don't have permission to set ipv6 address. this is for systems that do not allow ipv6 routes.


tewinget (approved):
seems fine

PR #1598, opened 2021-04-16
new logos and new lokinet-bootstrap backport for stable

majestrate:
cherry-pick rebranding and lokinet-bootstrap for stable branch


jagerman:
This isn't building -- looks like lokinet-bootstrap is depending on llarp stuff but missing the cmake dep?  That said, I don't think we *want* the liblokinet dep linked into lokinet-bootstrap, so this might need a bigger fix.

daemon/lokinet-bootstrap.cpp
```diff
@@ -0,0 +1,64 @@
...
+}  // namespace
+
+int
+main(int argc, char* argv[])
+{
+  std::string bootstrap_url{"https://seed.lokinet.org/lokinet.signed"};
+  if (argc > 1)
+  {
+    bootstrap_url = argv[1];
+  }
```
jagerman, line 48: This should also support passing
"mainnet" or "lokinet" for the mainnet bootstrap
"testnet" for the testnet bootstrap

daemon/lokinet-bootstrap.cpp
```diff
@@ -0,0 +1,64 @@
...
+  ss << resp.text;
+
+  std::string data{ss.str()};
+  llarp_buffer_t buf{&data[0], data.size()};
+
+  llarp::RouterContact rc;
+  if (not rc.BDecode(&buf))
+    return exit_with_message("invalid bootstrap data was fetched", 1);
+
+  const auto path = llarp::GetDefaultBootstrap();
```
jagerman: If there is an argv[2], use that instead of the default path

tewinget (approved):

PR #1620, opened 2021-05-01
make default log level warn 

### LINKED ISSUE #1593: Change default log level to warn
For most common users (both SN operators and clients) lokinet produces a large volume of log output which is probably only really useful for advanced users and developers; for the most part the logs just fill up syslog.

I suggest we lower the default to warn level.

However this isn't quite so trivial: we output quite a bit of useful information at startup at `info` level, which *is* useful to log.  So perhaps, if no log level has been specified on command line or in the config, then at startup the log level continues to be `info`, but once everything is initialized and running we drop to `warn` log level.

majestrate:
fixes #1593

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -989,7 +989,7 @@ namespace llarp
 
     constexpr Default DefaultLogType{"file"};
     constexpr Default DefaultLogFile{""};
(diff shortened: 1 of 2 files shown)


tewinget (approved):
makes sense

PR #1622, opened 2021-05-03
stricter profiling

majestrate:
decrease chances in profiling from 8 to 4, this will make clients far less forgiving for path failures and especially path timeouts.

### DIFF
--- a/llarp/profiling.hpp
+++ b/llarp/profiling.hpp
@@ -50,17 +50,20 @@ namespace llarp
   {
     Profiling();
 
+    inline static const int profiling_chances = 4;
+
     /// generic variant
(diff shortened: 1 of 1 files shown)


llarp/profiling.cpp
```diff
@@ -75,7 +75,8 @@ namespace llarp
   {
     if (connectTimeoutCount > chances)
```
tewinget: probably a good time to comment this about what it's supposed to be doing / why.

majestrate: excluded this hunk it's not needed.

tewinget (approved):

PR #1623, opened 2021-05-03
dont use non tested paths

majestrate:
* if a path's latency is zero dont use it because it's not actually a zero latency path it's probably about to be failed or timed out
* increase default path alignment timeout

### DIFF
--- a/llarp/path/pathset.cpp
+++ b/llarp/path/pathset.cpp
@@ -156,7 +156,8 @@ namespace llarp
           {
             if (chosen == nullptr)
               chosen = itr->second;
-            else if (chosen->intro.latency > itr->second->intro.latency)
+            else if (
+                chosen->intro.latency != 0s and chosen->intro.latency > itr->second->intro.latency)
               chosen = itr->second;
           }
         }
@@ -429,7 +430,7 @@ namespace llarp
       llarp_time_t minLatency = 30s;
       for (const auto& path : established)
       {
-        if (path->intro.latency < minLatency)
+        if (path->intro.latency < minLatency and path->intro.latency != 0s)
         {
           minLatency = path->intro.latency;
           chosen = path;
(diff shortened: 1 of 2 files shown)


llarp/path/pathset.cpp
```diff
@@ -156,7 +156,7 @@ namespace llarp
           {
             if (chosen == nullptr)
               chosen = itr->second;
-            else if (chosen->intro.latency > itr->second->intro.latency)
+            else if (chosen->intro.latency > 0s and chosen->intro.latency > itr->second->intro.latency)
```
tewinget: very tiny nitpick, but it feels like it should be `!= 0s` instead of `> 0s`, to better convey that what we care about is that it is nonzero.  Obviously any nonzero in this case is also positive, so the conditions are equivalent.

llarp/path/pathset.cpp
```diff
@@ -429,7 +429,7 @@ namespace llarp
       llarp_time_t minLatency = 30s;
       for (const auto& path : established)
       {
-        if (path->intro.latency < minLatency)
+        if (path->intro.latency < minLatency and path->intro.latency > 0s)
```
tewinget: as above, consider `!= 0s` in favor of `> 0s`

llarp/path/pathset.cpp
majestrate: good point

tewinget (approved):

llarp/path/pathset.cpp
```diff
@@ -156,7 +156,8 @@ namespace llarp
           {
             if (chosen == nullptr)
               chosen = itr->second;
-            else if (chosen->intro.latency > itr->second->intro.latency)
+            else if (
+                chosen->intro.latency != 0s and chosen->intro.latency > itr->second->intro.latency)
```
jagerman, line 160: Could this potentially be an issue with localhost test networks?  Maybe the latency could be an `optional` or set to -1s or something?

PR #1624, opened 2021-05-03
dont invalidate cache on lns timeout, only on explicit negative

majestrate:


llarp/service/session.cpp
```diff
@@ -27,7 +27,7 @@ namespace llarp
       const auto lastUsed = std::max(lastSend, lastRecv);
       if (lastUsed == 0s)
         return intro.IsExpired(now);
-      return now > lastUsed && (now - lastUsed > lifetime || intro.IsExpired(now));
+      return now > lastUsed && now - lastUsed > lifetime;
```
tewinget: probably keep the parentheses around the latter condition, just for ease of reading.

majestrate: fair

majestrate:
rebased atop dev

tewinget (approved):
makes sense

PR #1626, opened 2021-05-05
limit path builds across all builders

majestrate:
this limits path builds per edge router for all path builders


tewinget:
Would it be better to refactor such that we only have one instance of path builder ever rather than having them be ephemeral?

majestrate:
> Would it be better to refactor such that we only have one instance of path builder ever rather than having them be ephemeral?

no

tewinget (approved):
seems fine, if a bit ugly

PR #1627, opened 2021-05-05
fixups in service endpont

majestrate:
* increase publish introset timeout so that it does not time out on the network
* remove pedantic log warn
* make sure the path we are using for replying on inbound sessions is alive
* include convotag in log message so we know wtf is going on

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -622,8 +622,12 @@ namespace llarp
       Endpoint* m_Endpoint;
       uint64_t m_relayOrder;
       PublishIntroSetJob(
-          Endpoint* parent, uint64_t id, EncryptedIntroSet introset, uint64_t relayOrder)
-          : IServiceLookup(parent, id, "PublishIntroSet")
+          Endpoint* parent,
+          uint64_t id,
+          EncryptedIntroSet introset,
+          uint64_t relayOrder,
+          llarp_time_t timeout)
+          : IServiceLookup(parent, id, "PublishIntroSet", timeout)
           , m_IntroSet(std::move(introset))
           , m_Endpoint(parent)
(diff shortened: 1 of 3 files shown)


majestrate:
additionally we were marking paths as dead as we were not doing path tests.... :potato:

llarp/service/endpoint.cpp
```diff
@@ -1379,7 +1386,7 @@ namespace llarp
               path->Endpoint(),
               order,
               GenTXID(),
-              timeout);
+              timeout + (path->intro.latency * 2));
```
tewinget: seems arbitrary on the surface; could use a comment as to why this addition.

majestrate:
removed the network fixing commit from this PR and made into its own PR #1628

llarp/service/outbound_context.cpp
```diff
@@ -567,7 +567,7 @@ namespace llarp
         // verify source
         if (!frame.Verify(si))
         {
-          LogWarn("signature failed");
+          LogWarn("signature failed, T=", frame.T);
```
tewinget: should be "signature verification failed", right?  Also perhaps instead of `T=` put a descriptive name.

llarp/service/outbound_context.cpp
```diff
@@ -567,7 +567,7 @@ namespace llarp
         // verify source
         if (!frame.Verify(si))
         {
-          LogWarn("signature failed");
+          LogWarn("signature verification failed, T=", frame.T);
```
tewinget, line 570: Would prefer to change the `T=` to something descriptive (i.e. `ConvoTag=`) but we can do that across the codebase atomically, per #1632.

tewinget (approved):

PR #1634, opened 2021-05-08
increase session reestablish timeout

majestrate:
when we have to reestablish an outbound session use a much much higher timeout for restablishment

so it doesn't time out and get into a state that's totally screwed.

### DIFF
--- a/llarp/service/endpoint.hpp
+++ b/llarp/service/endpoint.hpp
@@ -306,6 +306,12 @@ namespace llarp
       bool
       ShouldBuildMore(llarp_time_t now) const override;
 
+      virtual llarp_time_t
+      DefaultPathAlignmentTimeout() const
+      {
+        return 10s;
+      }
+
       bool
       EnsurePathTo(
           std::variant<Address, RouterID> addr,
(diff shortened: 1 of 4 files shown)


llarp/service/endpoint.hpp
```diff
@@ -306,6 +306,12 @@ namespace llarp
       bool
       ShouldBuildMore(llarp_time_t now) const override;
 
+      virtual llarp_time_t
+      DefaultPathAlignmentTimeout() const
+      {
+        return 10s;
```
tewinget: I know it's in a function with a descriptive name, but this value is a named constant in `tun.hpp` and probably should be here as well.

majestrate: it's also actually overridden by that file too :^)

tewinget: On that note, probably should take `Default` out of the method name, and then name a constant as the default and return that here.

majestrate: fair

tewinget (approved):

PR #1636, opened 2021-05-09
shift off bad intros

majestrate:
when we get a discard message in the event of an expired intro shift off the intro we are using

### DIFF
--- a/llarp/service/outbound_context.cpp
+++ b/llarp/service/outbound_context.cpp
@@ -46,6 +46,7 @@ namespace llarp
       {
         LogWarn(Name(), " message ", seq, " dropped by endpoint ", p->Endpoint(), " via ", dst);
         MarkCurrentIntroBad(Now());
+        ShiftIntroduction(false);
       }
       return true;
     }


llarp/service/outbound_context.cpp
```diff
@@ -46,6 +46,7 @@ namespace llarp
       {
         LogWarn(Name(), " message ", seq, " dropped by endpoint ", p->Endpoint(), " via ", dst);
         MarkCurrentIntroBad(Now());
+        ShiftIntroduction(false);
       }
       return true;
```
tewinget, line 51: while you're here modifying this method...perhaps change it to `void` return type?

majestrate: good idea

majestrate: actually this is an implementation of a virtual function that uses the return value.

tewinget: that's...that's some code smell, that is.

majestrate: it's in a hot path so i don't want to throw also it's not really that exceptional of a case, unless it is.

tewinget (approved):

PR #1638, opened 2021-05-11
refactor logging to use std::source_location

majestrate:
a long overdue refactor for logging to not use macro hell.

* refactor logging to use `std::source_location`
* fix a bug in android logger when it doesn't respect log none option
* remove json log stream type as it's stupid and unused
* with apple platforms you need to think different :tm: (read: think incorrectly) so apple does not implement `std::source_location`, hence we provide an implementation of it inline just for them.


llarp/util/bencode.hpp
```diff
@@ -75,7 +75,7 @@ namespace llarp
     {
       if (!item.BDecode(buf))
       {
-        llarp::LogWarnTag("llarp/bencode.hpp", "failed to decode key ", k, " for entry in dict");
+        llarp::LogWarn("failed to decode key ", k, " for entry in dict");
```
tewinget, line 78: if we're removing `llarp::` from log statements within the `llarp` namespace, this file should be done as well.

llarp/util/logging/source_location.hpp
```diff
@@ -0,0 +1,72 @@
...
+{
+  struct source_location
+  {
+   public:
+    static constexpr source_location
+    current(
+        const char* fileName = __builtin_FILE(),
+        const char* functionName = __builtin_FUNCTION(),
+        const uint_least32_t lineNumber = __builtin_LINE(),
+#if defined(__GNUC__) and (__GNUC__ > 4 or (__GNUC__ == 4 and __GNUC_MINOR__ >= 8))
```
jagerman: This macro logic doesn't make sense to me.  It should probably just drop the version check, but the version logic here is strange because it's going to *use* `__builtin_COLUMN` for ancient gcc?

majestrate: technically we don't even need column either so we can omit it entirely in this implementation.

llarp/util/bencode.hpp
tewinget: bump

tewinget: meh, can always do some other time I suppose.

llarp/util/logging/source_location.hpp
majestrate: updated

tewinget (approved):
LGTM

PR #1639, opened 2021-05-11
prevent router connect spam when deregistered

majestrate:
if we look deregistered we will now:

* not gossip our rc
* not explore the network to prevent outbound session attempts
* not establish sessions to other service nodes
* close all open sessions we have to tell clients we don't want them

additionally catch exceptions flushing peerdb in disk thread

### DIFF
--- a/llarp/router/rc_lookup_handler.cpp
+++ b/llarp/router/rc_lookup_handler.cpp
@@ -49,7 +49,7 @@ namespace llarp
   }
 
   bool
-  RCLookupHandler::HaveReceivedWhitelist()
+  RCLookupHandler::HaveReceivedWhitelist() const
   {
     util::Lock l(_mutex);
     return not whitelistRouters.empty();
@@ -127,14 +127,12 @@ namespace llarp
       return false;
     }
 
-    util::Lock l(_mutex);
+    if (not useWhitelist)
+      return true;
 
-    if (useWhitelist && whitelistRouters.find(remote) == whitelistRouters.end())
-    {
-      return false;
-    }
+    util::Lock lock{_mutex};
 
-    return true;
+    return whitelistRouters.count(remote);
   }
 
   bool
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -375,8 +375,20 @@ namespace llarp
   }
 
   bool
+  Router::LooksDeregistered() const
+  {
+    return IsServiceNode() and whitelistRouters and _rcLookupHandler.HaveReceivedWhitelist()
+        and not _rcLookupHandler.RemoteIsAllowed(pubkey());
+  }
+
+  bool
   Router::ConnectionToRouterAllowed(const RouterID& router) const
   {
+    if (LooksDeregistered())
+    {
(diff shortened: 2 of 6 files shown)


llarp/router/router.cpp
```diff
@@ -886,7 +898,16 @@ namespace llarp
...
         LogDebug("Queing database flush...");
-        QueueDiskIO([this]() { m_peerDb->flushDatabase(); });
+        QueueDiskIO([this]() {
+          try
+          {
+            m_peerDb->flushDatabase();
+          }
+          catch (std::exception& ex)
+          {
+            LogError("did not flush peer stats: ", ex.what());
```
tewinget: "did not" sounds like a decision was made, probably "could not" or "failed to" instead.

majestrate: fixed

llarp/router/router.cpp
```diff
@@ -859,7 +861,17 @@ namespace llarp
...
     }
 
-    if (connected < connectToNum)
+    if (looksDeregistered)
+    {
+      // kill all sessions that are open because we think we are deregistered
+      _linkManager.ForEachPeer([](auto* peer) {
+        if (peer)
+          peer->Close();
+      });
```
jagerman, line 881: Isn't this a bit harsh?  If I have a path going through router X, and X gets decommissioned, this is going to immediately kill all the paths through it, which is going to result in client-side failures.  Could we do something a bit more graceful here, such as putting this behind a timer so that we kill all paths only once we have looked deregistered for more than xx minutes (where xx = whatever maximum path validity is, I think 20min?).

jagerman:
I don't think this will solve the issue I'm seeing: what I see is that *my* router is trying to establish a session with other routers that are no longer on the network.  So it establishes a session, the other end accepts it (probably because I am registered), then my end immediately kills the session it just established because the other end is deregistered.

This change would, I think, make the other end refuse it earlier, but we shouldn't be trying to establish an outbound session to a deregistered router in the first place.

llarp/router/router.cpp
majestrate: being decommissioned means that other routers won't connect to us so this will propagate that failure information to clients a lot faster than if we didn't do this.

majestrate: i want to shut down the inbound links too in this case but that would take a LOT of effort to spin them back up again.

llarp/router/rc_lookup_handler.cpp
```diff
@@ -127,14 +127,12 @@ namespace llarp
...
+      return true;
 
-    if (useWhitelist && whitelistRouters.find(remote) == whitelistRouters.end())
-    {
-      return false;
-    }
+    util::Lock lock{_mutex};
 
-    return true;
+    return whitelistRouters.find(remote) != whitelistRouters.end();
```
jagerman:
```suggestion
    return whitelistRouters.count(remote);
```

tewinget (approved):
looks fine to squash and merge

majestrate:
squashed

PR #1640, opened 2021-05-12
add format verify option to contrib/format.sh

majestrate:
* add format verify mode to `contrib/format.sh` (run with `./contrib/format.sh verify`) exits nonzero if you need to run the script
* make `contrib/format.sh` non-gnu portable (for macos)
* add my pre-push git hook so it doesn't get lost


tewinget (approved):
the comments in your pre-commit hook are...amusing

PR #1643, opened 2021-05-13
win32 fixups

majestrate:
* correctly poke routes when exit added via rpc
* include liblokinet option in installer
* allow exit "null" for sinkholing internet traffic in `.loki` only mode to be added via rpc


llarp/rpc/rpc_server.cpp
```diff
@@ -456,39 +456,54 @@ namespace llarp::rpc
                   {
                     auto mapExit = [=](service::Address addr) mutable {
                       ep->MapExitRange(range, addr);
+                      r->routePoker().Enable();
```
tewinget, line 459: this should be a result of the MapExitRange call, not the caller's responsibility.

tewinget (approved):
should work, but is not DRY.  making issue for future refactor.

PR #1647, opened 2021-05-15
last minute fixes

majestrate:
fixes the last of the noticeable problems:

* exits work again
* convos no longer die despite being active
* re-enable multi threading on clients only

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -933,7 +933,6 @@ namespace llarp
               PathAlignmentTimeout());
           return;
         }
-        bool rewriteAddrs = true;
         std::variant<service::Address, RouterID> to;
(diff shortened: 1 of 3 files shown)


majestrate:
ideally i'd like to retag `;_;`

llarp/handlers/tun.cpp
```diff
@@ -1074,7 +1074,12 @@ namespace llarp
           }
```
jagerman, line 1078: I think there should be a `break;` two lines above this (but github won't let me comment there), otherwise we'd get `!allow` if the first one allows and the second one doesn't.

jagerman (approved):
So shall we retag, and bump the deb release to 0.9.2-2, or tag a new one as 0.9.2?

PR #1651, opened 2021-05-18
add --help and -h flags to lokinet-bootstrap to show help message

### LINKED ISSUE #1650: lokinet-bootstrap needs --help
lokinet-bootstrap appears undocumented and doens't have a `--help` which is rather unfriendly:
```
$ lokinet-bootstrap --help
failed to fetch '--help' HTTP 0
```
It really needs one (or just some output if the input doesn't look like a URL or the mainnet/lokinet/testnet magic values).  Here's the output from the old script which would probably work fine as is (just with the bash variables replaced):

```
Usage: $0 [URL [DEST]] -- download bootstrap file from URL (default: lokinet) and save to DEST (default: $default_dest).
URL can be a full URL, or else 'lokinet' or 'testnet' to use the default lokinet/testnet seed URL.  'mainnet' can be used
as an alias for 'lokinet'.
```

majestrate:
fixes #1650

### DIFF
--- a/daemon/lokinet-bootstrap.cpp
+++ b/daemon/lokinet-bootstrap.cpp
@@ -9,6 +9,7 @@
 #include <iostream>
 
 #include <unordered_map>
+#include <unordered_set>
 
 #ifndef _WIN32
 #include <openssl/x509.h>
@@ -28,6 +29,24 @@ namespace
     std::cout << msg << std::endl;
     return 1;
   }
+
+  int
+  print_help(std::string exe)
+  {
+    std::cout << R"(Lokinet bootstrap.signed fetchy program thing
+
+Downloads the initial bootstrap.signed for lokinet into a local file from a default or user defined server via the reachable network.
+
+Usage: )" << exe
+              << R"( [bootstrap_url [output_file]]
+
+bootstrap_url can be a named value in which we will use a predefined url that
+holds a bootstrap file
+
+)";
+    return 0;
+  }
+
 }  // namespace
 
 int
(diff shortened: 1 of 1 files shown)


daemon/lokinet-bootstrap.cpp
```diff
@@ -28,6 +29,31 @@ namespace
...
+
+  template <typename Map_t>
+  int
+  print_help(std::string exe, const Map_t& urls)
+  {
+    std::cout << "Lokinet bootstrap.signed fetchy program thing" << std::endl;
+    std::cout << "Downloads the initial bootstrap.signed for lokinet into a local file from a "
+                 "default or user defined server via the reachable network"
+              << std::endl
+              << std::endl;
```
jagerman: Should just use `\n`'s, `<< std::endl` is... misunderstood.  It's equivalent to `<< "\n" << std::flush;` which forces a flush and is almost always unnecessary (e.g. stdout is already flush-on-newline when output is a tty).

You maybe could also just literal string the whole thing to make it a bit nicer:

```C++
  std::cout << R"(Lokinet bootstrap.signed fetchy program thing

Downloads the initial bootstrap.signed for lokinet into a local file from a default or user
defined server via the reachable network.

Usage: )" << exe << R"( [bootstrap_url [output_file]]

bootstrap_url can be a named value in which we will use a predefined url that
holds a bootstrap file

)";
```
etc.

jagerman: (People commonly think `std::endl` means "do the right thing on (windows|other archaic systems)" but it doesn't; that magic happens with "\n" in the first place.  `std::endl` just says "print \n and then force a flush").

daemon/lokinet-bootstrap.cpp
```diff
@@ -28,6 +29,24 @@ namespace
...
+  print_help(std::string exe)
+  {
+    std::cout << R"(Lokinet bootstrap.signed fetchy program thing
+
+Downloads the initial bootstrap.signed for lokinet into a local file from a default or user defined server via the reachable network.
+
+Usage: )" << exe
+              << R"( [bootstrap_url [output_file]]
+
+bootstrap_url can be a named value in which we will use a predefined url that
```
tewinget: can't hurt to enumerate the list of options here

jagerman: Huh.  I'm pretty sure there *was* an enumeration of the values here when I looked at it yesterday.

jagerman (approved):

PR #1656, opened 2021-05-30
DNS fixes and features

### LINKED ISSUE #1655: Setting an empty upstream doesn't disable non-Loki DNS
**Describe the bug**
Setting `upstream=` does not disable DNS resolution for clearnet domains.

**To Reproduce**
1. `lokinet -g`
2. `lokinet-bootstrap`
3. Set `upstream=` in lokinet.ini
4. Start lokinet
5. `dig @127.3.2.1 google.com`

**Device and Operating system:**
 - OS: Linux (Fedora 33)
-  Device: PC
 - Lokinet Version number or Git commit hash: cbb78b52800f3f6e2f9ee0c13e5caa65b9112801

majestrate:
* make it so that we don't set up unbound resolver when we have no resolvers provided by config
* clean up dns codepath and make it use llarp::SockAddr instead of llarp::IpAddress

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -748,16 +748,15 @@ namespace llarp
             m_bind.setPort(53);
         });
 
-    conf.defineOption<std::string>(
+    conf.defineOption<fs::path>(
         "dns",
         "add-hosts",
         ClientOnly,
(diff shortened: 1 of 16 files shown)


majestrate:
added the ability to add multiple dns hosts file via `add-hosts` in dns section

tewinget (approved):

PR #1658, opened 2021-06-02
Network stability fixes

majestrate:
replaces #1653

### DIFF
--- a/llarp/messages/relay_commit.cpp
+++ b/llarp/messages/relay_commit.cpp
@@ -204,6 +204,7 @@ namespace llarp
     static void
     OnForwardLRCMResult(
         AbstractRouter* router,
+        std::shared_ptr<path::TransitHop> path,
         const PathID_t pathid,
         const RouterID nextHop,
         const SharedSecret pathKey,
@@ -236,9 +237,8 @@ namespace llarp
           std::abort();
           break;
       }
-
-      router->QueueWork([router, pathid, nextHop, pathKey, status] {
-        LR_StatusMessage::CreateAndSend(router, pathid, nextHop, pathKey, status);
+      router->QueueWork([router, path, pathid, nextHop, pathKey, status] {
+        LR_StatusMessage::CreateAndSend(router, path, pathid, nextHop, pathKey, status);
       });
     }
 
@@ -251,6 +251,7 @@ namespace llarp
         llarp::LogError("duplicate transit hop ", self->hop->info);
         LR_StatusMessage::CreateAndSend(
             self->context->Router(),
+            self->hop,
             self->hop->info.rxID,
             self->hop->info.downstream,
             self->hop->pathKey,
@@ -269,6 +270,7 @@ namespace llarp
           llarp::LogError("client path build hit limit ", *self->fromAddr);
           OnForwardLRCMResult(
               self->context->Router(),
+              self->hop,
               self->hop->info.rxID,
               self->hop->info.downstream,
               self->hop->pathKey,
@@ -288,6 +290,7 @@ namespace llarp
             "not allowed, dropping build request on the floor");
         OnForwardLRCMResult(
             self->context->Router(),
+            self->hop,
             self->hop->info.rxID,
             self->hop->info.downstream,
             self->hop->pathKey,
@@ -308,6 +311,7 @@ namespace llarp
       auto func = std::bind(
           &OnForwardLRCMResult,
           self->context->Router(),
+          self->hop,
           self->hop->info.rxID,
           self->hop->info.downstream,
           self->hop->pathKey,
@@ -338,6 +342,7 @@ namespace llarp
 
       if (!LR_StatusMessage::CreateAndSend(
               self->context->Router(),
+              self->hop,
               self->hop->info.rxID,
               self->hop->info.downstream,
               self->hop->pathKey,
--- a/llarp/messages/relay_commit.cpp
+++ b/llarp/messages/relay_commit.cpp
@@ -308,14 +308,15 @@ namespace llarp
       self->context->PutTransitHop(self->hop);
       // forward to next hop
       using std::placeholders::_1;
-      auto func = std::bind(
-          &OnForwardLRCMResult,
-          self->context->Router(),
-          self->hop,
-          self->hop->info.rxID,
-          self->hop->info.downstream,
-          self->hop->pathKey,
-          _1);
+      auto func = [self](auto status) {
+        OnForwardLRCMResult(
+            self->context->Router(),
+            self->hop,
+            self->hop->info.rxID,
+            self->hop->info.downstream,
+            self->hop->pathKey,
+            status);
+      };
       self->context->ForwardLRCM(self->hop->info.upstream, self->frames, func);
       self->hop = nullptr;
     }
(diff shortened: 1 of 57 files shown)


llarp/messages/relay_commit.cpp
```diff
@@ -308,6 +311,7 @@ namespace llarp
       auto func = std::bind(
```
tewinget: may as well make this a lambda while we're here.

llarp/path/path.cpp
```diff
@@ -678,28 +708,48 @@ namespace llarp
...
+        return mean;
+      for (const auto& samp : samps)
+        mean += samp;
+      return mean / samps.size();
     }
 
+    constexpr auto MaxLatencySamples = 8;
+
     bool
     Path::HandlePathLatencyMessage(const routing::PathLatencyMessage& msg, AbstractRouter* r)
```
tewinget, line 734: now always returns true, change to void?

llarp/path/path_context.cpp
```diff
@@ -172,6 +173,46 @@ namespace llarp
           });
     }
 
+    std::optional<std::weak_ptr<TransitHop>>
+    PathContext::TransitHopByInfo(const TransitHopInfo& info)
+    {
+      // this is ugly as sin
```
tewinget, line 179: you're goddamn right it is

llarp/path/path_context.cpp
```diff
@@ -95,8 +95,9 @@ namespace llarp
         typename Map_t,
         typename Key_t,
         typename CheckValue_t,
-        typename GetFunc_t>
-    HopHandler_ptr
+        typename GetFunc_t,
+        typename Return_ptr = HopHandler_ptr>
+    Return_ptr
     MapGet(Map_t& map, const Key_t& k, CheckValue_t check, GetFunc_t get)
```
jagerman, line 101:
```C++
template <typename Lock, typename Map, typename Key, typename CheckValue, typename GetFunc>
auto MapGet(Map& map, const Key& k, CheckValue check, GetFunc get)
    -> decltype(get(std::declval<const Map::mapped_type&>()) {
    ...
}
```

llarp/path/path_context.cpp
```diff
@@ -172,6 +173,46 @@ namespace llarp
...
+          SyncTransitMap_t::Lock_t,
+          decltype(m_TransitPaths),
+          PathID_t,
+          std::function<bool(const std::shared_ptr<TransitHop>&)>,
+          std::function<TransitHop*(const std::shared_ptr<TransitHop>&)>,
+          TransitHop*>(
+          m_TransitPaths,
+          info.txID,
+          [info](const auto& hop) -> bool { return hop->info == info; },
+          [](const auto& hop) -> TransitHop* { return hop.get(); });
```
jagerman, line 190: ... and then this can become:
```C++
auto own = MapGet<SyncTransitMap_t::Lock_t>(
    m_TransitPaths,
    info.txID,
    [info](const auto& hop) { return hop->info == info; },
    [](const auto& hop) { return hop.get(); });
 ```

File: llarp/path/pathbuilder.cpp [hunk 6]
```diff
@@ -192,6 +194,9 @@ namespace llarp
     void Builder::Tick(llarp_time_t)
     {
       const auto now = llarp::time_now_ms();
+
+      m_router->pathBuildLimiter().Decay(now);
```
tewinget, line 198: why would this not be in `router::Tick()`?  Simply unnecessary to decay it that often, I assume?

llarp/router/router.cpp
```diff
@@ -1402,8 +1407,10 @@ namespace llarp
   void
   Router::QueueWork(std::function<void(void)> func)
   {
-    _loop->call_soon(func);
-    // m_lmq->job(std::move(func));
+    if (m_isServiceNode)
```
tewinget, line 1410: could stand to have a comment here clarifying the two cases and why the decision to handle client/relay differently.

llarp/service/endpoint.hpp
```diff
@@ -284,8 +285,8 @@ namespace llarp
       bool
       WantsOutboundSession(const Address&) const override;
 
-      void
-      MarkAddressOutbound(const Address&) override;
+      /// this MUST be called if you want to call EnsurePathTo on the given address
+      void MarkAddressOutbound(AddressVariant_t) override;
```
tewinget, line 289: guessing clang-format is not gonna like this line

llarp/service/endpoint_util.cpp
```diff
@@ -37,18 +37,22 @@ namespace llarp
     void
     EndpointUtil::ExpirePendingTx(llarp_time_t now, PendingLookups& lookups)
     {
+      std::vector<std::unique_ptr<IServiceLookup>> timedout;
```
tewinget, line 40: nitpick, but probably call this `timed_out` so it doesn't look like the noun `timeout`

llarp/service/intro.hpp
```diff
@@ -77,6 +77,16 @@ namespace llarp
     {
       return i.print(out, -1, -1);
     }
+
+    /// comparitor for introset timestamp
```
tewinget: `comparator`

llarp/service/outbound_context.cpp
```diff
@@ -264,7 +289,7 @@ namespace llarp
             path->Endpoint(),
             relayOrder,
             m_Endpoint->GenTXID(),
-            5s);
+            5s + (2 * path->intro.latency));
```
tewinget: changing this anyway, may as well demystify the `5s` constant

llarp/service/outbound_context.cpp
```diff
@@ -346,40 +368,48 @@ namespace llarp
...
         const auto path = GetPathByRouter(remoteIntro.router);
         if (not path)
         {
           LogWarn(Name(), " ready but no path to ", remoteIntro.router, " ???");
           return true;
         }
-        m_ReadyHook(this);
-        m_ReadyHook = nullptr;
+        for (const auto& hook : m_ReadyHooks)
+          hook(this);
```
tewinget, line 380: should these calls be sent to the event loop thread?

llarp/service/outbound_context.cpp
```diff
@@ -422,12 +458,24 @@ namespace llarp
       m_BadIntros[intro] = now;
     }
 
+    bool
```
tewinget, line 461: this and the function after appear identical; one is probably wrong

llarp/service/sendcontext.cpp
```diff
@@ -26,13 +26,15 @@ namespace llarp
     bool
     SendContext::Send(std::shared_ptr<ProtocolFrame> msg, path::Path_ptr path)
     {
+      if (not path->IsReady())
+        return false;
       if (m_SendQueue.empty() or m_SendQueue.full())
```
tewinget: `if empty` makes sense here, as that means we don't currently have a flush queued.  `if full`, however, makes little sense with the below change from `call now` to `call soon`.

majestrate:
it is already in the event loop thread

llarp/path/path_context.cpp
tewinget: made this its own separate issue; here there be dragons

tewinget (approved):

PR #1659, opened 2021-06-04
initial lokinet router testing

majestrate:
* report via rpc to oxen core connection stats on success and failure
* connect to random service node by pubkey every 5 seconds for testing

### DIFF
--- a/llarp/rpc/lokid_rpc_client.cpp
+++ b/llarp/rpc/lokid_rpc_client.cpp
@@ -235,10 +235,14 @@ namespace llarp
       // inform router about the new list
       if (auto router = m_Router.lock())
       {
-        router->loop()->call([this, activeNodeList, nonActiveNodeList, keymap]() {
-          m_KeyMap = keymap;
-          if (auto router = m_Router.lock())
-            router->SetRouterWhitelist(activeNodeList, nonActiveNodeList);
+        auto& loop = router->loop();
+        loop->call([this,
+                active=std::move(activeNodeList),
+                inactive=std::move(nonActiveNodeList),
+                keymap=std::move(keymap),
+                router=std::move(router)]() mutable {
+          m_KeyMap = std::move(keymap);
+          router->SetRouterWhitelist(active, inactive);
         });
       }
       else
--- a/llarp/rpc/lokid_rpc_client.cpp
+++ b/llarp/rpc/lokid_rpc_client.cpp
@@ -204,8 +204,8 @@ namespace llarp
             PubKey pk;
             if (rid.FromHex(ed_itr->get<std::string>()) and pk.FromHex(svc_itr->get<std::string>()))
             {
-              nodeList.emplace_back(std::move(rid));
               keymap[rid] = pk;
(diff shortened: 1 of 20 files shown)


llarp/rpc/lokid_rpc_client.hpp
```diff
@@ -42,6 +42,10 @@ namespace llarp
           dht::Key_t namehash,
           std::function<void(std::optional<service::EncryptedName>)> resultHandler);
 
+      /// record that if connected to a router successfully
```
tewinget: typo in comment

jagerman:
Some suggestions from recently doing this on the storage server side:

- random testing is fairly erratic because some nodes can go a very long time without getting tested.  Rather in SS we use a shuffled list and each time we test a node we pull off the top, and regenerate the list when it empties.
- 5s seems a bit faster than necessary (and we can make it a bit slower especially if we go through the list randomly, rather than just randomly selected each time).
- oxend expects frequent retesting of failing nodes (and needs it, as it considers a node to be no longer failing if it doesn't get an update within 5 minutes).  SS's (recently reworked) logic on this is:
  - keep a list of failing nodes, along with their scheduled re-test time and number of previous failures.
  - on a test tick, remove one random node from the shuffled list (skipping any that are already on the fail list), plus nodes on the fail list due to be tested.
  - when a node fails, add it onto the failing list, with a re-test time of:
    - 10s + R from now if it was not previously failing
    - 10s*N + R from now if it was previously failing with N previous failures
    - R here is some random fuzz (so that we space out re-tests), drawn from a `inline static thread_local std::normal_distribution<float> TESTING_INTERVAL{10.0, 3.0};`
- tick the re-test queue faster (SS uses every 100ms); most of the time it will do nothing, but this gives more spreading out of test times.
- keep track of when the next "general" (i.e. random working snode) test is and, after each general test, reset it to `now + TESTING_INTERVAL(rng())` (so that we also space out random testing, but keep them at 10s apart on average).

You can probably actually lift storage server's reachability_testing.cpp/.h code more or less as is (really the only thing SS-specific in it is the `swarm` object—used to access the current list of active nodes from within storage server—and the logger).

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -190,9 +197,16 @@ namespace llarp
...
+            const auto svc_itr = j_itr->find("service_node_pubkey");
+            if (svc_itr == j_itr->end() or not svc_itr->is_string())
+              continue;
             RouterID rid;
-            if (rid.FromHex(ed_itr->get<std::string>()))
+            PubKey pk;
+            if (rid.FromHex(ed_itr->get<std::string>()) and pk.FromHex(svc_itr->get<std::string>()))
+            {
               nodeList.emplace_back(std::move(rid));
+              keymap[rid] = pk;
```
majestrate, line 218: this here, move above `nodeList.emplace_back`

jagerman (approved):

majestrate:
ah yes, the monthly patch tuesday github outage breaks CI... amazin.

PR #1663, opened 2021-06-09
Last minute fixes

majestrate:
* properly do latency testing FEC, makes it so that paths don't die for "no real reason"
* flush and pump when we send to the network instead of just flushing, this makes it so that the packets actually go out.

### DIFF
--- a/llarp/path/path.cpp
+++ b/llarp/path/path.cpp
@@ -458,7 +458,7 @@ namespace llarp
         {
           SendLatencyMessage(r);
           // latency test FEC
-          r->loop()->call_later(1s, [self = shared_from_this(), r]() {
+          r->loop()->call_later(2s, [self = shared_from_this(), r]() {
             if (self->m_LastLatencyTestID)
               self->SendLatencyMessage(r);
           });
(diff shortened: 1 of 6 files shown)


llarp/path/path.cpp
```diff
@@ -439,15 +456,12 @@ namespace llarp
...
-          m_LastLatencyTestID = latency.T;
-          m_LastLatencyTestTime = now;
-          SendRoutingMessage(latency, r);
-          FlushUpstream(r);
-          // reset ID so we don't mark ourself as dead if we drop a latency sample
-          r->loop()->call_later(
-              1s, [self = shared_from_this()]() { self->m_LastLatencyTestID = 0; });
+          SendLatencyMessage(r);
+          // latency test FEC
+          r->loop()->call_later(1s, [self = shared_from_this(), r]() {
```
jagerman: Perhaps this should be a little longer, say 2s?  1s is rare but possible for a path (and might be more possible if someone runs in paranoid 8-hop path mode).  2s seems rare enough to be a fine cutoff.

majestrate: yea 2s works

jagerman (approved):

PR #1669, opened 2021-06-14
[bugfix] make inbound convotags reply on the correct path

majestrate:
inbound convotags were not replying on the right path, this fixes this.
note: i really want to redo the convotag layer's design.

### DIFF
--- a/llarp/service/outbound_context.cpp
+++ b/llarp/service/outbound_context.cpp
@@ -314,11 +314,6 @@ namespace llarp
       obj["currentRemoteIntroset"] = currentIntroSet.ExtractStatus();
       obj["nextIntro"] = m_NextIntro.ExtractStatus();
       obj["readyToSend"] = ReadyToSend();
-      std::transform(
-          m_BadIntros.begin(),
-          m_BadIntros.end(),
-          std::back_inserter(obj["badIntros"]),
-          [](const auto& item) -> util::StatusObject { return item.first.ExtractStatus(); });
       return obj;
     }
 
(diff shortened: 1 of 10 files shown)


llarp/service/outbound_context.cpp
```diff
@@ -351,22 +350,36 @@ namespace llarp
...
-      while (itr != m_BadIntros.end())
-      {
-        if (now > itr->second && now - itr->second > path::default_lifetime)
-          itr = m_BadIntros.erase(itr);
-        else
-          ++itr;
-      }
 
       if (ReadyToSend() and not m_ReadyHooks.empty())
       {
```
tewinget, line 386: `ReadyToSend()` checks for `GetPathByRouter()` success now, so the check after this is redundant.

llarp/service/protocol.cpp
```diff
@@ -367,9 +367,15 @@ namespace llarp
                 AuthResult result) {
               if (result.code == AuthResultCode::eAuthAccepted)
               {
-                handler->PutSenderFor(msg->tag, msg->sender, true);
-                handler->PutIntroFor(msg->tag, msg->introReply);
-                handler->PutReplyIntroFor(msg->tag, fromIntro);
+                if (handler->WantsOutboundSession(msg->sender.Addr()))
```
tewinget, line 370: Could be simplified to:
```
bool x = handler->WantsOutboundSession(...);
handler->PutSenderFor(..., x);
```

tewinget (approved):

PR #1672, opened 2021-06-17
[feature] optionally keep inbound convos mapped to same IP on restart 

majestrate:
* add `[network]:persist-addrmap-file` config option to store ip address mappings to disk
* restore any inbound convo's address/ip mappings if configured to do so


majestrate:
[needs squash]

tewinget (approved):

PR #1682, opened 2021-06-25
[linux] route blackholing

majestrate:
* add / remove route blackhole so we dont leak if we crash, currently linux only.


tewinget (approved):
seems fine, pending confirmation that it works as intended.

PR #1713, opened 2021-08-10
Remove unused defaults.hpp

jagerman:
The definitions in here aren't actually used anywhere anymore, so just drop it.


jagerman:
Also remove the obsolete lokinet-bootstrap script.

majestrate (approved):

PR #1714, opened 2021-08-11
Remove unneeded split(str, char) method

jagerman:
4f1bd14d3c64bfec63f8574cc88790b390956d14 fixed a bug that also surfaced on Fedora 34 (I think it's actually a gcc-11 thing) but on closer inspection this whole function is just duplicating what you can do with the somewhat more generic `split` and/or `split_any` defined just below it (which takes a string_view instead of char delimiter).

Simplify the code by just deleting it and changing everything using it to pass "c" instead of 'c'.

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -798,7 +798,7 @@ namespace llarp
     {
       info.interface = std::string{name};
 
-      std::vector<std::string_view> splits = split(value, ',');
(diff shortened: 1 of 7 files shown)


jagerman:
I'm actually rather curious why the current code on `stable` doesn't segfault: it seems like it would sometimes be accessing string element std::string::npos.

majestrate (approved):

PR #1715, opened 2021-08-11
Fix default upstream DNS not working

jagerman:
The default upstream DNS was being set to 1.1.1.1:0, which doesn't work.  This fixes it to also set the port so that default upstream resolution (i.e. with an empty config) works again.

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -711,6 +711,8 @@ namespace llarp
     // Default, but if we get any upstream (including upstream=, i.e. empty string) we clear it
     constexpr Default DefaultUpstreamDNS{"1.1.1.1"};
     m_upstreamDNS.emplace_back(DefaultUpstreamDNS.val);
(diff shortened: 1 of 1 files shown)


llarp/config/config.cpp
```diff
@@ -711,6 +711,8 @@ namespace llarp
     // Default, but if we get any upstream (including upstream=, i.e. empty string) we clear it
     constexpr Default DefaultUpstreamDNS{"1.1.1.1"};
     m_upstreamDNS.emplace_back(DefaultUpstreamDNS.val);
+    if (!m_upstreamDNS.back().getPort())
+        m_upstreamDNS.back().setPort(53);
```
jagerman, line 715: Did it this way rather than putting :53 in the default value so that the default shown in the config remains 1.1.1.1 rather than 1.1.1.1:53

majestrate (approved):

PR #1721, opened 2021-08-27
initial routing table platform bits refactor

majestrate:
move routing table manipulation into vpn platform


llarp/ev/vpn.hpp
```diff
@@ -59,6 +61,38 @@ namespace llarp::vpn
...
+
+    virtual void
+    DelRoute(IPVariant_t ip, IPVariant_t gateway) = 0;
+
+    virtual void
+    AddDefaultRouteViaInterface(std::string ifname) = 0;
+
+    virtual void
+    DelDefaultRouteViaInterface(std::string ifname) = 0;
+
```
jagerman, line 85: These should take an optional IP range argument for when running in specific-range exit mode.  Perhaps it could take an optional set of ranges; if nullopt we make it the default, otherwise we apply just the ranges in the set.

llarp/ev/vpn.hpp
```diff
@@ -59,6 +61,38 @@ namespace llarp::vpn
...
+    virtual std::vector<IPVariant_t>
+    GetGatewaysNotOnInterface(std::string ifname) = 0;
+
+    virtual void
+    AddBlackhole() = 0;
+
+    virtual void
+    DelBlackhole() = 0;
```
jagerman: Everything except linux is (I think) going to have empty implementations here, so probably better to just make these default empty rather than abstract.

llarp/ev/vpn.hpp
jagerman: ... or a new one for adding specific routes rather when in non-default mode.

llarp/ev/vpn.hpp
majestrate: done

llarp/ev/vpn.hpp
majestrate: added new ones for range via interface

jagerman (approved):
Missing the mac version, but that comes in the macOS PR.

PR #1723, opened 2021-08-30
limit RC gossip to 20 peers max

majestrate:
only gossip RCs to 20 peers at a time maximum.
this should reduce the load on service nodes when gossiping RCs.

### DIFF
--- a/llarp/router/rc_gossiper.cpp
+++ b/llarp/router/rc_gossiper.cpp
@@ -76,17 +76,38 @@ namespace llarp
     DHTImmediateMessage gossip;
     gossip.msgs.emplace_back(new dht::GotRouterMessage(dht::Key_t{}, 0, {rc}, false));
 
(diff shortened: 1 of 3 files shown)


llarp/router/rc_gossiper.cpp
```diff
@@ -76,17 +76,38 @@ namespace llarp
     DHTImmediateMessage gossip;
     gossip.msgs.emplace_back(new dht::GotRouterMessage(dht::Key_t{}, 0, {rc}, false));
 
-    // send it to everyone
+    constexpr size_t MaxGossipPeers = 20;
```
jagerman: Should be a bit more prominent, at least in the header for this.

majestrate: moved to header

jagerman (approved):

PR #1726, opened 2021-09-01
dns refactor

majestrate:
when we reset the unbound context we need to cancel any pending queries so the unbound context cleanly tears itself down, on windows.

on everything else use a pollable fd.

### DIFF
--- a/llarp/dns/unbound_resolver.cpp
+++ b/llarp/dns/unbound_resolver.cpp
@@ -25,11 +25,18 @@ namespace llarp::dns
   UnboundResolver::Reset()
   {
     started = false;
-    if (runner)
+#ifdef _WIN32
+    if (runner.joinable())
     {
-      runner->join();
-      runner.reset();
+      runner.join();
     }
+#else
+    if (udp)
+    {
+      udp->close();
+    }
+    udp.reset();
+#endif
     if (unboundContext)
     {
       ub_ctx_delete(unboundContext);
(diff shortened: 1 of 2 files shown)


llarp/dns/unbound_resolver.cpp
```diff
@@ -24,9 +25,18 @@ namespace llarp::dns
...
   {
-    started = false;
     if (runner)
     {
+      // cancel all pending lookups
+      for (auto id : pending_resolve_jobs)
+      {
+        if (unboundContext)
+          ub_cancel(unboundContext, id);
+      }
```
jagerman: This shouldn't be necessary: `ub_ctx_delete` already cancels all in-progress async queries.  (We should still clear the pending_resolve_jobs, though).

llarp/dns/unbound_resolver.cpp
```diff
@@ -54,6 +64,9 @@ namespace llarp::dns
     if (not this_ptr)
       return;  // resolver is gone, so we don't reply.
 
+    // remove from pending jobs
+    this_ptr->pending_resolve_jobs.erase(lookup->id);
+
```
jagerman: ~~Was this just leaking before?~~

jagerman: Oh, never mind, this is a new field.

jagerman:
Confirmed working fine on macos (when pulled into #1688).

jagerman (approved):

PR #1728, opened 2021-09-01
Unbound callbacks also need arguments reversed

jagerman:
PR #1725 reversed argument orders but UnboundResolver was still using
(from,to) ordering in its callbacks, which leaked through to make a
wrong order in our reply function (which simply forwards arguments).

This fixes that bug by making UnboundResolver callback argument order
consistent (i.e. using to, from) with the PacketHandler argument order.

Also renames the arguments in the `using` for the callbacks to make it a little more obvious which argument is which.

### DIFF
--- a/llarp/dns/server.cpp
+++ b/llarp/dns/server.cpp
@@ -58,7 +58,7 @@ namespace llarp::dns
       return true;
 
     auto failFunc = [self = weak_from_this()](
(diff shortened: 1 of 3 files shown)


majestrate:
does this work?

jagerman:
On mac, yes.  Haven't tried on android.

majestrate (approved):

Issue #1734, opened 2021-09-10
Create rpc call for get_stats


darcys22:
For the GUI frontend we are currently calling a dump_stats endpoint which sends too much information for the frontend.

Design a smaller endpoint to send only relevant information for the frontend to query

majestrate:
for calculating global in/out bandwidth you can access `llarp::AbstractRouter`'s `ILinkManager` iterate all links, iterate all peers on each link and sum up the total current bandwidth use from each session. you'd need to add a method to `ILinkSession` for getting a session's current bandwidth use.

PR #1742, opened 2021-09-17
RPC call for summary get_status

darcys22:


darcys22:
Fixes #1734 

This was mostly driven by the GUI usage of `/status` in the electron codebase:
https://github.com/Bilb/lokinet-gui/blob/dev/src/ipc/ipcRenderer.ts#L297-L404

Essentially filtering the current status endpoint for the following:
```
      return util::StatusObject{
          {"running", true},
          {"exitAuthCode", auth},
          {"exitNode", exit_node},
          {"lokiAddress", loki_address},
          {"numPathsBuilt", paths},
          {"numPeersConnected", peers},
          {"numRoutersKnown", _nodedb->NumLoaded()},
          {"ratio", ratio},
          {"txRate", tx_rate},
          {"rxRate", rx_rate},
      };
```
The most noticable difference being that `txRate` and `rxRate` is returned where the GUI then takes this to make an upload and download rate.

darcys22:
@bilb thoughts

Bilb (approved):

Bilb:
LGTM

llarp/router/router.cpp
```diff
@@ -97,6 +97,105 @@ namespace llarp
...
+        }
+      }
+
+      std::string loki_address = services["default"]["identity"];
+
+      std::string exit_node;
+      auto exit_map = services["default"]["exitMap"];
+      if (!exit_map.is_null())
+      {
+        // exitMap should be of length 1 only, but it's an object with keys an IP (not as string)
```
majestrate: this assumption of there being only one entry is not correct, we use a map because there are more than one.

llarp/router/router.cpp
```diff
@@ -97,6 +97,105 @@ namespace llarp
...
+        for (auto it = exit_map.begin(); it != exit_map.end(); it++)
+          exit_node = it.value();
+      }
+
+      auto auth_codes = services["default"]["authCodes"];
+      std::string auth;
+      if (!auth_codes.is_null())
+      {
+        for (const auto& loki_exit : auth_codes)
+          auth = loki_exit;
```
majestrate: ditto for auth codes

llarp/router/router.cpp
```diff
@@ -97,6 +97,105 @@ namespace llarp
...
+        {
+          if (builder["length"].is_number())
+            paths += builder["length"].get<int>();
+          if (builder["success"].is_number())
+            success += builder["success"].get<int>();
+          if (builder["attempts"].is_number())
+            attempts += builder["attempts"].get<int>();
+        }
+      }
+      float ratio = ceil((success * 100) / (attempts + 1));
```
majestrate: `float` -> `double`

majestrate: also this calculation is wrong

llarp/rpc/rpc_server.cpp
```diff
@@ -133,6 +133,22 @@ namespace llarp::rpc
                 defer.reply(data);
               });
             })
+        .add_request_command(
+            "get_status",
+            [&](oxenmq::Message& msg) {
+              m_Router->loop()->call([defer = msg.send_later(), r = m_Router]() {
+                std::string data;
+                if (r->IsRunning())
```
majestrate: redundant check for `IsRunning`

this can be simplified to 

```
defer.reply(CreateJSONResponse(r->ExtractSummaryStatus()));
```

llarp/router/router.cpp
```diff
@@ -97,6 +97,84 @@ namespace llarp
     }
   }
 
+  util::StatusObject
+  Router::ExtractSummaryStatus() const
+  {
+    if (_running)
```
majestrate: kind of feel like a `if(not _running) return ...` is better .

llarp/router/router.cpp
```diff
@@ -97,6 +97,84 @@ namespace llarp
...
+    {
+      auto services = _hiddenServiceContext.ExtractStatus();
+      auto link_types = _linkManager.ExtractStatus();
+
+      int tx_rate = 0;
+      int rx_rate = 0;
+      int peers = 0;
+      for (const auto& links : link_types)
+      {
+        if (links == nlohmann::json::array())
```
jagerman: ~~I think `if (links.empty())` would work here (and 4 lines down).~~

llarp/router/router.cpp
```diff
@@ -97,6 +97,84 @@ namespace llarp
...
+          continue;
+        for (const auto& link : links)
+        {
+          if (links == nlohmann::json::array())
+            continue;
+          const auto& established_peers = link["sessions"]["established"];
+          for (const auto& peer : established_peers)
```
jagerman: Can combine this easily:
```c++
for (const auto& peer : link["sessions"]["established"])
{ ... }
```

llarp/router/router.cpp
jagerman: Actually, this check doesn't look needed at all: if it's empty then the next for loop won't do anything anyway.

llarp/router/router.cpp
```diff
@@ -97,6 +97,84 @@ namespace llarp
...
+      int rx_rate = 0;
+      int peers = 0;
+      for (const auto& links : link_types)
+      {
+        if (links == nlohmann::json::array())
+          continue;
+        for (const auto& link : links)
+        {
+          if (links == nlohmann::json::array())
+            continue;
```
jagerman: This doesn't look right...

llarp/router/router.cpp
```diff
@@ -97,6 +97,84 @@ namespace llarp
...
+            peers++;
+          }
+        }
+      }
+
+      // Compute all stats on all path builders on the default endpoint
+      // Merge snodeSessions, remoteSessions and default into a single array
+      std::vector<nlohmann::json> builders;
+      auto snode_sessions = services["default"]["snodeSessions"];
+      for (const auto& session : snode_sessions)
```
jagerman, line 133: This is going to copy; either get as a reference, or just drop the temporary and use it directly in the `for` loop

llarp/router/router.cpp
```diff
@@ -97,6 +97,84 @@ namespace llarp
...
+        builders.push_back(session["buildStats"]);
+
+      auto remote_sessions = services["default"]["remoteSessions"];
+      for (const auto& session : remote_sessions)
```
jagerman, line 137: And likewise here.

llarp/router/router.cpp
```diff
@@ -97,6 +97,84 @@ namespace llarp
...
+        if (links == nlohmann::json::array())
+          continue;
+        for (const auto& link : links)
+        {
+          if (links == nlohmann::json::array())
+            continue;
+          const auto& established_peers = link["sessions"]["established"];
+          for (const auto& peer : established_peers)
+          {
+            tx_rate += peer["tx"].get<int>();
```
majestrate: `uint64_t` not `int` as it will easily overflow

llarp/router/router.cpp
```diff
@@ -97,6 +97,77 @@ namespace llarp
...
+      if (builder.is_null())
+        continue;
+      if (builder["length"].is_number())
+        paths += builder["length"].get<uint64_t>();
+      if (builder["success"].is_number())
+        success += builder["success"].get<uint64_t>();
+      if (builder["attempts"].is_number())
+        attempts += builder["attempts"].get<uint64_t>();
+    }
+    double ratio = static_cast<double>(success) / (attempts + 1);
```
majestrate, line 155: `const double`

llarp/router/router.cpp
```diff
@@ -97,6 +97,77 @@ namespace llarp
...
+    }
+
+    // Compute all stats on all path builders on the default endpoint
+    // Merge snodeSessions, remoteSessions and default into a single array
+    std::vector<nlohmann::json> builders;
+    auto snode_sessions = services["default"]["snodeSessions"];
+    for (const auto& session : snode_sessions)
+      builders.push_back(session["buildStats"]);
+
+    auto remote_sessions = services["default"]["remoteSessions"];
```
majestrate, line 134: `const auto`

llarp/router/router.cpp
```diff
@@ -97,6 +97,77 @@ namespace llarp
...
+          rx_rate += peer["rx"].get<uint64_t>();
+          peers++;
+        }
+      }
+    }
+
+    // Compute all stats on all path builders on the default endpoint
+    // Merge snodeSessions, remoteSessions and default into a single array
+    std::vector<nlohmann::json> builders;
+    auto snode_sessions = services["default"]["snodeSessions"];
```
majestrate, line 130: `const auto`

llarp/router/router.cpp
```diff
@@ -97,6 +97,77 @@ namespace llarp
...
   }
 
+  util::StatusObject
+  Router::ExtractSummaryStatus() const
+  {
+    if (!_running)
+      return util::StatusObject{{"running", false}};
+
+    auto services = _hiddenServiceContext.ExtractStatus();
+    auto link_types = _linkManager.ExtractStatus();
```
majestrate, line 107: `const auto` for these 2 as well.

majestrate (approved):
good enough

jagerman:
> Fixes #1734

@darcys22  - for future reference this "fixes #xxx" thing only does the right thing (i.e. automatically closing) if github sees it in the *original* message; it appears to have no effect on subsequent comments.  (I do believe it is okay to edit it in, though).

PR #1748, opened 2021-09-21
liblokinet updates

majestrate:
required for liblokinet wrappers


contrib/mac.sh
```diff
@@ -21,7 +21,7 @@ cmake \
       -DBUILD_PACKAGE=ON \
       -DBUILD_SHARED_LIBS=OFF \
       -DBUILD_TESTING=OFF \
-      -DBUILD_LIBLOKINET=OFF \
+      -DBUILD_LIBLOKINET=ON \
```
jagerman: Probably was only for testing, revert?

external/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ add_ngtcp2_lib()
 
 # cpr configuration.  Ideally we'd just do this via add_subdirectory, but cpr's cmake requires
 # 3.15+, and we target lower than that (and this is fairly simple to build).
-
+if(WITH_BOOTSTRAP)
 if(NOT BUILD_STATIC_DEPS)
```
jagerman: Indent?

contrib/mac.sh
majestrate: fixed

include/lokinet/lokinet_udp.h
```diff
@@ -2,103 +2,119 @@
...
     /// a socket id used to close a lokinet udp socket
     int socket_id;
   };
 
+  /// flow acceptor hook, return 0 success, return nonzero with errno on failure
+  typedef int (*lokinet_udp_flow_filter)(
+      void* /*user*/,
+      const struct lokinet_udp_flowinfo* /* remote address */,
+      void** /* flow-userdata */,
+      int* /* timeout seconds */);
```
jagerman: You can have argument names in a function pointer typedef, so these can all get uncommented.

include/lokinet/lokinet_udp.h
```diff
@@ -2,103 +2,119 @@
...
 
-  struct lokinet_udp_pkt
-  {
-    char remote_addr[256];
-    int remote_port;
-    struct iovec pkt;
-  };
+  /// @brief send on an established flow to remote endpoint
+  ///
+  /// @param flowinfo populated after call on success
```
jagerman: This doc looks wrong

include/lokinet/lokinet_udp.h
majestrate: names added

include/lokinet/lokinet_udp.h
majestrate: it is, fixed.

external/CMakeLists.txt
majestrate: done

llarp/lokinet_shared.cpp
```diff
@@ -523,6 +824,13 @@ extern "C"
     return id;
   }
 
+  char* EXPORT
+  lokinet_hex_to_base32z(const char* hex)
+  {
+    const auto base32z = oxenmq::to_base32z(oxenmq::from_hex(std::string{hex}));
+    return strdup(base32z.c_str());
```
jagerman: I'm thinking this should return nullptr if you give non-hex.  Also we can cut it down to a single allocation instead of 4 (3 strings + strdup) via:

```C++
char* EXPORT
lokinet_hex_to_base32z(const char* hex)
{
  std::string_view hexview{hex};
  if (!oxenmq::is_hex(hexview))
    return nullptr;

  const size_t byte_len = hexview.size() / 2;
  const size_t b32z_len = (byte_len*8+4) / 5; // = ⌈N×8÷5⌉ because 5 bits per 32z char
  auto buf = std::make_unique<char[]>(b32z_len + 1);
  char* end = buf.get() + b32z_len;
  *end = 0; // null terminate
  // Write the bytes into the *end* of the buffer so that when we rewrite the final b32z chars into
  // the buffer we won't overwrite any byte values until after we've consumed them.
  char* bytepos = end - byte_len;
  oxenmq::from_hex(hexview.begin(), hexview.end(), bytepos);
  // In-place conversion into the buffer
  oxenmq::to_base32z(bytepos, end, buf.get());
  return buf.release(); // leak the buffer to the caller
}
```

jagerman (approved):

external/CMakeLists.txt
```diff
@@ -56,26 +56,27 @@ add_ngtcp2_lib()
 
 # cpr configuration.  Ideally we'd just do this via add_subdirectory, but cpr's cmake requires
 # 3.15+, and we target lower than that (and this is fairly simple to build).
+if(WITH_BOOTSTRAP)
+  if(NOT BUILD_STATIC_DEPS)
```
tewinget, line 60: this could be a single `if` line with the line above.

tewinget: nvm misread.

llarp/lokinet_shared.cpp
```diff
@@ -28,6 +59,166 @@ namespace
...
+      }
+    }
+
+    void
+    HandlePacketFrom(AddressVariant_t from, llarp::net::IPPacket pkt)
+    {
+      bool isNewFlow{false};
+      {
+        std::unique_lock lock{m_Access};
+        isNewFlow = m_Flows.count(from) == 0;
```
tewinget: if not `isNewFlow`, call HandlePacket here and return

llarp/lokinet_shared.cpp
```diff
@@ -28,6 +59,166 @@ namespace
...
+          flow_addr.remote_port = ToHost(*srcport).h;
+        }
+        else
+          return;  // invalid data so we bail
+        void* flow_userdata = nullptr;
+        int flow_timeoutseconds{};
+        // got a new flow, let's check if we want it
+        if (m_Filter(m_User, &flow_addr, &flow_userdata, &flow_timeoutseconds))
+          return;
+        AddFlow(from, flow_addr, flow_userdata, flow_timeoutseconds);
```
tewinget: maybe add an optional argument to `AddFlow` so it can take a packet if we intend to send one right away after adding the flow.  This way we don't have to release the mutex acquired in AddFlow, then re-lock it; if we do have to re-lock it we should also re-check if the flow exists.

llarp/service/endpoint.hpp
```diff
@@ -168,6 +170,12 @@ namespace llarp
       void
       HandlePathDied(path::Path_ptr p) override;
 
+      virtual vpn::EgresPacketRouter*
+      EgresPacketRouter()
```
tewinget, line 174: I love this speling of egres.

llarp/vpn/packet_router.hpp
```diff
@@ -17,7 +17,6 @@ namespace llarp::vpn
 
     virtual void AddSubHandler(nuint16_t, PacketHandlerFunc){};
   };
-
```
tewinget, line 20: style nitpick: empty lines between class declarations are nice to have.

llarp/lokinet_shared.cpp
majestrate: added

llarp/lokinet_shared.cpp
majestrate: revamped

PR #1752, opened 2021-09-26
CentOS/8

necro-nemesis:
New branch which builds lokinet-bin and lokinet binaries for CentOS Stream 8. Postinst still requires tweaking for resolver.  Also port 953 may or may not be required vice 53 but remains as it is for Fedora 34 for the present time.


jagerman:
This also needs some changes in .drone.jsonnet -- at a minimum the first three lines to set up the build name & docker container for a centos image.  Build images are from here: https://hub.docker.com/_/centos?tab=tags&page=1&ordering=last_updated -- maybe `centos:centos8`?

necro-nemesis:
I just went by Fedora branch convention distro/release when naming it. Not sure there's any merit adding in also "Stream" to the naming convention as I do no believe there's any other development going forward other than Stream. "CentOS Linux" EOL (end of life) 2021-12-31*. CentOS Stream EOL  2024-05-31 with CentOS Stream 9 reported as "coming soon" and nothing indicated for Linux after EOL.  At them moment there's a crossover with 8 available on both Linux and Stream but that's not to say Lokinet won't function on both to require any distinction between the two current versions. The distinction of Stream in future appears to be superfluous as a result of discontinuation of CentOS Linux.

necro-nemesis:
centos8.4.2105 appears to be CentOS Linux leading me to believe centos8 is the Stream version.

jagerman:
> centos8.4.2105 appears to be CentOS Linux leading me to believe centos8 is the Stream version.

They are the same image; the more specific version is used in case you want a specific 8.x.y version, while the "centos8" should always be the latest centos8 image.

It seems like RedHat doesn't want to bother releasing CentOS Stream on dockerhub at all and uses their own custom proprietary subscription-required service, because IBM.

necro-nemesis:
- Working now with Patch 2 removed for dns listener port patch.
- Current postinst does not enable/start lokinet systemd process. Currently must be started manually.

necro-nemesis:
The CI is building working versions of oxenmq and lokinet. What's being reported here isn't up to date until I make code changes to force a build. The final issues with the build were related to lokinet getting oxenmq-lib dependency from the oxenmq build which is now resolved.

SPECS/lokinet.spec
```diff
@@ -153,6 +150,8 @@ if ! [ -e /etc/loki/lokinet.ini ]; then
 fi
 
 %systemd_post lokinet.service
+systemctl enable lokinet
```
majestrate: `systemctl enable --now lokinet` does what these 2 lines do in 1.

jagerman:
Squashed this and put into a new `centos/8` branch, outside github because github is kind of crappy at doing non-trivial git.

PR #1754, opened 2021-10-01
prevent bizare half open state.

majestrate:
sometimes outbound contexts get stuck in a half open state. this prevents that by making sure we remove ones that happen to hit that case.

### DIFF
--- a/llarp/service/outbound_context.cpp
+++ b/llarp/service/outbound_context.cpp
@@ -64,16 +64,17 @@ namespace llarp
         , currentIntroSet{introset}
 
     {
+      assert(not introset.intros.empty());
       updatingIntroSet = false;
 
       // pick random first intro
-      std::vector<Introduction> intros = introset.intros;
-      if (intros.size() > 1)
+      auto it = introset.intros.begin();
+      if (introset.intros.size() > 1)
       {
-        std::shuffle(intros.begin(), intros.end(), CSRNG{});
+        CSRNG rng{};
+        it += std::uniform_int_distribution<size_t>{0, introset.intros.size() - 1}(rng);
       }
-      m_NextIntro = intros[0];
-
+      m_NextIntro = *it;
       currentConvoTag.Randomize();
       lastShift = Now();
       // add send and connect timeouts to the parent endpoints path alignment timeout
@@ -201,16 +202,12 @@ namespace llarp
       path::Builder::HandlePathBuilt(p);
       p->SetDataHandler([self = weak_from_this()](auto path, auto frame) {
         if (auto ptr = self.lock())
-        {
           return ptr->HandleHiddenServiceFrame(path, frame);
-        }
-        LogWarn("no weak_ptr on data handler");
         return false;
       });
       p->SetDropHandler([self = weak_from_this()](auto path, auto id, auto seqno) {
         if (auto ptr = self.lock())
           return ptr->HandleDataDrop(path, id, seqno);
-        LogWarn("no weak_ptr on data drop");
         return false;
       });
       if (markedBad)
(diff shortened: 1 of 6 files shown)


majestrate:
i really want to merge this so that it fixes and edge case that makes the network suck a lot.

llarp/service/outbound_context.cpp
```diff
@@ -197,12 +201,16 @@ namespace llarp
       path::Builder::HandlePathBuilt(p);
       p->SetDataHandler([self = weak_from_this()](auto path, auto frame) {
         if (auto ptr = self.lock())
+        {
           return ptr->HandleHiddenServiceFrame(path, frame);
+        }
+        LogWarn("no weak_ptr on data handler");
```
jagerman: debug?

llarp/service/outbound_context.cpp
```diff
@@ -65,11 +65,15 @@ namespace llarp
...
+
+      // pick random first intro
+      std::vector<Introduction> intros = introset.intros;
+      if (intros.size() > 1)
       {
-        if (m_NextIntro.latency == 0s or m_NextIntro.latency > intro.latency)
-          m_NextIntro = intro;
+        std::shuffle(intros.begin(), intros.end(), CSRNG{});
       }
+      m_NextIntro = intros[0];
```
jagerman: No need to copy and shuffle here; can just grab a random element via something like:
```C++
auto it = introset.intros.begin();
if (introset.intros.size() > 1) {
  CSRNG rng{};
  it += std::uniform_int_distribution<size_t>{0, intros.size()-1}(rng);
}
m_nextIntro = *it;
```

llarp/service/outbound_context.cpp
```diff
@@ -65,11 +65,15 @@ namespace llarp
...
+
```
jagerman: The change doesn't look safe if this is called with an empty `introset.intros`, while the previous code was.

llarp/service/outbound_context.cpp
```diff
@@ -348,7 +357,10 @@ namespace llarp
         // expunge
         SharedSecret discardme;
         if (not m_DataHandler->GetCachedSessionKeyFor(currentConvoTag, discardme))
+        {
+          LogWarn(Name(), " no cached key after sending intro, we are in a fugged state, oh no");
```
jagerman: This seems like it could only be trigger by a bug, which makes me think it should be LogError, not just Warn.

llarp/service/outbound_context.cpp
```diff
@@ -404,21 +416,33 @@ namespace llarp
...
-        return true;
-
       const auto timeout = std::max(lastGoodSend, m_LastInboundTraffic);
       if (lastGoodSend > 0s and now >= timeout + (sendTimeout / 2))
       {
         // send a keep alive to keep this session alive
         KeepAlive();
         if (markedBad)
+        {
+          LogWarn(Name(), " keepalive timeout hit");
```
jagerman, line 423: Can we include some description info here about *what* is timing out?  I.e. keepalive to whom?

jagerman: Mainly because if we're hitting this a lot it'd be useful to know "we hit this randomly" versus "we hit this with this particular node consistently"

jagerman (approved):

llarp/service/outbound_context.cpp
majestrate: `Name()` has that

llarp/service/outbound_context.cpp
majestrate: added assert

majestrate: `assert(not introset.intros.empty())`

PR #1756, opened 2021-10-06
always build ngtcp2 as static

### LINKED ISSUE #1744: ngtcp2 stuff & shared build incompat
**Describe the bug**
lokinet built in shared mode doesn't install ngtcp2 related libs and thus fails to run as it fails to find relevant .so
i worked this around in archlinux aur pkgbuild by pulling in ngtcp2 dependency, but recently it started failing as apparently upstream have changed something around and removed some functions and thus lokinet fails to start because symbols are missing

**To Reproduce**
build stuff as in https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=lokinet

**Screenshots Or Logs**
`/usr/bin/lokinet: symbol lookup error: /usr/bin/lokinet: undefined symbol: ngtcp2_conn_client_new`

 - OS: arch btw
-  Device: pc
 - Lokinet Version number or Git commit hash: 0.9.5

majestrate:
fixes #1744

### DIFF
--- a/cmake/ngtcp2_lib.cmake
+++ b/cmake/ngtcp2_lib.cmake
@@ -12,6 +12,8 @@ function(add_ngtcp2_lib)
   HexVersion(PACKAGE_VERSION_NUM ${CMAKE_MATCH_1} ${CMAKE_MATCH_2} ${CMAKE_MATCH_3})
   configure_file("ngtcp2/lib/includes/ngtcp2/version.h.in" "ngtcp2/lib/includes/ngtcp2/version.h" @ONLY)
 
+  set(BUILD_SHARED_LIBS OFF)
+
   # Checks for header files.
   include(CheckIncludeFile)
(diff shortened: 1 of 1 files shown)


cmake/ngtcp2_lib.cmake
```diff
@@ -12,6 +12,9 @@ function(add_ngtcp2_lib)
   HexVersion(PACKAGE_VERSION_NUM ${CMAKE_MATCH_1} ${CMAKE_MATCH_2} ${CMAKE_MATCH_3})
   configure_file("ngtcp2/lib/includes/ngtcp2/version.h.in" "ngtcp2/lib/includes/ngtcp2/version.h" @ONLY)
 
+  set(OLD_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
```
jagerman: Saving the value here isn't necessary: cmake functions define a variable scope, so `set`s expire at the end of the function.

jagerman (approved):

PR #1769, opened 2021-10-13
fix debian sid compiler error

majestrate:
not sure why this was even compiling to begin with


jagerman:
What is the compilation error?  That code looks perfectly valid for C code.

majestrate:
> What is the compilation error?  That code looks perfectly valid for C code.

The buffers passed in are 2 dimensional and often the data straddles between dimensions. the use of a pointer is more appropriate.

jagerman (approved):

PR #1770, opened 2021-10-15
merge upstream/stable & bump v0.9.7 centos/8

necro-nemesis:
minor formatting for this purpose.


necro-nemesis:
Patch if still necessary requires refreshing but it appears that /llarp/config/config.cpp now incorporates these changes to upstream dns. Patch2: /SOURCES file currently retained for reference.

SPECS/lokinet.spec
```diff
@@ -24,7 +24,7 @@ BuildRequires:  libsqlite3x-devel
 # Puts the rpm version instead of the git tag in the version string:
 Patch1: version-as-rpm-version.patch
 # Backport default upstream dns not working from PR 1715:
-Patch2: default-upstream-dns.patch
+# Patch2: default-upstream-dns.patch
```
jagerman: You should just delete this line rather than commenting it out, and delete the patch file itself from the repository.

necro-nemesis: Removed. ci failed and required manual restart to build successfully. Building on centos this seems to be a common occurrence.

jagerman (approved):

PR #1773, opened 2021-10-16
merge upstream/stable & bump v0.9.7 fedora/34

necro-nemesis:
Suggest common .gitignore for rpm and deb builds which include SOURCES and contrib


SPECS/lokinet.spec
```diff
@@ -148,6 +148,7 @@ if ! [ -e /etc/loki/lokinet.ini ]; then
 fi
 
 %systemd_post lokinet.service
+systemctl enable --now lokinet
```
jagerman: I asked some Fedora users about this when I originally updated the rpm code, and they felt quite strongly that it shouldn't get enabled by default; but rather than the local user should have to turn it on manually.  This seems to be the norm for Fedora (unlike Debian/Ubuntu): Fedora installs tons of crap you don't need, so nothing get activated by default (whereas on Debian/Ubuntu, you don't install system services unless you want them).

necro-nemesis: I am quite happy to stick with platform norms and remove enable/start from the package.

jagerman (approved):

PR #1778, opened 2021-10-22
check for intersecting ip ranges correctly, add unit test

majestrate:
when a network uses 172.16.0.1/24 we don't detect it right because of a bug, this fixes that bug and adds a unit test to verify that.

### DIFF
--- a/llarp/net/ip_range.hpp
+++ b/llarp/net/ip_range.hpp
@@ -63,6 +63,13 @@ namespace llarp
       return bits::count_bits(netmask_bits);
     }
 
+    /// return true if our range and other intersect
+    constexpr bool
+    operator*(const IPRange& other) const
+    {
+      return Contains(other) or other.Contains(*this);
+    }
+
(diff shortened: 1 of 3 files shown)


llarp/net/ip_range.hpp
```diff
@@ -63,6 +63,13 @@ namespace llarp
       return bits::count_bits(netmask_bits);
     }
 
+    /// return true if our range and other intersect
+    constexpr bool
+    operator*(const IPRange& other) const
+    {
+      return Contains(other) or other.Contains(*this);
+    }
```
jagerman, line 71: I know I suggested the `*` earlier, but thinking some more -- would this be more intuitive as a `&` operator?

majestrate: i feel like `&` could be mistaken as bitwise and which is more likely a thing one would do vs using `*`

jagerman (approved):

PR #1784, opened 2021-10-26
Generate CI images with much DRY-er Python dockerfile generating script

jagerman:
~~Adds cached images that install just the base tools (which are quite substantial), but not the deps (which get installed as part of the build).~~

Replaces the dockerfiles with a Python script that generates them, and adds a bunch of them that are useful for builds debs (namely the new `{debian,ubuntu}-[release]-builder` images, which have only basic deb building tools installed).


contrib/ci/docker/20-deb-builder-bullseye.dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG ARCH=amd64
+FROM registry.oxen.rocks/lokinet-ci-debian-bullseye-base/${ARCH}
```
majestrate: what if we added an 
`ARG FLAVOR=bullseye`
and an 
`ARG DISTRO=debian`

and then do just one dockerfile or something?

jagerman: It might be easier to just replace the whole crappy dockerfile BS (such as lack of includes) with a sane language that generates a dockerfile on the fly.

majestrate: heh

majestrate (approved):

PR #1791, opened 2021-10-29
exclude fully failed nodes from service node testing list

majestrate:
do not requeue nodes for testing from failing queue if we do not have them marked as failing anymore


jagerman (approved):
LGTM, but it's going to fail the lint check :)

majestrate:
fixed

PR #1795, opened 2021-11-09
make PumpLL idempotent to reduce cpu use a bit

majestrate:
in order to reduce excess cpu use in PathContext's `Pump{Up,Down}stream()` make the only place in the code that calls those functions idempotent (`Router::PumpLL`).


llarp/handlers/exit.cpp
```diff
@@ -837,18 +809,14 @@ namespace llarp
...
     {
-      auto range = m_ActiveExits.equal_range(ep->PubKey());
-      auto itr = range.first;
-      while (itr != range.second)
+      for (auto [itr, end] = m_ActiveExits.equal_range(ep->PubKey()); itr != end; ++itr)
       {
         if (itr->second->GetCurrentPath() == ep->GetCurrentPath())
         {
-          itr = m_ActiveExits.erase(itr);
+          m_ActiveExits.erase(itr);
```
majestrate, line 816: this erase will change the thing we are iterating over if it rehashes

jagerman: Yes, but as the next line says right before it returns, "// now ep is gone af"

llarp/handlers/exit.cpp
```diff
@@ -677,19 +655,14 @@ namespace llarp
...
+        return nullptr;
+      if (auto itr = m_ActiveExits.find(pk); itr != m_ActiveExits.end())
       {
-        auto itr = m_ActiveExits.find(pk);
-        if (itr != m_ActiveExits.end())
-        {
-          if (itr->second->PubKey() == pk)
-            endpoint = itr->second.get();
-        }
+        if (itr->second->PubKey() == pk)
```
tewinget, line 664: this `if` seems redundant with the above changes.

tewinget: actually it seems it may have been redundant before as well?

llarp/router/router.cpp
```diff
@@ -69,13 +69,30 @@ namespace llarp
     _running.store(false);
     _lastTick = llarp::time_now_ms();
     m_NextExploreAt = Clock_t::now();
+    m_Pump = _loop->make_waker([this]() {
+      paths.PumpDownstream();
```
majestrate: move these two to PumpLL

tewinget (approved):

PR #1800, opened 2021-11-16
v0.9.8

majestrate:
# v0.9.8

## windows is broken in this tag this release is for linux only (both clients and service nodes)

* fix long uptime cpu usage leak on service nodes
* fix excessive network jitter
* fix issue with service node testing algorithm
* redo ci pipeline docker images


jagerman:
Would like to see a fix for the introset spamming bug we're seeing as well before pushing out.

jagerman (approved):

PR #1811, opened 2021-11-23
parallelize android build much better

majestrate:
* split up configure step and build step so that build steps goes all at once
* update ci to use new build structure for android

### DIFF
--- a/contrib/android-configure.sh
+++ b/contrib/android-configure.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+set -e
+set +x
+
+default_abis="armeabi-v7a arm64-v8a x86 x86_64"
+build_abis=${ABIS:-$default_abis}
+
+test x$NDK = x && echo "NDK env var not set"
+test x$NDK = x && exit 1
+
+echo "building abis: $build_abis"
+
+root="$(readlink -f $(dirname $0)/../)"
+build=$root/build-android
+mkdir -p $build
+cd $build
+
+for abi in $build_abis; do
+    mkdir -p build-$abi
+    cd build-$abi
+    cmake \
+        -G 'Unix Makefiles' \
+        -DANDROID=ON \
+        -DANDROID_ABI=$abi \
+        -DANDROID_ARM_MODE=arm \
+        -DANDROID_PLATFORM=android-23 \
+        -DANDROID_STL=c++_static \
+        -DCMAKE_TOOLCHAIN_FILE=$NDK/build/cmake/android.toolchain.cmake \
+        -DBUILD_STATIC_DEPS=ON \
+        -DBUILD_PACKAGE=ON \
+        -DBUILD_SHARED_LIBS=OFF \
+        -DBUILD_TESTING=OFF \
+        -DBUILD_LIBLOKINET=OFF \
+        -DWITH_TESTS=OFF \
+        -DNATIVE_BUILD=OFF \
+        -DSTATIC_LINK=ON \
+        -DWITH_SYSTEMD=OFF \
+        -DFORCE_OXENMQ_SUBMODULE=ON \
+        -DSUBMODULE_CHECK=OFF \
+        -DWITH_LTO=OFF \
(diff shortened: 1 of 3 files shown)


jagerman:
What if we split this up into 4 separate CI pipelines, i.e. one for each arch?  Given that we're now working from a base-already-installed CI image that ought to parallelize it better still (i.e. across machines) without adding much overhead on the individual jobs.

jagerman:
Oh, never mind, this is building a multi-arch apk, isn't it?

majestrate:
`ABIS=x86 ./contrib/android.sh` should do it given you want to build x86 only

contrib/android-configure.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+set -e
+set +x
+
+default_abis="armeabi-v7a arm64-v8a x86 x86_64"
```
jagerman, line 5: We could probably drop x86 from this.  ISTR there were some issues with it on the wallet, and so we dropped it and no one noticed because there is virtual no x86 android hardware in the wild.

majestrate: the reason i am doing this cleanup is in preparation for potentially doing fdroid packaging. i do not know what the fdroid platforms that are supported

jagerman (approved):

majestrate:
sid clang broke and idk why

majestrate:
json header library broken on clang 13 sid, amazin

jagerman:
> json header library broken on clang 13 sid, amazin

It's not json that broke things, that looks like clang-13 + libstdc++-11 incompatibility in the <valarray> header.

jagerman:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1000398

jagerman:
```
hades:~$ cat foo.cpp 
#include <valarray>
int main() {}

hades:~$ clang++-13 -std=c++17 foo.cpp 
In file included from foo.cpp:1:
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/valarray:1215:5: error: exception specification in declaration does not match previous declaration
    begin(valarray<_Tp>& __va) noexcept
    ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/range_access.h:107:31: note: previous declaration is here
  template<typename _Tp> _Tp* begin(valarray<_Tp>&);
                              ^
In file included from foo.cpp:1:
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/valarray:1226:5: error: exception specification in declaration does not match previous declaration
    begin(const valarray<_Tp>& __va) noexcept
    ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/range_access.h:108:37: note: previous declaration is here
  template<typename _Tp> const _Tp* begin(const valarray<_Tp>&);
                                    ^
In file included from foo.cpp:1:
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/valarray:1237:5: error: exception specification in declaration does not match previous declaration
    end(valarray<_Tp>& __va) noexcept
    ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/range_access.h:109:31: note: previous declaration is here
  template<typename _Tp> _Tp* end(valarray<_Tp>&);
                              ^
In file included from foo.cpp:1:
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/valarray:1253:5: error: exception specification in declaration does not match previous declaration
    end(const valarray<_Tp>& __va) noexcept
    ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/range_access.h:110:37: note: previous declaration is here
  template<typename _Tp> const _Tp* end(const valarray<_Tp>&);
                                    ^
4 errors generated.
```

PR #1812, opened 2021-11-28
demote log statement levels

majestrate:
get rid of log spam


jagerman (approved):
LGTM.

However: do we want to promote this back up to Error/Warn in the future once we solve the known client connect issue?  If so, perhaps drop a `// TODO: bump back to Warn once log spam root cause fixed` comment on them?

PR #1816, opened 2021-12-01
client side exit node pooling

majestrate:
users can now provide multiple `exit-node` values in network section to randomly pool over all provided exits for each range they are mapped to.
implement exit node pooling, allows users to use multiple exits for an address range.
mappings per ip stick to the same exit, each new ip is mapped to a random exit in the specified pool.

### DIFF
--- a/llarp/handlers/tun.cpp
+++ b/llarp/handlers/tun.cpp
@@ -1,7 +1,5 @@
 #include <algorithm>
 #include <variant>
-// harmless on other platforms
-#define __USE_MINGW_ANSI_STDIO 1
 #include "tun.hpp"
 #include <sys/types.h>
 #ifndef _WIN32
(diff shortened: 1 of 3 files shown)


llarp/handlers/tun.cpp
```diff
@@ -1023,6 +1021,38 @@ namespace llarp
...
+      for (const auto& entry : m_ExitMap.FindAllEntries(ip))
+      {
+        // make sure it is allowed by the range if the ip is a bogon
+        if (not IsBogon(ip) or entry.first.BogonContains(ip))
+          candidates.emplace(entry.second);
+      }
+      // no candidates? bail.
+      if (candidates.empty())
+        return std::nullopt;
+      if (exitSelectionStrat == nullptr)
```
tewinget: maybe better as `if (not exitSelectionStrat)` ?

llarp/handlers/tun.hpp
```diff
@@ -222,7 +222,20 @@ namespace llarp
       /// a hidden service
       std::unordered_map<AlignedBuffer<32>, bool> m_SNodes;
 
+      /// maps ip address to an exit endpoint, usefu when we have multiple exits on a range
```
tewinget: `s/usefu/useful/`

tewinget (approved):
LGTM

PR #1818, opened 2021-12-04
configurable route poker

majestrate:
this allows you to use exit nodes without forcing routes over the interface, useful for using lokinet with an exit and selectively routing over the lokinet interface using an external socks proxy or binding to device explicitly.

* make route poker configurable, defaults to enabled but allows disabling it on runtime if desired
* add config option [network]:auto-routing to enable/disable route poker

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -539,10 +539,11 @@ namespace llarp
         ClientOnly,
         Default{true},
         Comment{
-            "enable / disable auto routing. When using an exit lokinet will add routes to "
-            "the OS to make traffic go over the network interface via lokinet.",
-            "enabled by default.",
-        },
+            "Enable / disable automatic route configuration.",
+            "When this is enabled and an exit is used Lokinet will automatically configure "
+            "operating system routes to route traffic through the exit node.",
+            "This is enabled by default, but can be disabled to perform advanced exit routing "
+            "configuration manually."},
         AssignmentAcceptor(m_EnableRoutePoker));
 
     conf.defineOption<std::string>(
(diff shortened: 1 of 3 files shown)


llarp/config/config.cpp
```diff
@@ -531,6 +531,18 @@ namespace llarp
           m_ExitAuths.emplace(exit, auth);
         });
 
+    conf.defineOption<bool>(
+        "network",
+        "route-poker",
```
jagerman: A better user-facing name would be "add-routes"; `route-poker` is more of an internal implementation detail name.

jagerman: Or `"auto-routing"`

llarp/config/config.cpp
```diff
@@ -531,6 +531,18 @@ namespace llarp
...
 
+    conf.defineOption<bool>(
+        "network",
+        "auto-routing",
+        ClientOnly,
+        Default{true},
+        Comment{
+            "enable / disable auto routing. When using an exit lokinet will add routes to "
+            "the OS to make traffic go over the network interface via lokinet.",
+            "enabled by default.",
```
jagerman: Suggested rewording:

"Enable / disable automatic route configuration. When this is enabled and an exit is used Lokinet will automatically configure operating system routes to route traffic through the exit node.  This is enabled by default, but can be disabled to perform advanced exit routing configuration manually."

majestrate: done

necro-nemesis:
I'd like to play with this option if/when it's merged into dev.

jagerman (approved):

PR #1822, opened 2021-12-07
Make windows work again

majestrate:
on win32/apple reading packets from the interface does not count as an io operation.
manually trigger pump on win32/apple to pretend that it is an io event.
add platform quark function MaybeWakeUpperLayers on vpn::Interface to manaully wake up the other components on platforms that need that (ones on which packet io is not done via io events).
on non linux platforms, use uv_prepare_t instead of uv_check_t as the former triggers before blocking for io, instead of after. this better matches linux's order of operations in libuv.

### DIFF
--- a/llarp/rpc/rpc_server.cpp
+++ b/llarp/rpc/rpc_server.cpp
@@ -512,8 +512,7 @@ namespace llarp::rpc
                                   }
                                   onGoodResult(result.reason);
                                 });
-                          },
-                          2s);
+                          });
                     };
                     if (exit.has_value())
                     {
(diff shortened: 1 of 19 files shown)


llarp/rpc/rpc_server.cpp
```diff
@@ -513,7 +513,7 @@ namespace llarp::rpc
                                   onGoodResult(result.reason);
                                 });
                           },
-                          5s);
+                          2s);
```
tewinget: named constant perhaps?

majestrate: done

tewinget (approved):
LGTM

PR #1828, opened 2021-12-19
Add missing libssl-dev

zoobab:
Add missing libssl-dev on Ubuntu to compile it from source.


majestrate (approved):
would've thought libcurl4-openssl-dev would also pull in libssl-dev, guess not. looks good.

PR #1830, opened 2021-12-26
static endpoint auth codes

### LINKED ISSUE #1831: Add static authentication codes on the exit side
Static auth codes have not yet been implemented on the exit side

majestrate:
allow providing a static list of auth codes via config file.

fixes: #1831


llarp/config/config.cpp
```diff
@@ -366,6 +366,16 @@ namespace llarp
           m_AuthWhitelist.emplace(std::move(addr));
         });
 
+    conf.defineOption<std::string>(
+        "network",
+        "auth-static",
+        ClientOnly,
+        MultiValue,
+        Comment{
+            "manually add a static auth code to accept for endpoint auth",
```
jagerman: Should this be "... for exit authentication" rather than "endpoint auth"?

jagerman: Also (I think): += ".  Can be specified multiple times"

majestrate: no, it's for exit or non exit traffic authentication.

jagerman:
Would it make sense to instead have the option point to a file, rather than hard-coding the value, so that you can list multiple in the file and change it over time (without having to restart lokinet)?

(We'd also need some logic to detect file modifications and reload if the mtime changes).

majestrate:
i'd actually like to have another implementation of the endpoint auth type that just takes static values and another that reads info from a file.

majestrate:
rebased

tewinget (approved):
Looks fine, pending CI

PR #1838, opened 2022-01-14
redo systemd status line 

majestrate:
includes number of paths and endpoint. optionally warn about low path success.


tewinget (approved):
LGTM

PR #1839, opened 2022-01-14
route poker: allow not blackholing routes.

majestrate:
allow runtime configuration to disable ip blackholing. some other VPN tunnels that provide ip6 ranges are blocked by lokinet's vpn route poker. this adds a configuration option to toggle this, defaults remain unchanged.


tewinget (approved):
seems reasonable

PR #1843, opened 2022-01-25
Disable unit tests by default (to avoid trigger gcc lto test suite issues)

majestrate:
~~on super old broken versions of gcc compiling unit tests make gcc segfault becuase of a bug in gcc
disable unit tests by default on these targets for users who do not know how to disable building unit tests~~


majestrate:
~~disables LTO by default if on super old compilers now instead of unit tests~~

majestrate:
disable unit tests by default

jagerman (approved):

PR #1844, opened 2022-01-27
wire up sigusr1 to trigger a network thaw on non win32 platforms

majestrate:
add signal handling for lokinet clients that will reset network state on sigusr1, this can be used to manually get out of a weird state.

### DIFF
--- a/llarp/context.cpp
+++ b/llarp/context.cpp
@@ -166,6 +166,14 @@ namespace llarp
       SigINT();
     }
 #ifndef _WIN32
+    if (sig == SIGUSR1)
+    {
+      if (router and not router->IsServiceNode())
+      {
+        LogInfo("SIGUSR1: resetting network state");
+        router->Thaw();
+      }
+    }
     if (sig == SIGHUP)
     {
       Reload();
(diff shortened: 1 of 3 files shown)


llarp/context.cpp
```diff
@@ -166,6 +166,14 @@ namespace llarp
       SigINT();
     }
 #ifndef _WIN32
+    if (sig == SIGUSR1)
+    {
+      if (router and not router->IsServiceNode())
```
jagerman, line 171: Why do we not want this to be possible for a service node?

majestrate: it is a state that a service node should never get into and should be a hard failure if it is. allowing a service node to do this provides no good things only bad things, if a service node can kill all of its outbound link sessions on command it feels like only bad things can happen from allowing admins to do that.

llarp/router/router.cpp
```diff
@@ -807,6 +807,13 @@ namespace llarp
       return;
     // LogDebug("tick router");
     const auto now = Now();
+    constexpr auto TimeskipDetectedDuration = 1min;
```
jagerman: Should move this constant somewhere less buried

jagerman (approved):

PR #1847, opened 2022-02-02
update cross compiling infra

majestrate:
add a `contrib/cross.sh` that makes linux/linux cross compile easier.


.gitignore
```diff
@@ -66,4 +66,3 @@ lokinet-bootstrap.exe
 regdbhelper.dll
 # xcode
 xcuserdata/
-
```
tewinget: extraneous change

majestrate: yeah added stuff to gitignore then removed it later

tewinget (approved):
LGTM

PR #1849, opened 2022-02-02
dry cross compile

majestrate:
this makes the cross compile stuff DRYer


contrib/cross.sh
```diff
@@ -8,10 +8,11 @@ die() {
     exit 1
 }
 
+platform=${PLATFORM:-Linux}
 root="$(readlink -e $(dirname $0)/../)"
 cd $root
 set -e
-set -x
+set +x
```
jagerman:
```suggestion
```

tewinget (approved):

PR #1856, opened 2022-02-14
fix numPaths value in Router::ExtractSummaryStatus

Bilb:


llarp/router/router.cpp
```diff
@@ -150,15 +151,12 @@ namespace llarp
...
 
     // Iterate over all items on this array to build the global pathStats
-    uint64_t paths = 0;
     uint64_t success = 0;
     uint64_t attempts = 0;
     for (const auto& builder : builders)
     {
       if (builder.is_null())
         continue;
-      if (builder["length"].is_number())
```
majestrate, line 160: replacing "length" with "numPaths" in this hunk would fix this issue much better.

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
     // Compute all stats on all path builders on the default endpoint
     // Merge snodeSessions, remoteSessions and default into a single array
     std::vector<nlohmann::json> builders;
-    auto snode_sessions = services["default"]["snodeSessions"];
+
+    auto serviceDefault = services["default"];
```
majestrate: `const auto & serviceDefault = `

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
...
+    builders.push_back(serviceDefault);
+
+    auto snode_sessions = serviceDefault["snodeSessions"];
     for (const auto& session : snode_sessions)
-      builders.push_back(session["buildStats"]);
+      builders.push_back(session);
```
majestrate, line 148: this breaks backwards compat.

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
...
       if (builder.is_null())
         continue;
-      if (builder["length"].is_number())
-        paths += builder["length"].get<uint64_t>();
-      if (builder["success"].is_number())
-        success += builder["success"].get<uint64_t>();
-      if (builder["attempts"].is_number())
-        attempts += builder["attempts"].get<uint64_t>();
+
+      auto paths = builder["paths"];
```
majestrate: `const auto &`

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
...
+      if (paths.is_array()) {
+        for (auto& [key, value] : paths.items()) {
```
majestrate: `const auto &`

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
...
+          if (!value.is_null() && value["status"].is_string() && value["status"] == "established")
```
jagerman:
```suggestion
          if (value.is_object() && value["status"].is_string() && value["status"] == "established")
```

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
...
+      }
+
+      auto buildStats = builder["buildStats"];
+      if (buildStats.is_null())
+        continue;
+
+      if (buildStats["success"].is_number())
+        success += buildStats["success"].get<uint64_t>();
+      if (buildStats["attempts"].is_number())
+        attempts += buildStats["attempts"].get<uint64_t>();
```
jagerman, line 180:
```suggestion
      if (const auto& s = buildStats["success"]; s.is_number())
        success += s.get<uint64_t>();
      if (const auto& a = buildStats["attempts"]; a.is_number())
        attempts += a.get<uint64_t>();
```

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
...
+
+      auto paths = builder["paths"];
+      if (paths.is_array()) {
+        for (auto& [key, value] : paths.items()) {
+          if (!value.is_null() && value["status"].is_string() && value["status"] == "established")
+              pathsCount++;
+        }
+      }
+
+      auto buildStats = builder["buildStats"];
```
jagerman:
```suggestion
      const auto& buildStats = builder["buildStats"];
```

llarp/router/router.cpp
```diff
@@ -139,30 +139,45 @@ namespace llarp
     // Compute all stats on all path builders on the default endpoint
     // Merge snodeSessions, remoteSessions and default into a single array
     std::vector<nlohmann::json> builders;
-    auto snode_sessions = services["default"]["snodeSessions"];
+
+    auto serviceDefault = services["default"];
+    builders.push_back(serviceDefault);
+
+    auto snode_sessions = serviceDefault["snodeSessions"];
```
jagerman: Most of these should probably be `variable.at("key")` rather than `variable["key"]`; the latter will create an element (with a default null value) if the key doesn't already exist, while the former checks and throws in such a case (and I believe most of these will always be present, so throwing if they aren't seems acceptable).

llarp/router/router.cpp
jagerman: I think that's probably okay?  The only consumer of this endpoint is going to be `lokinetmon` which ought to be easily fixed.

llarp/router/router.cpp
jagerman: I think these values should always be set (and I think it's a bug if not), so this can simply be:
```suggestion
      success += buildStats["success"].get<uint64_t>();
      attempts += buildStats["attempts"].get<uint64_t>();
```
to throw if not present or not an int.

llarp/router/router.cpp
Bilb: I took that from the old control panel https://github.com/oxen-io/lokinet-control-panel/blob/master/res/qml/ControlPanel.qml#L229

jagerman: We'll fix in another PR if it breaks lokinetmon (though Jeff thinks it might actually fix something so...)

jagerman (approved):

PR #1857, opened 2022-02-15
use quad nine as default upstream dns

majestrate:
replaces 1.1.1.1 with ~~9.9.9.9~~ 9.9.9.10 for default upstream dns

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -740,7 +740,7 @@ namespace llarp
 #endif
 
     // Default, but if we get any upstream (including upstream=, i.e. empty string) we clear it
-    constexpr Default DefaultUpstreamDNS{"1.1.1.1"};
+    constexpr Default DefaultUpstreamDNS{"9.9.9.10"};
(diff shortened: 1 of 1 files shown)


rany2 (approved):
based

llarp/config/config.cpp
```diff
@@ -740,7 +740,7 @@ namespace llarp
 #endif
 
     // Default, but if we get any upstream (including upstream=, i.e. empty string) we clear it
-    constexpr Default DefaultUpstreamDNS{"1.1.1.1"};
+    constexpr Default DefaultUpstreamDNS{"9.9.9.9"};
```
jagerman: 9.9.9.11 might be a better default.  `.9` says it blocks some things, and it isn't clear what.

jagerman (approved):

llarp/config/config.cpp
jagerman: Oops, mixed it up.  I mean .10, not .11.

PR #1860, opened 2022-02-20
add version and uptime to summary status endpoint

Bilb:


llarp/router/router.cpp
```diff
@@ -182,6 +182,8 @@ namespace llarp
 
     return util::StatusObject{
         {"running", true},
+        {"version", llarp::VERSION_FULL},
+        {"uptime", to_json(this->Uptime())},
```
majestrate: superfluous use of `this`

Bilb: fixed

majestrate (approved):

PR #1862, opened 2022-02-23
summary status: no services means not running

Bilb:


majestrate (changes requested):
when running as service node, this data is also null, add check for when running as service node.

llarp/router/router.cpp
```diff
@@ -116,6 +116,9 @@ namespace llarp
       return util::StatusObject{{"running", false}};
 
     auto services = _hiddenServiceContext.ExtractStatus();
+    if (services.is_null())
```
majestrate:
```suggestion
    if (services.is_null() and not IsServiceNode())
```

majestrate: actually no, let me add a commit for this instead.

majestrate (approved):

PR #1863, opened 2022-02-25
add new rpc endpoint

majestrate:
llarp.async-exit command will reply with a llarp.notify-exit command with the same values as llarp.exit
this can be used to bypass request timeoutes in omq

refactor internal bits in rpc to be a bit more dry
(still wet af)


llarp/rpc/rpc_server.cpp
```diff
@@ -94,6 +94,210 @@ namespace llarp::rpc
...
+  void
+  HandleJSONCommand(
+      oxenmq::Message& msg,
+      std::string replyCmd,
+      std::function<void(nlohmann::json, ReplyFunction_t)> handleRequest)
+  {
+    const auto maybe = MaybeParseJSON(msg);
+    if (not maybe.has_value())
+    {
+      msg.send_back(replyCmd, CreateJSONError("failed to parse json"));
```
tewinget: very minor style thought: "JSONError" looks like it could be a type and this could just be a constructor, is the "Create" superfluous?

llarp/rpc/rpc_server.cpp
```diff
@@ -94,6 +94,210 @@ namespace llarp::rpc
...
   }
 
+  void
+  HandleJSONCommand(
+      oxenmq::Message& msg,
+      std::string replyCmd,
+      std::function<void(nlohmann::json, ReplyFunction_t)> handleRequest)
+  {
+    const auto maybe = MaybeParseJSON(msg);
+    if (not maybe.has_value())
```
tewinget: don't need that `.has_value()`

llarp/rpc/rpc_server.cpp
```diff
@@ -94,6 +94,210 @@ namespace llarp::rpc
...
+    if (not maybe->is_object())
+    {
+      msg.send_back(replyCmd, CreateJSONError("data not a json object"));
+      return;
+    }
+    try
+    {
+      handleRequest(*maybe, [defer = msg.send_later(), cmd = replyCmd](std::string result) {
+        defer(cmd, result);
+      });
```
jagerman: Could move the above cases in here and then `throw std::runtime_error{"bad stuff"}` so that all the error cases DRY the `send_back` to one place in the `catch`.

llarp/rpc/rpc_server.cpp
```diff
@@ -94,6 +94,210 @@ namespace llarp::rpc
...
+  {
+    if (r->IsServiceNode())
+    {
+      reply(CreateJSONError("not supported"));
+      return;
+    }
+    std::optional<llarp::service::Address> exit;
+    std::optional<std::string> lnsExit;
+    IPRange range;
+    bool map = true;
```
tewinget, line 139: not a big deal, but could be declared closer to where it's used.  also could use a better variable name.

llarp/rpc/rpc_server.cpp
```diff
@@ -94,6 +94,210 @@ namespace llarp::rpc
...
+        return;
+      }
+      else
+      {
+        exit = addr;
+      }
+    }
+
+    const auto unmap_itr = obj.find("unmap");
+    if (unmap_itr != obj.end() and unmap_itr->get<bool>())
```
jagerman:
```suggestion
    if (auto it = obj.find("unmap"); it != obj.end() and it->get<bool>())
```

llarp/rpc/rpc_server.cpp
```diff
@@ -94,6 +94,210 @@ namespace llarp::rpc
...
+    if (r->IsServiceNode())
+    {
+      reply(CreateJSONError("not supported"));
+      return;
+    }
+    std::optional<llarp::service::Address> exit;
+    std::optional<std::string> lnsExit;
+    IPRange range;
+    bool map = true;
+    const auto exit_itr = obj.find("exit");
```
tewinget: this amount of find->if->set with json definitely warrants a helper function to simplify it.

llarp/rpc/rpc_server.cpp
```diff
@@ -94,6 +94,210 @@ namespace llarp::rpc
...
+                {
+                  onBadResult("could not find exit");
+                  return;
+                }
+                if (not shouldSendAuth)
+                {
+                  onGoodResult("OK: connected to " + addrStr);
+                  return;
+                }
+                ctx->AsyncSendAuth([onGoodResult, onBadResult](service::AuthResult result) {
```
tewinget: A lambda this deep is basically an omicron at this point...

llarp/rpc/rpc_server.cpp
majestrate: no we can't because this lambda can be called from the logic thread or a random zmq thread depending on what branch is taken.

majestrate: (this is why the defer is there to begin with)

llarp/rpc/rpc_server.cpp
```diff
@@ -1,7 +1,7 @@
 #include "rpc_server.hpp"
 #include <llarp/router/route_poker.hpp>
 #include <llarp/constants/version.hpp>
-#include <nlohmann/json.hpp>
```
tewinget, line 4: may as well still include this explicitly, rather than implicitly, since we're directly using the `nlohmann::json` type in here.

llarp/rpc/rpc_server.cpp
```diff
@@ -128,173 +127,140 @@ namespace llarp::rpc
...
     std::optional<llarp::service::Address> exit;
     std::optional<std::string> lnsExit;
     IPRange range;
     bool map = true;
-    const auto exit_itr = obj.find("exit");
-    if (exit_itr != obj.end())
+
+    if (auto maybe = llarp::json::maybe_get<bool>(obj, "unmap", false))
+    {
+      map = *maybe != true;
```
tewinget, line 143: `map = not *maybe`?

tewinget: though that's hideous with the var name "maybe"

majestrate: the meaning is more explicit with `!= true`

llarp/rpc/rpc_server.cpp
majestrate: `llarp/util/json.hpp` pulls this in

tewinget: I assumed as much, but I still think it would be better to include nlohmann/json explicitly since we're using types from it directly, rather than abstracted via llarp/util/json

majestrate: we do that already.

majestrate:
rebased ontop of dev

llarp/rpc/rpc_server.cpp
```diff
@@ -265,390 +428,195 @@ namespace llarp::rpc
...
-                if (const auto itr = obj.find("qtype"); itr != obj.end())
-                {
-                  qtype = itr->get<dns::QType_t>();
-                }
+                if (auto maybe = json::maybe_get<dns::QType_t>(obj, "qtype", dns::qTypeA))
+                  qtype = *maybe;
 
                 dns::Message msg{dns::Question{qname, qtype}};
-
+                // TODO: race condition
```
tewinget: TODO: this

llarp/rpc/rpc_server.cpp
```diff
@@ -265,390 +428,195 @@ namespace llarp::rpc
...
-                      reply(CreateJSONError(stringify(
-                          "failed to set [", section, "]:", key, " value is not a string")));
-                      return;
-                    }
-                    r->GetConfig()->Override(section, key, value.get<std::string>());
-                  }
+                  if (not value.is_string())
+                    throw std::runtime_error{
+                        stringify("failed to set [", section, "]:", key, " value is not a string")};
+                  // TODO: race condition
```
tewinget: TODO: this

majestrate:
conflicts with #2121 so closing

PR #1865, opened 2022-03-06
fix wire protocol race condition

majestrate:
only send close packet once, before we were sending a close after we got a close causing excess log spam.
include handshake phase when checking for connection timeouts.

when we change our rc make sure to put it into nodedb too when we are a service node to prevent weirdness in dht lookups.


jagerman (approved):
LGTM.

PR #1880, opened 2022-03-30
"fast open" snode sessions

majestrate:
initial support for "fast open" paths to snodes.

### DIFF
--- a/llarp/iwp/message_buffer.hpp
+++ b/llarp/iwp/message_buffer.hpp
@@ -55,9 +55,7 @@ namespace llarp
       bool
       operator<(const OutboundMessage& msg) const
       {
-        // yes, the first order is reversed as higher means more important
-        // second part is for queue order
-        return msg.m_ResendPriority < m_ResendPriority or m_MsgID < msg.m_MsgID;
+        return msg.m_ResendPriority < m_ResendPriority;
       }
 
       ILinkSession::Packet_t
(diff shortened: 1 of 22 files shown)


llarp/iwp/message_buffer.hpp
```diff
@@ -49,6 +50,13 @@ namespace llarp
       llarp_time_t m_LastFlush = 0s;
       ShortHash m_Digest;
       llarp_time_t m_StartedAt = 0s;
+      uint16_t m_ResendPriority;
+
+      bool
+      operator<(const OutboundMessage& msg) const
+      {
+        return msg.m_ResendPriority < m_ResendPriority;
+      }
```
jagerman: Rather than defining `operator <` in a way that seems wrong at first glance, it might be better to define this `operator<` the "natural" way and then use a templated typedef in a util header somewhere to define a lower-is-better priority queue:
```C++
template <typename T, typename Container = std::vector<T>>
using ascending_priority_queue = std::priority_queue<T, Container, std::greater<typename Container::value_type>>;
```
which you can use as `negative_priority_queue<MyType> stuff{};` to give you a smaller-comes-first priority queue.

majestrate: added

majestrate:
this branch is broke af, i do not think it can be in the 0.9.9 tag.

majestrate:
i am going to need to redo this to make it not horrible.

PR #1890, opened 2022-04-10
[chores] add -Werror and such compiler warnings

majestrate:
* add virtual destructors
* add cmake-isms to bump up deafult warning / error levels to check for such because gcc doesn't by default (why tho?)


tewinget (approved):

majestrate:
moved the modifications to win32 vpn platform to own commit so when i rebase the win32 route poker branch onto dev after this is merged i can drop 1 commit that will conflict.

majestrate:
dead PR, most fixed by now by other commits.

PR #1891, opened 2022-04-12
idempotent flush queues on path builds

majestrate:
this cherry-picks the path build queue flushing from my testnet perf branch so it can easily be merged into dev.


tewinget (approved):
LGTM

PR #1895, opened 2022-04-16
freebsd vpn platform

### LINKED ISSUE #1513: proper freebsd tun interface
after redoing the tun code the freebsd support that was rather jenk was removed.
we need to add it back and do it correctly this time.

majestrate:
adds freebsd vpn platform code 

fixes #1513


majestrate:
will be redone in internals refactor

PR #1896, opened 2022-04-18
handle edge case better

majestrate:
when our path is fine but the recipiant's path on the pivot router isn't we should pivot to another router


majestrate:
there is no guarantee that the same pivot router will have another intro we can use and it's far more likely it doesn't so we'll just pivot.

tewinget (approved):
Makes sense.

PR #1897, opened 2022-04-18
split up window.sh like we do for android.sh

majestrate:
this makes it easier to build windows in a directory that isn't build-windows

i do this on my machine so i can exempt just one directory from my editor for all generated build files


majestrate:
upserted by #1930

PR #1901, opened 2022-04-22
add lokinet.swift cli args

majestrate:
* add --start flag to start lokinet
* add --stop flag to stop
* by default lokinet on macos needs a flag or it will be a nop

### DIFF
--- a/daemon/lokinet.swift
+++ b/daemon/lokinet.swift
@@ -4,31 +4,70 @@ import NetworkExtension
 
 let app = NSApplication.shared
 
+let START = "--start"
+let STOP = "--stop"
+
+let HELP_STRING = "usage: lokinet [--start|--stop]"
+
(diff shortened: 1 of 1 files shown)


daemon/lokinet.swift
```diff
@@ -4,31 +4,70 @@ import NetworkExtension
...
+
 class LokinetMain: NSObject, NSApplicationDelegate {
     var vpnManager = NETunnelProviderManager()
     let lokinetComponent = "com.loki-project.lokinet.network-extension"
+    var mode = ""
 
     func applicationDidFinishLaunching(_: Notification) {
-        setupVPNJizz()
+        if self.mode == START {
+            setupVPNJizz()
```
jagerman: s/Jizz//g

majestrate: jizz removal completed

jagerman:
Included in #1942

PR #1903, opened 2022-04-28
Replace llarp/util/endian.hpp with oxenc/endian.h

jagerman:


majestrate:
we'll need this https://github.com/oxen-io/oxen-encoding/pull/5 and a submodule bump of oxenc before we can merge this

majestrate (changes requested):
needs https://github.com/oxen-io/oxen-encoding/pull/5 and a submodule bump after that is merged to oxenc

majestrate (approved):

PR #1905, opened 2022-04-28
connect to routers even if we are decomissioned

### LINKED ISSUE #1904: Allow&make snode connections while decommissioned
With lokinet reachability enforcement, you can get stuck decommissioned because, while decommissioned, lokinet currently isn't establishing sessions with other routers, and so doesn't get its RC out to the network, which means other routers don't necessarily get the updated RC info they might need to perform a proper reachability test (and thus allow you to get recommissioned).

majestrate:
fixes #1904

### DIFF
--- a/llarp/link/server.cpp
+++ b/llarp/link/server.cpp
@@ -8,6 +8,7 @@
 #include <utility>
 #include <unordered_set>
 #include <llarp/router/abstractrouter.hpp>
+#include <oxenc/variant.h>
 
 static constexpr auto LINK_LAYER_TICK_INTERVAL = 100ms;
 
@@ -129,7 +130,7 @@ namespace llarp
   }
 
   bool
-  ILinkLayer::Configure(AbstractRouter* router, const std::string& ifname, int af, uint16_t port)
+  ILinkLayer::Configure(AbstractRouter* router, std::string ifname, int af, uint16_t port)
   {
     m_Router = router;
     m_udp = m_Router->loop()->make_udp(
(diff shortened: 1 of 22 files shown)


jagerman:
Confirmed on testnet that this works.

While testing this I noticed something odd that should also be addressed: unregistered routers are still performing tests -- these tests always fail, however, because the testee rejects a connection from an unregistered router.  We should just stop testing completely if we aren't active or decomissioned.

llarp/router/router.cpp
```diff
@@ -889,15 +908,22 @@ namespace llarp
...
+        and _rcLookupHandler.SessionIsAllowed(pubkey());
 
-    if (_rc.ExpiresSoon(now, std::chrono::milliseconds(randint() % 10000))
-        || (now - _rc.last_updated) > rcRegenInterval)
+    if (isSvcNode
+        and (_rc.ExpiresSoon(now, std::chrono::milliseconds(randint() % 10000)) or (now - _rc.last_updated) > rcRegenInterval))
     {
       LogInfo("regenerating RC");
-      if (!UpdateOurRC(false))
-        LogError("Failed to update our RC");
```
jagerman, line 898: I think we should keep this in an `else` so that we know from log output when something getting wrong.

llarp/util/time.cpp
```diff
@@ -74,4 +78,27 @@ namespace llarp
...
+  }
+  namespace util
+  {
+    std::string
+    time_delta_string(const TimePoint_t& at)
+    {
+      std::stringstream ss;
+      const auto dlt = TimePoint_t::clock::now() - at;
+      if (dlt > 0s)
+        ss << std::chrono::duration_cast<Duration_t>(TimePoint_t::clock::now() - at) << " ago ";
```
jagerman:
```suggestion
        ss << std::chrono::duration_cast<Duration_t>(dlt) << " ago ";
```

llarp/util/time.cpp
```diff
@@ -74,4 +78,27 @@ namespace llarp
...
+      else if (dlt < 0s)
+        ss << "in " << std::chrono::duration_cast<Duration_t>(TimePoint_t::clock::now() - at);
```
jagerman: I think this is supposed to be negated, i.e. `at - now`, or alternatively:
```suggestion
        ss << "in " << std::chrono::duration_cast<Duration_t>(-dlt);
```

llarp/link/server.cpp
```diff
@@ -139,28 +140,60 @@ namespace llarp
           std::copy_n(buf.base, buf.sz, pkt.data());
           RecvFrom(from, std::move(pkt));
         });
-
+    std::string _ifname{ifname};
```
jagerman: If it needs an unconditional copy, might as well take it by value rather than `const&`.

jagerman (approved):

PR #1907, opened 2022-05-02
propagate message priority to link layer

majestrate:
this adds awareness of link layer message priority to the link layer itself so that it can re transmit as needed given said priority.


llarp/iwp/session.cpp
```diff
@@ -127,7 +129,7 @@ namespace llarp
       {
         LogError("failed to encode LIM for ", m_RemoteAddr);
```
tewinget, line 130: should `return`

llarp/iwp/session.cpp
```diff
@@ -127,7 +129,7 @@ namespace llarp
...
       }
-      if (!SendMessageBuffer(std::move(data), h))
+      if (not SendMessageBuffer(std::move(data), h))
       {
         LogError("failed to send LIM to ", m_RemoteAddr);
```
tewinget, line 134: should `return`

llarp/iwp/session.cpp
tewinget, line 130:
```suggestion
        LogError("failed to encode LIM for ", m_RemoteAddr);
        return;
```

llarp/iwp/session.cpp
tewinget, line 135:
```suggestion
        LogError("failed to send LIM to ", m_RemoteAddr);
        return;
```

llarp/iwp/session.cpp
```diff
@@ -253,15 +256,22 @@ namespace llarp
             msg.SendACKS(util::memFn(&Session::EncryptAndSend, this), now);
           }
         }
+        std::priority_queue<
+            OutboundMessage*,
+            std::vector<OutboundMessage*>,
+            ComparePtr<OutboundMessage*>>
+            resend;
```
tewinget: I'd rather if this were called `resend_queue`, as `resend` feels like either a function or a boolean.

llarp/router/outbound_message_handler.cpp
```diff
@@ -86,7 +84,7 @@ namespace llarp
     m_Killer.TryAccess([this]() {
       recentlyRemovedPaths.Decay();
       ProcessOutboundQueue();
-      if (/*bool more = */ SendRoundRobin())
```
tewinget, line 89: Without this comment it's easy to conclude that `SendRoundRobin` returns a success/fail rather than the actual purpose of its return, so I think it should stay.

majestrate: honestly the entire reason for this type is that you dont want to continue processing if you dont process it all.
we should revert that, but we'd need a lot of testing here for that. i dont feel like it is okay to do 1 week before hard fork.

llarp/router/outbound_message_handler.cpp
```diff
@@ -86,7 +84,7 @@ namespace llarp
...
+      if (SendRoundRobin())
```
tewinget:
```suggestion
      //TODO: this probably shouldn't be pumping, as it defeats the purpose
      // of having a limit on sends per tick, but chaning it is potentially bad
      // and requires testing so it should be changed later.
      if (/*bool more = */ SendRoundRobin())
```

tewinget (approved):

cmake/StaticBuild.cmake
```diff
@@ -48,11 +48,11 @@ set(ZMQ_SOURCE zeromq-${ZMQ_VERSION}.tar.gz)
 set(ZMQ_HASH SHA512=e198ef9f82d392754caadd547537666d4fba0afd7d027749b3adae450516bcf284d241d4616cad3cb4ad9af8c10373d456de92dc6d115b037941659f141e7c0e
     CACHE STRING "libzmq source hash")
 
-set(LIBUV_VERSION 1.43.0 CACHE STRING "libuv version")
+set(LIBUV_VERSION 1.44.1 CACHE STRING "libuv version")
 set(LIBUV_MIRROR ${LOCAL_MIRROR} https://dist.libuv.org/dist/v${LIBUV_VERSION}
     CACHE STRING "libuv mirror(s)")
 set(LIBUV_SOURCE libuv-v${LIBUV_VERSION}.tar.gz)
-set(LIBUV_HASH SHA256=90d72bb7ae18de2519d0cac70eb89c319351146b90cd3f91303a492707e693a4
+set(LIBUV_HASH SHA512=b4f8944e2c79e3a6a31ded6cccbe4c0eeada50db6bc8a448d7015642795012a4b80ffeef7ca455bb093c59a8950d0e1430566c3c2fa87b73f82699098162d834)
```
jagerman:
```suggestion
set(LIBUV_HASH SHA512=b4f8944e2c79e3a6a31ded6cccbe4c0eeada50db6bc8a448d7015642795012a4b80ffeef7ca455bb093c59a8950d0e1430566c3c2fa87b73f82699098162d834
```

PR #1910, opened 2022-05-07
use libuv threadpool instead of oxenmq's threadpool.

majestrate:
add EventLoop work pure virtual method for doing threaded work on the event loop.

switch over libuv to provide work via their threadpool.

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -41,6 +41,12 @@ namespace llarp
     constexpr Default DefaultWorkerThreads{0};
     constexpr Default DefaultBlockBogons{true};
 
+#if defined(__linux__) && !defined(_WIN32) && !defined(ANDROID)
+    constexpr Default DefaultThreadPoolImpl{"libuv"};
+#else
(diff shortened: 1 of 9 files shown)


majestrate:
this had the cost of 1 file handle per job which is no bueno

majestrate:
On Sun, 08 May 2022 06:15:53 -0700
majestrate ***@***.***> wrote:

> this had the cost of 1 file handle per job which is no bueno
> 
> 

i am pretty sure i need to revisit how we do threaded work, probably
adding a concept for batching.

jagerman:
I'm not sure about Windows (and that might be worth investigating), but on Linux (and I'd expect macos and pretty much anything else unix-like) oxenmq's job scheduling has noticeably less overhead per job than a traditional thread pool because it avoids any mutex lock contention between worker threads looking for jobs (it is instead all using inproc sockets to get job info from the master thread to worker threads).

jagerman:
Though a filehandler per job seems like pretty big overhead even compared to a traditional mutex-locking thread queue.

PR #1914, opened 2022-05-10
diversify intros in introset

majestrate:
when we generate a new introset before we were using the most recent intros deterministically, this means if we have more than the requested intros we will always use the same ones. there are cases in which the recipient does not put out their introset often enough as the intros remained unchanged.

this mixes which ones we publish so that each publish has a different set in each publish, the net effect being that we will never put out an introset with the same intros in it each time. 

this will permit traffic to spread out more instead of sticking to a deterministically selected set of intros.

### DIFF
--- a/llarp/crypto/crypto.hpp
+++ b/llarp/crypto/crypto.hpp
@@ -178,4 +178,15 @@ namespace llarp
     };
   };
 
+  namespace util
+  {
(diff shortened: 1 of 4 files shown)


majestrate:
this will be re approached in internals refactor.

PR #1917, opened 2022-05-18
clean up ip packet code

majestrate:
make the ip packet code happy with new oxenc stuff, remove older cruft around it.


llarp/net/ip_packet.hpp
```diff
@@ -4,324 +4,261 @@
...
 
-      inline ipv6_header*
-      HeaderV6()
-      {
-        return (ipv6_header*)&buf[0];
-      }
+    inline ip_header*
+    Header()
+    {
+      return reinterpret_cast<ip_header*>(&buf[0]);
```
jagerman, line 155: `buf` should probably be declared: `alignas(ip_header) byte_t buf[MaxSize];` to make these reinterpret_cast's valid (on arches that don't like unaligned access).

jagerman (approved):

llarp/net/ip_packet.hpp
majestrate: resolved

jagerman: (Given that buf follows a size_t member it is *probably* already properly aligned, but still worth being explicit about the alignment requirement).

PR #1918, opened 2022-05-19
fixup edge case on windows

majestrate:
* add platform detection constexprs
* add quark for platforms without native ipv6 like windows, exit mapping cannot work with ipv6 yet

### DIFF
--- a/llarp/constants/platform.hpp
+++ b/llarp/constants/platform.hpp
@@ -1,0 +1,65 @@
+#pragma once
+
+/// namespace for platform feature detection constexprs
+namespace llarp::platform
+{
+  ///  are we linux  ?
+  inline constexpr bool is_linux =
+#ifdef __linux__
+      true
+#else
+      false
+#endif
+      ;
+
+  ///  are we freebsd ?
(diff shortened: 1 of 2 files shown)


llarp/constants/platform.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+/// namespace for platform feature detection constexprs
+namespace llarp::platform
+{
+  ///  are we linux  ?
+  constexpr bool is_linux =
```
jagerman: These need `inline` added to avoid potential duplicate symbol clashes:
```suggestion
  inline constexpr bool is_linux =
```

llarp/constants/platform.hpp
```diff
@@ -0,0 +1,65 @@
...
+#ifdef __apple__
+      true
+#else
+      false
+#endif
+      ;
+
+  /// are we an android platform ?
+  constexpr bool is_android =
+#if defined(ANDROID)
```
jagerman: Minor nitpick: this is a bit inconsistent with the `ifdef`s above.

jagerman (approved):

PR #1930, opened 2022-06-03
redo bind section of config

### LINKED ISSUE #1929: Binding to wrong IP
On a service node I have the following config (irrelevant bits of config snipped):
```
[router]
public-ip=149.56.165.112
public-port=10909
[bind]
0.0.0.0=10909
```
but the lokinet's open udp ports end up as:
```
udp        0      0 51.79.20.22:10909       0.0.0.0:*                           1138933/lokinet     
udp        0      0 127.109.0.0:53          0.0.0.0:*                           1138933/lokinet     
udp        0      0 149.56.165.112:41937    0.0.0.0:*                           1138933/lokinet     
```
but this isn't exhibiting the right behaviour: the *output* IP is correctly bound to the public-ip, but the inbound IP looks like it is using the first IP listed on the interface rather than `public-ip`.

The intended behaviour here was to fail in such a case -- i.e. I have more than one IP, and didn't specify one in the `[bind]` section, but it isn't failing: lokinet is starting and apparently just using the first address on the interface.  (The result is an unreachable lokinet, because 149.56.165.112 is the default source IP, but that doesn't work right when we reply to something that we received on 51.x).

I think we need to split up this configuration better, and make the behaviour a little more obvious:

- add new options `incoming=1.2.3.4:1090` and `outgoing=5.6.7.8:1099` options to the `[bind]` section, and make this the only non-deprecated way to specify IPs.  We should probably allow `incoming=:1090` as an alias for `0.0.0.0:1090`
- we likely need to allow multiple `incoming` addresses, probably not `outgoing`.
- for future use, we'll likely want some other keys here (e.g. `quic-in=`, `quic-out=`).
- if you bind to IP 0.0.0.0 (either with the new, or the deprecated, interface, or using a new `:1090` value) then:
  - if we have a `public-ip` given in the `[router]` section and that IP exists on the system, bind to it for both incoming/outgoing IPs.
  - otherwise go looking for a unique public IP and, if we find one (and only one), bind to that.
  - if we find more than one public IP (or no public IP), die with a message that you need to list a specific IP in the `[bind]` section.
- deprecate specifying interface names.  They are non-portable across machines, and also don't uniquely identify an IP (i.e. one interface can have multiple IPs).

majestrate:
fixes #1929


llarp/config/config.cpp
```diff
@@ -834,48 +834,66 @@ namespace llarp
   }
 
   LinksConfig::LinkInfo
-  LinksConfig::LinkInfoFromINIValues(std::string_view name, std::string_view value)
+  LinksConfig::LinkInfoFromINIValues(
+      std::string_view name, std::string_view value, const net::Platform& net)
   {
     // we treat the INI k:v pair as:
-    // k: interface name, * indicating outbound
+    // k: interface name, * indicating outbound, inbound for inbound addr, outbound for outbound
```
tewinget: This was already kinda weird for an ini file, but I think is weird enough to merit a set of examples.

jagerman: I don't think it needs an example: the k:v pair stuff is what is being deprecated here; the (non-deprecated) config is now:
```ini
[bind]
inbound=1.2.3.4:5678
outbound=5.6.7.8:9876
```

llarp/config/config.cpp
```diff
@@ -834,48 +834,66 @@ namespace llarp
...
-      if (not addr.hasPort())
-        throw std::invalid_argument("no port provided in link address");
-      info.m_interface = addr.toHost();
-      info.port = *addr.getPort();
+      const SockAddr addr{value};
+      LinkInfo info{};
+
+      if (net.IsLoopbackAddress(addr.getIP()))
+        throw llarp::make_exception<std::invalid_argument>(
+            value, " is a loopback address and cannot be used with link addresses");
```
jagerman:
```suggestion
            value, " is a loopback address and cannot be used with bind addresses");
```

llarp/config/config.cpp
```diff
@@ -834,48 +834,66 @@ namespace llarp
...
+
+      info.m_interface = var::visit([](auto&& ip) { return ip.ToString(); }, addr.getIP());
```
jagerman: It feels strange to set something that is always an IP address into a variable named `m_interface`.

llarp/config/config.cpp
```diff
@@ -834,48 +834,66 @@ namespace llarp
...
     else
     {
-      info.m_interface = std::string{name};
-
-      std::vector<std::string_view> splits = split(value, ",");
-      for (std::string_view str : splits)
+      // deprecated
+      LogWarn(
+          "[bind] section specified a link by interface name, this is deprecated and will be "
+          "removed later");
```
jagerman:
```suggestion
          "removed in a future release");
```

llarp/config/config.cpp
majestrate: it really should be a `llarp::SockAddr` at this point instead of string and int

llarp/config/config.cpp
```diff
@@ -834,48 +834,66 @@ namespace llarp
...
     }
     else
     {
-      info.m_interface = std::string{name};
-
-      std::vector<std::string_view> splits = split(value, ",");
-      for (std::string_view str : splits)
+      // deprecated
+      LogWarn(
+          "[bind] section specified a link by interface name, this is deprecated and will be "
```
jagerman: This is also going to catch the case where you have specified `0.0.0.0=1234`, in which case this error message is going to be confusing.

llarp/config/config.cpp
```diff
@@ -903,22 +921,26 @@ namespace llarp
             "IP address and use that (with port 1090).",
```
jagerman: The section comments need to be updated as they are describing all the deprecated behaviour now.

llarp/config/config.cpp
```diff
@@ -903,22 +921,26 @@ namespace llarp
...
-    conf.defineOption<std::string>(
-        "bind",
-        "*",
-        DefaultOutboundLinkValue,
-        Comment{
-            "Specify a source port for **outgoing** Lokinet traffic, for example if you want to",
-            "set up custom firewall rules based on the originating port. Typically this should",
-            "be left unset to automatically choose random source ports.",
-        },
-        [this](std::string arg) { m_OutboundLink = LinkInfoFromINIValues("*", arg); });
```
jagerman, line 920: We need some replacement `defineOption`s here to populate the default generated config with the new versions

llarp/config/config.cpp
```diff
@@ -1019,7 +1038,7 @@ namespace llarp
         },
         AssignmentAcceptor(whitelistRouters));
 
-    conf.defineOption<std::string>("lokid", "jsonrpc", RelayOnly, [](std::string arg) {
+    conf.defineOption<std::string>("lokid", "jsonpc", RelayOnly, [](std::string arg) {
```
jagerman: What?

majestrate: typo was comitted

llarp/config/config.hpp
```diff
@@ -35,11 +35,17 @@ namespace llarp
 
   // TODO: don't use these maps. they're sloppy and difficult to follow
   /// Small struct to gather all parameters needed for config generation to reduce the number of
-  /// parameters that need to be passed around.
+  /// parameters that need to be passed arocund.
```
jagerman: -c

llarp/net/net_int.cpp
```diff
@@ -128,6 +128,17 @@ namespace llarp
...
+
+  template <>
+  std::string
+  nuint128_t::ToString() const
+  {
+    char tmp[INET6_ADDRSTRLEN] = {0};
+    if (!inet_ntop(AF_INET6, (void*)&n, tmp, sizeof(tmp)))
+      return "";
+    return tmp;
+  }
```
jagerman, line 143: This feels wrong.  If nuint128_t is only supposed to be an IPv6 address, then it is misnamed; and if it is meant to be a more generic network-order-128-bit-int then ToString is misnamed (and should rather be something like `ToIPv6String`).

(The same applies to the existing [nh]uint32_t and IPv4 address stringification).

majestrate: huint types need to be made invisible or internal and everything really should be in network byte order, the reason this pr took so long is because i actually did make it do that but ended up reverting it as it became.... huge.

the better names for `nuint16_t` `nuint32_t` and `nuint128_t` would be something like `net_port_t` `ip4_address_t` and `ip6_address_t` respectively, i supposed i will do this next time i refactor all the endian related parts.

majestrate:
review on monday.

llarp/config/config.cpp
```diff
@@ -833,111 +833,162 @@ namespace llarp
...
+            "address.",
             "",
             "Examples:",
             "",
-            "    eth0=1090",
-            "    0.0.0.0=1090",
-            "    1.2.3.4=1090",
+            "    inbound=15.5.29.5:443",
+            "    inbound=10.0.2.2",
+            "    outbound=0.0.0.0:9000",
```
jagerman, line 870: Should add support for `inbound=:1234` that does the same go-find-an-ip behaviour, but overrides the default port to 1234 instead of 1090.

jagerman:
I'm thinking we may need a shortcut here to specify both incoming/outgoing IPs at once.  E.g. if you are on AWS (or some other provider that NATs) then you'll have some internal IP 172.16.2.3 and, I think, you'll need something like this:

```ini
[bind]
inbound=172.16.2.3:1090
outbound=172.16.2.3:1091
```
because, if I'm understanding correctly, the defaults won't be able to find anything because you don't have any public IPs.

But the duplication of needing to list the IP twice is annoying (mildly, but still).

A couple possibilities for thought:
- if outbound is unspecified and we can't find a public IP, try the inbound IP if one is specified
- add a `ip` to the `[bind]` setting that sets the default for inbound/outbound options, so you could do:

```ini
[bind]
ip=172.16.2.3
```
for a default-port, but special IP, or:
```ini
[bind]
ip=172.16.2.3
inbound=:1090
outbound=:1190
```
to set the ports as well.

majestrate:
to expand on this, i think it'd be wise to do a `public-ip` and ~~`bind-ip`~~ `listen-ip` split here, the previous option in router section makes little sense going forward it feels like now that we have a non idiotic config syntax here we can now do that.

llarp/config/config.cpp
majestrate: should be in now

PR #1937, opened 2022-06-16
Show router pubkey at startup

jagerman:
Can help diagnose lokinet/oxend sync issues when reviewing logs.


majestrate (approved):
i feel like we should add a log level in between INFO and WARN

PR #1938, opened 2022-06-16
Don't build empty cpp files

jagerman:
We shouldn't be compiling these .cpp files at all on other platforms,
rather than compiling empty .cpp files (which later results in "... has
no symbols" warnings).


majestrate:
it seems this needs rebase to merge

jagerman:
> it seems this needs rebase to merge

I've turned off that annoying setting on the repo.  It's sometimes useful, but more often just a nuisance.

"Oh no!  I changed the readme!  Have to rebase all my PRs now!"

majestrate (approved):

PR #1941, opened 2022-06-22
Remove dead code, cmake cleanups

jagerman:
- Remove shell hooks.  Unused, probably bitrotten, and not wanted anymore.
- Remove netns code.  Unused, bitrotten, and we can pull it out of git history if we want to resurrect it.
- Remove '=1' from compiler definitions (because we only check for definedness, not value, and someone might get the impression that they can set it to 0 to disable it, which they can't).


majestrate (approved):

llarp/service/endpoint.hpp
```diff
@@ -7,6 +7,7 @@
 #include <llarp/path/path.hpp>
 #include <llarp/path/pathbuilder.hpp>
 #include "address.hpp"
+#include "endpoint_state.hpp"
```
majestrate: i wonder if we could not require this header and forward declare as much as we can.

jagerman: This was required because I removed the destructor, but that meant the implicit one needs to be in the header.  I've made the destructor explicit again and put a defaulted one in the .cpp, which solves this.

PR #1942, opened 2022-06-22
macOS system extension

jagerman:
~~Still needs:~~

- [x] Wiring up lokinet start/stop from the GUI
- [x] Script to package the GUI + sign everything


jagerman:
Now works.  Relies on https://github.com/oxen-io/lokinet-gui/pull/17

jagerman:
We have a new mac icon (added as an svg here and built during the build), but it still requires some tweaks (that can come in a separate PR though).

jagerman:
The way this now works is the following.  It is dumb, but this is macos and apparently that's how Apple wants it to be.
```
// If we are invoked with no arguments then exec the gui.  This is dumb, but there doesn't seem to
// be a nicer way to do this on Apple's half-baked platform because:
// - we have three "bundles" we need to manage: the GUI app, the system extension, and the Lokinet
//   app (this file) which loads the system extension.
// - if we embed the system extension directly inside the GUI then it fails to launch because the
//   electron GUI's requirements (needed for JIT) conflict with the ability to load a system
//   extensions.
// - if we embed Lokinet.app inside Lokinet-GUI.app and then the system extension inside Lokinet.app
//   then it works, but macos loses track of the system extension and doesn't remove it when you
//   remove the application.  (It breaks your system, leaving an impossible-to-remove system
//   extension, in just the same way it breaks if you don't use Finder to remove the Application.
//   Apple used to say (around 2 years ago as of writing) that they would fix this situation "soon",
//   but hasn't, and has stopped saying anything about it.)
// - if we try to use multiple executables (one to launch the system extension, one simple shell
//   script to execs the embedded GUI app) inside the Lokinet.app and make the GUI the default for
//   the application then Lokinet gets killed by gatekeeper because code signing only applies the
//   (required-for-system-extensions) provisioningprofile to the main binary in the app.
//
// So we are left needing *one* single binary that isn't the GUI but has to do double-duty for both
// exec'ing the binary and loading lokinet, depending on how it is called.
//
// But of course there is no way to specify command-line arguments to the default binary macOS runs,
// so we can't use a `--gui` flag or anything so abhorrent to macos purity, thus this nasty
// solution:
//   - no args -- exec the GUI
//   - `--start` -- load the system extension and start lokinet
//   - `--stop` -- stop lokinet
//
// macOS: land of half-baked implementations and nasty hacks to make anything work.
```

CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ project(lokinet
 if(APPLE)
     # Apple build number: must be incremented to submit a new build for the same lokinet version,
     # should be reset to 0 when the lokinet version increments.
-    set(LOKINET_APPLE_BUILD 0)
+    set(LOKINET_APPLE_BUILD 1)
```
majestrate, line 35: probably wise to wrap this in an `if(NOT LOKINET_APPLE_BUILD)` so you can override it when building

cmake/gui.cmake
```diff
@@ -18,20 +18,34 @@ if (BUILD_GUI)
...
     WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/gui")
 
   if(APPLE)
-    add_custom_target(copy_gui ALL
-      DEPENDS lokinet lokinet-extension lokinet-gui
-      # FIXME: we really shouldn't be building inside the source directory but this is npm...
-      COMMAND ${CMAKE_COMMAND} -E copy_directory
-        ${PROJECT_SOURCE_DIR}/lokinet-gui/release/mac/lokinet-gui.app
-        $<TARGET_BUNDLE_DIR:lokinet>
+    add_custom_target(assemble_gui ALL
```
majestrate, line 32: is using `ALL` here correct?

majestrate (approved):
other than the cmake ponderings this looks good to merge

CMakeLists.txt
jagerman: I thought about that, but the reason I didn't is that it's effectively like a version -- if you need to bump the apple build then that needs to get tracked (and possibly bumped again in the future), and when you bump the main version you want to reset this to 0 -- which is why this is closed to the main Lokinet versions in the first place.  I'm not seeing a case where you would want to bump it locally but *not* commit that into the repo.

cmake/gui.cmake
jagerman: Yes: that adds it to the "all" target, which is what gets built if you run `ninja` with no target.

PR #1945, opened 2022-07-01
v0.9.9 update

necro-nemesis:
Merged to v0.9.9 upstream/stable

changes to versioning and deletion of associated patch.
revised dns patch.
revised SPEC file - SPEC for changes.
built by CI and tested on native install of Fedora 35.


majestrate:
arm failed https://ci.oxen.rocks/oxen-io/lokinet/3387/2/3

necro-nemesis:
> arm failed https://ci.oxen.rocks/oxen-io/lokinet/3387/2/3

Building aarch64 is running into random issues building on CI.  Failing at various % states of completion where it previously completed. x86_84 builds reliably. Problems started to appear in the final cleanup of code which was to remove the build dependency of oxenmq-devel which is no longer a dependency with current flags to build oxenc and oxenmq during lokinet compilation in addition to not look for the dependency.

I suspect it's the compiler atm not the code.

majestrate:
okay let's see if ci is happy

necro-nemesis:
> okay let's see if ci is happy

https://builds.lokinet.dev/oxen-io/lokinet/fedora-35/rpm-fedora-35-20220702T142533Z-167d24ff0/

It appears you have the magic touch.

jagerman:
The builds are failing because they are running out of memory on ARM.  I think we need to lower concurrency to solve.

PR #1947, opened 2022-07-02
Fedora/34 v0.9.9 upgrade

necro-nemesis:
Merged to v0.9.9 upstream/stable

changes to versioning and deletion of associated patch.
revised dns patch.
revised SPEC file - SPEC for changes.


majestrate:
can we get @jagerman to do a once down of this to update it to 0.9.12 tag when we put that out?

necro-nemesis:
How far back in distros do you want to continue to support? We also didn't build for OpenSUSE or CentOS on v.0.9.11 and I didn't hear a public cry out for demand. I don't know if those are a "if you build it they will come" objective or just wait until people start asking for it. They build similarly to F35 but lack systemd-resolved F35 has thus presently require some manual post installation DNS tweaking to work.

majestrate:
for debian we support back to old stable, not sure what fedora's release cycles are bound by but i know centos is even more long term than debian.

necro-nemesis:
Fedora has 2 releases / year. 13 months of support for each release. I'd argue if Fedora no longer supports it themselves it's time to move on. OpenSUSE and CentOS appear to be considerably longer between releases.

necro-nemesis:
https://docs.fedoraproject.org/en-US/releases/eol/

jagerman:
> I'd argue if Fedora no longer supports it themselves it's time to move on.

Yup, that's our same policy for the short-lived Ubuntu releases (20.10, 21.04, 21.10, 22.10, etc.): we support them for the same 9 months that Ubuntu does.

PR #1955, opened 2022-07-16
Logging overhaul, fmt::format, and related

jagerman:
This replaces custom logging system with spdlog-based oxen logging.  This PR mainly replaces the backend logging with the spdlog-based system, but doesn't (yet) convert all the existing LogWarn, etc. to use the new format-based logging.

New logging statements will look like:

    llarp::log::warning(cat, "blah: {}", val);

where `cat` should be set up in each .cpp or cluster of .cpp files, as described in the [oxen-logging README](https://github.com/oxen-io/oxen-logging)

As part of spdlog we get fmt, which gives us nice format strings, where are applied generously in this PR.

Making types printable now requires two steps:
- add a ToString() method
- add this specialization:

      template <>
      constexpr inline bool llarp::IsToStringFormattable<llarp::Whatever> = true;

This will then allow the type to be printed as a "{}" value in a fmt::format string.  This is applied to all our printable types here, and all of the `operator<<` are removed.

This commit also:
- replaces various uses of `operator<<` to ToString()
- replaces various uses of std::stringstream with either fmt::format or plain std::string
- Rename some to_string and toString() methods to ToString() for consistency (and to work with fmt)
- Replace `stringify(...)` and `make_exception` usage with fmt::format (and remove stringify/make_exception from util/str.hpp).


jagerman:
Still todo:
- [x] macos support
- [x] get rid of the `.print(blah, -1, -1)` print layer

llarp/config/config.cpp
```diff
@@ -1097,26 +1100,27 @@ namespace llarp
         "type",
         DefaultLogType,
         [this](std::string arg) {
-          LogType type = LogTypeFromString(arg);
-          if (type == LogType::Unknown)
-            throw std::invalid_argument(stringify("invalid log type: ", arg));
+          auto type = log::type_from_string(arg);
```
majestrate: `log::type_from_string` should throw imo

llarp/dns/message.hpp
```diff
@@ -96,19 +96,23 @@ namespace llarp
...
     };
-
-    inline std::ostream&
-    operator<<(std::ostream& out, const Message& msg)
-    {
-      msg.print(out, -1, -1);
-      return out;
-    }
   }  // namespace dns
+
```
majestrate, line 107: should be outside the llarp namespace

llarp/net/address_info.hpp
```diff
@@ -55,23 +55,23 @@ namespace llarp
...
-    return a.print(out, -1, -1);
-  }
-
   bool
   operator==(const AddressInfo& lhs, const AddressInfo& rhs);
 
   bool
   operator<(const AddressInfo& lhs, const AddressInfo& rhs);
 
+  template <>
```
majestrate, line 69: move out of llarp namespace?

llarp/net/sock_addr.hpp
```diff
@@ -60,9 +61,9 @@ namespace llarp
     SockAddr&
     operator=(const in6_addr& addr);
 
-    operator const sockaddr*() const;
-    operator const sockaddr_in*() const;
-    operator const sockaddr_in6*() const;
+    explicit operator const sockaddr*() const;
+    explicit operator const sockaddr_in*() const;
```
majestrate, line 65: i like this change

llarp/service/info.hpp
```diff
@@ -101,11 +104,8 @@ namespace llarp
...
-    inline std::ostream&
-    operator<<(std::ostream& out, const ServiceInfo& i)
-    {
-      return i.print(out, -1, -1);
-    }
   }  // namespace service
 }  // namespace llarp
+
+template <>
+constexpr inline bool llarp::IsToStringFormattable<llarp::service::ServiceInfo> = true;
```
majestrate, line 108: wait... so does this template specialization belong inside or outside the llarp namespace?

llarp/config/config.cpp
jagerman: It throws now (as does `level_from_string`).

majestrate (approved):

PR #1957, opened 2022-07-16
synthesize tcp reject

majestrate:
send tcp reject for tcp packets when we have no mapped addresses so that the connections close.

when switching the exit on and off we want this to happen so that existing connections will be broken.

### DIFF
--- a/llarp/net/ip_packet.hpp
+++ b/llarp/net/ip_packet.hpp
@@ -253,9 +253,11 @@ namespace llarp::net
     void
     ZeroSourceAddress(std::optional<nuint32_t> flowlabel = std::nullopt);
 
-    /// make an icmp unreachable reply packet based of this ip packet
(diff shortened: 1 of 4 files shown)


llarp/net/ip_packet.hpp
```diff
@@ -253,9 +253,11 @@ namespace llarp::net
     void
     ZeroSourceAddress(std::optional<nuint32_t> flowlabel = std::nullopt);
 
-    /// make an icmp unreachable reply packet based of this ip packet
+    /// make an ip packet that will close or reject whatever upper layer sent it
+    /// makes a tcp rst packet for tcp, otherwise an icmp unreachble packet
+    /// returns nullopt if not implemented or applcable for this packet
```
tewinget: `s/applcable/applicable/`

PR #1958, opened 2022-07-16
add documentation on project structure

majestrate:
adds a bunch of readmes about the current structure of the project


external/readme.md
```diff
@@ -0,0 +1 @@
+directory for git submodules
```
tewinget, line 1: todo: add a list of submodules with a brief (or detailed, if it feels necessary, e.g. "why this instead of that?" if you want) description of each and why they're included in the project / what purpose they serve it.

contrib/readme.md
```diff
@@ -0,0 +1,75 @@
...
+
+linux cross compile build shim:
+
+* cross.sh
+
+apply multiple patches script:
+
+* apply-patches.sh
+
+bencode pretty-print script:
```
tewinget: I'd put something like:

`script for pretty-printing [Bittorrent Encoding](insert link here) (bencode) encoded data`

contrib/readme.md
```diff
@@ -0,0 +1,75 @@
...
+
+clang-format / jsonnetfmt / swiftformat helper, will check or correct code style:
+
+* format.sh
+
+git hooks:
+
+* git-hook-pre-push.sh
+
+base16 to base32z converter
```
tewinget: maybe include a link to a base32z spec page?

tewinget: also apparently (if wikipedia is correct, which is of course always true) it is actually called `z-base-32`?

tewinget (approved):
seems fine for a first pass, obviously will improve with time

PR #1959, opened 2022-07-16
clean up sntrup code

majestrate:
* add readme with explanations of what this lib is and why it is
* remove broken-ish avx2 version
* clean up and consolidate internal includes


jagerman:
> remove broken-ish avx2 version

Have done some testing on this, and can't find anything broken in either mode.  (Also this code is confirmed as active on pretty much all of our amd64 builds whenever you have a modern CPU).

PR #1961, opened 2022-07-18
re enable rc expiration

### LINKED ISSUE #1956: Don't load very old nodedb entries
When launching lokinet recently with an ancient nodedb in my home directory (because I rarely run it that way) my lokinet was completely unable to get onto the network.

We should avoid this, by dropping any old nodedb entries when we read the nodedb that have timestamps older than a certain threshold.

Perhaps 30d on the threshold?

In theory, if loading from an ancient directory where they are all >30d old, that should make us bootstrap again.

majestrate:
fixes #1956

### DIFF
--- a/llarp/nodedb.cpp
+++ b/llarp/nodedb.cpp
@@ -123,7 +123,7 @@ namespace llarp
         if (fs::is_regular_file(f) and f.extension() == RC_FILE_EXT)
         {
           RouterContact rc{};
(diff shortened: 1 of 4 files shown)


jagerman:
Tested this against my ancient nodedb that couldn't get on the network at all: this indeed fixes things.

One change though: we should remove valid-but-expired RCs during startup: currently we just skip them, but that ends up leaving behind a bunch of expired RCs forever.

jagerman (approved):

PR #1962, opened 2022-07-18
DNS message parsing fixes and cleanup

jagerman:
*NB: Builds on #1955 -- only the last commit is relevant to this PR.*

### Fixes:

- tighten reserved name detection to not match fooloki.loki, but instead only match "foo.loki.loki" and "loki.loki" (and similar for reserved name "snode.loki").
- IPv6 PTR parsing was completely broken.
- Added tests for the above two issues.

### Cleanups:

- Eliminate llarp::dns::Name_t typedef for std::string
- Use optional return instead of bool + output param
- Use string_views; we were doing a *lot* of string substr's during parsing, each of which allocates a new string.
- Use fmt instead of stringstream
- Simplify IPv4 PTR parsing

### DIFF
--- a/llarp/dns/message.cpp
+++ b/llarp/dns/message.cpp
@@ -218,7 +218,7 @@ namespace llarp
         rec.ttl = ttl;
         std::array<byte_t, 512> tmp = {{0}};
         llarp_buffer_t buf(tmp);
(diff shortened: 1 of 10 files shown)


majestrate:
lint

PR #1964, opened 2022-07-21
Fix inbound IP handling; update [bind] docs

### LINKED ISSUE #1929: Binding to wrong IP

jagerman:
- Fixes #1929 issue with `inbound=:1234` not defaulting to an auto-detected public IP
- Updates/rearranges some `[bind]` section config documentation


majestrate:
Outbound links should NOT use wildcard, that is the original bug that was being fixed here. Outbound listen ip matches any inbound listen ip.

jagerman:
> Outbound links should NOT use wildcard

This is not at all what was implemented in #1930, either in the config descriptions or the code.  The config says it defaults to wildcard, and indeed if you don't have any outbound at all, it defaults to wildcard.

jagerman:
I vaguely recall that we had a conversation about this a couple weeks ago, and decided it was probably still okay to use a wildcard for outgoing because, since we are the initiator, we ought to end up using a consistent source IP address anyway.

majestrate:
The issue was on boxes with multiple ip per interface wildcard did not always choose source correctly so we decided to match the listen ip on both inbound and outbound to make sure it does

jagerman:
> The issue was on boxes with multiple ip per interface wildcard did not always choose source correctly so we decided to match the listen ip on both inbound and outbound to make sure it does

The main issue was with incoming: if I'm listening on 1.2.3.4:5678, but my default source address is 1.2.3.5 then my reply to a packet sent to that 1.2.3.4 address will have unexpected source address 1.2.3.5, which caused problems at the other end.

For outbound, it's less clear to me that the same issue exists: I'm going to send a packet to some remote, the system will put the 1.2.3.5 source address on it, and that packet goes out.  A response to that packet directed at 1.2.3.5 will reach me and I'll be fine with it, and if I respond to *that* one, it ought to still use the 1.2.3.5 address, since that's my default routing IP.

jagerman:
(It's also rather unclear to me how a lokinet client, typically behind NAT, could reasonably have anything other than a wildcard default).

majestrate (approved):

PR #1969, opened 2022-08-02
use wintun and windivert for windows platform bits

### LINKED ISSUE #710: Wintun for windows
This is more an enhancement suggestion than a bug squash request.
The OpenVpn tun driver for windows feels like somewhat dirty. The loki-network node clash with my custom CJDNS node OpenVpn tun interface for Windows.
I don't feel that this a bug in loki-network tun code, but I feel sorry that I can't code Wintun tun code for loki-nework (and for CJDNS too).
Wintun is much a beta code and fast evolving but feels right. https://www.wintun.net/

WARNING WARNING - onion routing fan cheers below -
I am a longtime TOR and I2P user and now I am a happy loki-network tester and (small) bandwidth contributor, thanks to the developers for this very interesting project.

### LINKED ISSUE #1887: Incorrectly setting DNS on lokitun0 on start
**Describe the bug**

1. Start lokinet via systemctl 

2. 
image

After starting lokinet, query the `exit.loki` and systemd-resolved sends DNS requests through lokitun0 to destination `127.3.2.1` which is rejected.

It could be my fault too, IDK

 - OS: Ubuntu 21.10, 5.13.0-37-generic
-  Device: PC
 - Lokinet Version number or Git commit hash: lokinet-0.9.8-deb0.9.8-1~ubuntu2110

### LINKED ISSUE #1927: remove lokinet-bootstrap tool
lokinet-bootstrap pulls in curl and a bunch of other stuff we do not use in anything else, it really only exists for windows at this point.

we can remove it and just bundle bootstraps on windows and for custom bootstrap let the user figure it out.

### LINKED ISSUE #1939: RPM patch requirements
RPMs for all distros are patching cmake/Version.cmake but latest changes require an update to the patch. Additionally some RPMs were experiencing issues with using port 53 which were patched with default-dns.patch. I began to work on revised patches see:  

https://github.com/necro-nemesis/lokinet/blob/rpm/patch/SOURCES/default-dns.patch

for updating default-dns.patch but am at a loss as to the requirements now needed to patch the current version of Version.cmake appropriately to build with the recent changes.

### LINKED ISSUE #1944: windows dns still leaks
windows leaks dns queries with exit enabled.

this still needs to be fixed.

### LINKED ISSUE #1953: 0.99 JavaScript error
屏幕截图 2022-07-15 010827

### LINKED ISSUE #1981: No Windows Binaries
According to the docs there should be windows binaries on lokinet.org and on the github release page, however there aren't actually any. Is this an accident or was windows support dropped?

majestrate:
fixes #710 
fixes #1887
fixes #1981 
fixes #1944
fixes #1953 
fixes #1927
fixes #1939


cmake/macos.cmake
```diff
@@ -150,7 +150,12 @@ if(BUILD_PACKAGE)
       "${lokinet_installer}"
       COMMAND ./seticon lokinet.icns "${lokinet_installer}.dmg"
   )
-  add_dependancies(package "${lokinet_installer}.dmg")
+  add_custom_target(apple ALL COMMAND "true" DEPENDS "${lokinet_installer}.dmg")
```
jagerman: Please don't.  The `package` target is reserved only when cpack is in use, which we do not use on Apple.  Using the `package` target gives us a target that works the same way whether using cpack (non-apple) or not (apple).

contrib/format.sh
```diff
@@ -21,7 +21,7 @@ fi
 
 cd "$(dirname $0)/../"
 if [ "$1" = "verify" ] ; then
-    if [ $($binary --output-replacements-xml $(find jni daemon llarp include pybind | grep -E '\.([hc](pp)?|mm?)$' | grep -v '\#') | grep '</replacement>' | wc -l) -ne 0 ] ; then
+    if [ $($binary --output-replacements-xml $(find jni daemon llarp include pybind | grep -E '\.([hc](pp)?|mm)$' | grep -v '\#') | grep '</replacement>' | wc -l) -ne 0 ] ; then
```
jagerman: Why don't we just let it be reformatted, by adding the `?` into the pattern 4 lines down?

contrib/ci/drone-static-upload.sh
```diff
@@ -50,9 +50,9 @@ elif [ -e build-docs ]; then
     archive="$base.tar.xz"
     cp -av build-docs/docs/mkdocs.yml build-docs/docs/markdown "$base"
     tar cJvf "$archive" "$base"
-elif [ -e build-mac ]; then
+elif [ -e build/macos ]; then
```
jagerman: Please don't.  I routinely use multiple `build*` directories, including `build` for manual builds.  Nesting it under `build/` is a nuisance in multiple ways.

contrib/mac.sh
```diff
@@ -9,14 +9,11 @@
 
 set -e
 set -x
+root="$(pwd)"
+build="${root}/build/macos"
 
-if ! [ -f LICENSE ] || ! [ -d llarp ]; then
-    echo "You need to run this as ./contrib/mac.sh from the top-level lokinet project directory"
-fi
```
jagerman, line 15: Why remove this check?

daemon/CMakeLists.txt
```diff
@@ -46,15 +46,27 @@ if(WITH_BOOTSTRAP)
...
+add_library(hax_and_shims_for_cmake INTERFACE)
+if(WIN32)
+  target_link_libraries(hax_and_shims_for_cmake INTERFACE uvw oxenmq::oxenmq -lws2_32 -lshlwapi -ldbghelp -luser32 -liphlpapi -lpsapi -luserenv)
+endif()
+
 foreach(exe ${exetargets})
-  if(WIN32 AND NOT MSVC_VERSION)
+  if(WIN32)
     target_sources(${exe} PRIVATE ${CMAKE_BINARY_DIR}/${exe}.rc)
+    target_compile_options(${exe} PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-fpermissive>)
```
jagerman: Dear christ why?

daemon/CMakeLists.txt
```diff
@@ -46,15 +46,27 @@ if(WITH_BOOTSTRAP)
   endif()
 endif()
 
+# cmake interface library for bunch of cmake hacks to fix final link order
+add_library(hax_and_shims_for_cmake INTERFACE)
+if(WIN32)
+  target_link_libraries(hax_and_shims_for_cmake INTERFACE uvw oxenmq::oxenmq -lws2_32 -lshlwapi -ldbghelp -luser32 -liphlpapi -lpsapi -luserenv)
+endif()
```
jagerman, line 53: This seems counterproductive; why not just link to all of these directly in the one single place that links to this interface?

cmake/macos.cmake
majestrate: deoending on the package target was breaking ci on mac

contrib/format.sh
majestrate: it then requires an almost full reformat of the existing .m compilation units

jagerman: Good.

daemon/CMakeLists.txt
majestrate: i need to remove this, artifact from old cmake hunks removed

daemon/CMakeLists.txt
majestrate: link order matters here

jagerman: Doesn't answer my question as to why inject a single-use INTERFACE library here.

majestrate: that is exactly what this is doing.

majestrate: it needs to do that. but in a very specific order

jagerman: You link to this interface in exactly one place, which is going to be equivalent to just linking to the same list of items rather than the interface at the point where you are linking to the interface.

majestrate: oxemq target link order was borked and i did not want to figure out how to fix it in an upstreamable way, i used this approach

jagerman: "I put in a dirty hack because I couldn't be bothered to fix it properly" does not make me happy.

llarp/config/ini.cpp
```diff
@@ -17,8 +17,8 @@ namespace llarp
   {
     {
       std::ifstream f(fname, std::ios::in | std::ios::binary);
-      if (!f.is_open())
-        return false;
+      if (not f.is_open())
+        throw false;
```
jagerman: eep!

llarp/dns/multi_platform.hpp
```diff
@@ -0,0 +1 @@
+
```
jagerman: Was this supposed to be added?

llarp/dns/null_platform.hpp
```diff
@@ -0,0 +1 @@
+
```
jagerman: Was this supposed to be added?

contrib/ci/drone-static-upload.sh
```diff
@@ -77,6 +80,10 @@ $mkdirs
 put $archive $upload_to
 SFTP
 
+if [ -e "$base-debug-symbols.tar.xz" ] ; then
+    sftp -i ssh_key -b - -o StrictHostKeyChecking=off drone@oxen.rocks <<<"put $base-debug-symbols.tar.xz $upload_to"
+fi
+
```
jagerman: Would be slightly cleaner (and marginally faster) to put this in a variable and stick it into the previous SFTP invocation, something like:
```
put_debug=
if [ -e "$base-debug-symbols.tar.xz" ] ; then
    put_debug="put $base-debug-symbols.tar.xz"
fi
```
...
```
sftp -i ssh_key -b - -o StrictHostKeyChecking=off drone@oxen.rocks <<SFTP
$mkdirs
put $archive $upload_to
$put_debug
SFTP
```

llarp/util/compare_ptr.hpp
```diff
@@ -16,4 +17,16 @@ namespace llarp
       return Compare()(left, right);
     }
   };
+
+  /// type for comparing weak_ptr
```
jagerman:
```suggestion
  /// type for comparing weak_ptr by value
```
(to disambiguate from std::owner_less, which gives comparison by owned pointer).

daemon/CMakeLists.txt
majestrate: it also does not make me happy, when deadlines are imposed these kinds of hacks becomes the norm. this is a consistent source of our cruft across the repos, do it right or do it twice.

llarp/dns/server.cpp
```diff
@@ -411,22 +411,22 @@ namespace llarp::dns
...
                 m_ctx.get(),
                 q.Name().c_str(),
                 q.qtype,
                 q.qclass,
-                (void*)pending,
+                tmp.get(),
                 &Resolver::Callback,
                 nullptr))
         {
           // take back ownership on fail
```
majestrate: remove comment

jagerman (approved):

jagerman (approved):

Nickck182:
Only hard work can make progress.

PR #1978, opened 2022-08-18
bundle consensus info into RC

### LINKED ISSUE #1960: Add staking info to rc
Some additional hints should be added to the lokinet rc including but not limited to.

- Hash of operator wallet
- hashes of contributing stake wallets (maybe include portion?)
- Timestamp of expected stake unlock when applicable

majestrate:
when we think we are going to become an invalid service node (e.g. stake unlock) put the estimated timestamp of when when think it will happen into the RC.

* add time and duration helpers
* make time interally use time points not unix stamps
* fix up unit tests to be a bit cleaner when using cryptomanager
* Router::ModifyOurRC able to be called from any thread by making it interally use EventLoop::call
* add AbstractRouter::InvalidAt to set and unset when our RC is going to be invalid and republish as needed

fixes #1960

### DIFF
--- a/llarp/crypto/crypto.hpp
+++ b/llarp/crypto/crypto.hpp
@@ -142,15 +142,7 @@ namespace llarp
     static Crypto*
     instance()
     {
-#ifdef NDEBUG
       return m_crypto;
-#else
-      if (m_crypto)
-        return m_crypto;
-
-      assert(false && "Cryptomanager::instance() was undefined");
-      abort();
-#endif
     }
   };
 
--- a/llarp/router/abstractrouter.hpp
+++ b/llarp/router/abstractrouter.hpp
@@ -3,7 +3,7 @@
 #include <llarp/config/config.hpp>
 #include <llarp/config/key_manager.hpp>
 #include <memory>
-#include <llarp/util/types.hpp>
+#include <llarp/util/time.hpp>
 #include <llarp/util/status.hpp>
 #include "i_outbound_message_handler.hpp"
 #include <vector>
@@ -130,6 +130,10 @@ namespace llarp
     /// published out
     virtual void
     ModifyOurRC(std::function<std::optional<RouterContact>(RouterContact)> modify) = 0;
+
+    /// mark that think we will be invalid in the future, or nullopt to reset this to normal
+    virtual void
+    InvalidAt(std::optional<llarp::TimePoint_t> maybe_future_time) = 0;
 
     virtual exit::Context&
     exitContext() = 0;
--- a/llarp/rpc/lokid_rpc_client.cpp
+++ b/llarp/rpc/lokid_rpc_client.cpp
@@ -44,11 +44,6 @@ namespace llarp
       // new block handler
       m_lokiMQ->add_category("notify", oxenmq::Access{oxenmq::AuthLevel::none})
           .add_command("block", [this](oxenmq::Message& m) { HandleNewBlock(m); });
-
-      // TODO: proper auth here
-      auto lokidCategory = m_lokiMQ->add_category("lokid", oxenmq::Access{oxenmq::AuthLevel::none});
-      lokidCategory.add_request_command(
-          "get_peer_stats", [this](oxenmq::Message& m) { HandleGetPeerStats(m); });
       m_UpdatingList = false;
     }
 
@@ -107,6 +102,46 @@ namespace llarp
       // don't upadate on block notification if an update is pending
       if (not m_UpdatingList)
         UpdateServiceNodeList();
+      // update our own node info, like for requested unlock
+      if (not m_UpdatingSelfInfo)
+        UpdateSelfInfo();
+    }
+
+    void
+    LokidRpcClient::UpdateSelfInfo()
+    {
+      m_UpdatingSelfInfo = true;
+      Request(
+          "rpc.get_service_node_status",
+          [self = shared_from_this()](bool success, std::vector<std::string> data) {
+            self->m_UpdatingSelfInfo = false;
+
+            if (not success)
+              return;
+            if (data.size() < 2)
+              return;
+            auto obj = nlohmann::json::parse(data[1]);
+            auto& j = obj["service_node_states"];
+            // we are not unlocking
+            if (j.count("requested_unlock_height") == 0)
+            {
+              if (auto r = self->m_Router.lock())
+                r->InvalidAt(std::nullopt);
+              return;
+            }
+
+            // guess how far away we are from unlock
+            const auto unlock_at = j["requested_unlock_height"].get<uint64_t>();
+            const auto blocks_from_unlock =
+                std::max(unlock_at, self->m_BlockHeight) - self->m_BlockHeight;
+
+            if (auto r = self->m_Router.lock())
+            {
+              // tell router we will be invalid in the future
+              r->InvalidAt(
+                  TimePoint_t{time_now() + (constants::block_interval * blocks_from_unlock)});
+            }
+          });
     }
 
     void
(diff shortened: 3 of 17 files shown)


llarp/crypto/crypto.hpp
```diff
@@ -139,18 +139,10 @@ namespace llarp
       m_crypto = m_prevCrypto;
     }
 
-    static Crypto*
+    inline static Crypto*
```
jagerman: No need for `inline` on a class method definition.

llarp/router/abstractrouter.hpp
```diff
@@ -131,6 +131,36 @@ namespace llarp
...
+    {
+      if (maybe_future_time)
+      {
+        // when we think we are invalid in the future we will publish slightly faster so we can keep
+        // the information as accurate as possible as the backend's guess ebbs and flows
+        ModifyOurRC([future_time = *maybe_future_time](auto rc) -> std::optional<RouterContact> {
+          // dont modify if we did it recently
+
+          if (auto dlt = time_now() - rc.last_updated_at();
+              dlt < (RouterContact::UpdateInterval / 2.5) and dlt > 0s)
```
jagerman: 2.5 is a very magic value.

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -107,6 +102,46 @@ namespace llarp
...
+            self->m_UpdatingSelfInfo = false;
+
+            if (not success)
+              return;
+            if (data.size() < 2)
+              return;
+            auto j = nlohmann::json::parse(data[1]);
+
+            // we are not unlocking
+            if (j.count("requested_unlock_height") == 0)
```
jagerman, line 126: This doesn't look right; I believe this endpoint returns the SN details under a `"service_node_state"` key.

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -107,6 +102,46 @@ namespace llarp
...
+            {
+              if (auto r = self->m_Router.lock())
+                r->InvalidAt(std::nullopt);
+              return;
+            }
+
+            // guess how far away we are from unlock
+            const auto unlock_at = j["requested_unlock_height"].get<uint64_t>();
```
jagerman, line 134: Likewise.

llarp/util/time.hpp
```diff
@@ -29,7 +25,64 @@ namespace llarp
...
+  time_now()
+  {
+    return DateClock_t::now();
+  }
+
+  /// convert a timepoint to a unix millis stamp
+  uint64_t
+  to_unix_stamp(const TimePoint_t& t);
+
+  /// get time right now as milliseconds, this is monotonic
```
jagerman: Comment is wrong; this clock isn't guaranteed to be monotonic.

llarp/util/time.hpp
```diff
@@ -29,7 +25,64 @@ namespace llarp
...
+      return distance_from<delta_res_t>(DateClock_t::now());
+    }
+
+    template <typename delta_res_t = std::chrono::milliseconds>
+    auto
+    distance_from(const TimePoint_t& other) const
+    {
+      return std::chrono::duration_cast<delta_res_t>(other - at);
+    }
   };
```
jagerman: I don't understand why this is a struct; why not simply two free functions for this?

llarp/util/time.hpp
```diff
@@ -29,7 +25,64 @@ namespace llarp
...
+
+  /// get a time delta with a timepoint in the future
+  template <typename Time_Duration>
+  TimePoint_t
+  future_time_point(Time_Duration future)
+  {
+    return DateClock_t::now() + future;
+  }
```
jagerman: Is this function really needed?  Why not just write `DateClock_t::now() + x` where you want this?

llarp/util/time.hpp
```diff
@@ -29,7 +25,64 @@ namespace llarp
...
+  {
+    return time_delta<Time_Duration>{future_time_point(future)};
+  }
+
+  /// returns the time now as a time point
+  inline TimePoint_t
+  time_now()
+  {
+    return DateClock_t::now();
+  }
```
jagerman, line 40: This function seems unnecessary.

llarp/util/time.hpp
```diff
@@ -29,7 +25,64 @@ namespace llarp
...
+  {
+    return Duration_t{to_unix_stamp(time_now())};
+  }
+
+  /// convert a legacy time stamp to a time point
+  template <typename Time_Duration = std::chrono::milliseconds>
+  auto
+  to_time_point(llarp_time_t stamp)
+  {
+    return future_time_point<Time_Duration>(stamp - time_now_ms());
```
jagerman: I think this is the same as: `return DateClock_t{stamp};`

llarp/util/time.hpp
jagerman: I think this was a struct before because when using duration types for both timestamps and durations we couldn't tell which was which, and so had to wrap things that weren't timestamps into a `time_delta`.  That seems unnecessary now.

llarp/router/abstractrouter.hpp
```diff
@@ -131,6 +131,36 @@ namespace llarp
...
     ModifyOurRC(std::function<std::optional<RouterContact>(RouterContact)> modify) = 0;
 
+    /// mark that think we will be invalid in the future, or nullopt to reset this to normal
+    inline void
+    InvalidAt(std::optional<TimePoint_t> maybe_future_time)
+    {
+      if (maybe_future_time)
+      {
+        // when we think we are invalid in the future we will publish slightly faster so we can keep
+        // the information as accurate as possible as the backend's guess ebbs and flows
```
jagerman: This faster publishing feels a bit like solving the issue the wrong way around: the blockchain time may be a bit inaccurate, but we should just have the client code be aware of that and simply stop using a router a couple hours before its unlock timestamp estimate so that the accuracy won't matter.

majestrate: fair.

llarp/util/time.hpp
majestrate: `DateClock_t` is not a time point type.

llarp/util/time.hpp
majestrate: it is

llarp/util/time.hpp
majestrate: no it is not needed.

llarp/util/time.hpp
majestrate: indeed.

llarp/util/time.hpp
majestrate: it needs to be, as the code using it relies on it being monotonic

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -107,6 +102,46 @@ namespace llarp
...
+            if (not success)
+              return;
+            if (data.size() < 2)
+              return;
+            auto obj = nlohmann::json::parse(data[1]);
+            auto& j = obj["service_node_states"];
+            // we are not unlocking
+            if (j.count("requested_unlock_height") == 0)
+            {
+              if (auto r = self->m_Router.lock())
```
majestrate, line 128: todo: put note here about when the gossip happens here

llarp/rpc/lokid_rpc_client.cpp
```diff
@@ -107,6 +102,46 @@ namespace llarp
...
+
+            // guess how far away we are from unlock
+            const auto unlock_at = j["requested_unlock_height"].get<uint64_t>();
+            const auto blocks_from_unlock =
+                std::max(unlock_at, self->m_BlockHeight) - self->m_BlockHeight;
+
+            if (auto r = self->m_Router.lock())
+            {
+              // tell router we will be invalid in the future
+              r->InvalidAt(
```
majestrate, line 141: todo: put note here about when the gossip happens here

PR #1983, opened 2022-08-30
Macos packaging

jagerman:
macOS packaging, part 2.

Continuation of #1942 that fixes/updates a few things in the macOS packaging:

- update GUI submodule to proper dev commit (rather than a forked branch commit)
- remove leftover scripts from the (very old) macos packaging
- fix up mac.sh to have better defaults for making a mac package
- build the package into a .dmg
- fix an issue with the lokinet swift wrapper (which basically just either loads the main extension, or loads the gui) that required macOS 12 instead of 10.15+ (like the rest of the package).
- #1942 accidentally deleted a needed cmake option to properly codesign + "provision profile" the network extension
- work around Apple's shitty OS bugs that adds yellow streaks into the PNGs embedded in .icns files


majestrate (approved):
if it compiles, shipit

PR #1991, opened 2022-09-19
Lokinet vpn cleanups

jagerman:
Relax rpc handling of token and range:
- Accept empty string or `null` for token to mean "no token."
- Accept `null` for range to mean "default range."
- Don't use a default range (::0/0) in lokinet-vpn because this will fail if IPv6 ranges aren't supported on the platform (e.g. on Windows), and isn't necessary: if we omit it then the rpc code already uses ::0/0 or 0.0.0.0/0 by default, as needed.

Various lokinet-vpn cleanups:
- rename LMQ to OMQ
- Add a function to extract a value from parsed options, to DRY out the argument parsing code a little bit.
- Add a exit_error function to format a message to stdout and then return the code, to simplify the repeated print-and-return code used when errors occur.
- Use fmt for output formatting
- Add an error if multiple modes are specified at once (--up/--down/--status/--exit)
- Add error printing around unmap RPC request

### DIFF
--- a/daemon/lokinet-vpn.cpp
+++ b/daemon/lokinet-vpn.cpp
@@ -16,11 +16,11 @@
 #include <sys/wait.h>
 #endif
 
-/// do a oxenmq request on an lmq instance blocking style
+/// do a oxenmq request on an omq instance blocking style
 /// returns a json object parsed from the result
 std::optional<nlohmann::json>
-LMQ_Request(
-    oxenmq::OxenMQ& lmq,
+OMQ_Request(
+    oxenmq::OxenMQ& omq,
     const oxenmq::ConnectionID& id,
     std::string_view method,
     std::optional<nlohmann::json> args = std::nullopt)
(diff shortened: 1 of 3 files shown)


daemon/lokinet-vpn.cpp
```diff
@@ -95,69 +140,48 @@ main(int argc, char* argv[])
...
-    return 1;
+    return exit_error(2, "{}", ex.what());
   }
+
+  int num_commands = goUp + goDown + printStatus + killDaemon;
+
+  if (num_commands == 0)
+    return exit_error(3, "One of --up/--down/--status/--kill must be specified");
+  if (num_commands != 1)
+    return exit_error(3, "Only one of --up/--down/--status/--kill may be specified");
```
tewinget, line 169: This could just check for `!= 1` and use error "Exactly one of ... must be specified"

jagerman: Technically correct but...

tewinget (approved):

Issue #1995, opened 2022-09-27
Lokinet router failing to communicate after service node re-registration


jagerman:
A couple of times now I've had a service node get decommissioned due to Lokinet unreachability after an unlock-and-reregister. Connectivitity seems fine after a restart of lokinet.

Timeline:

    Lokinet-router is running fine as a registered node
    Registration expires
    SN gets re-registered (without any restarts)

Up to the re-registration we have, every 30 seconds, the expected whining:

```
Sep 26 19:46:07 skoll lokinet-router-27[1049811]: [ERR] [](53) 2022-09-26 22:46:07.424 GMT [+621h18m16.693s]
    ../llarp/router/router.cpp:1027        We are running as a service node but we seem to be deregistered
```

which is fine, we aren't a registered service node. These stop after the re-registration, there are no log statements at all for the next 5.5 hours, then we got deregistered because of failing lokinet connectivity.  Other nodes at this time were reporting that Lokinet is unreachable; there are no log statements (logging is at warning level) for 5.5 hours, then a decomm, a recomm, and then a dereg.

Lokinet appears to track the decomm/recomm states fine, but isn't reachable by anyone else on the network until it gets restarted.

jagerman:
On further investigation it appears that this node has an outdated bootstrap file.  So the connectivity issue makes sense, but there really needs to be a serious warning when the nodedb is close to empty and we are active.

ianmacd:
> Lokinet appears to track the decomm/recomm states fine, but isn't reachable by anyone else on the network until it gets restarted.

Yes, I narrowly escaped a deregistration once, due to this bug. I also know a couple of other ops who weren't so lucky.

Now, when I bring up a shared node and I don't know when exactly it will finally go live on the network, I insulate myself from this bug with the following job in `/etc/crontab`.

`*/15 * * * *	root	systemctl restart lokinet-router`

jagerman:
TODO:
- identify invalid outdated bootstrap
- add system default bootstrap config option
- ignore the known-invalid old bootstrap file (the one in question here was a little over two years old, and hasn't been valid for a long time).
- update debs to install and use a system default bootstrap if no (valid) bootstrap is otherwise configure that can get upgraded with the package (by not being in the config directory).
- add "error" level warning if we are in a state where we are an active service node but have too few known routers (e.g. less than 5).

majestrate:
did this get fixed?

PR #1996, opened 2022-09-27
Fallback bootstrap router build parameter, ignore obsolete bootstrap routers

tewinget:
Addresses #1995 .


llarp/router_contact.cpp
```diff
@@ -532,6 +532,21 @@ namespace llarp
     return false;
   }
 
+  bool
+  RouterContact::IsObsoleteBootstrap() const
+  {
+    constexpr std::array obsolete_bootstraps = {
```
jagerman: Should move this outside of the function (and make it static constexpr)

llarp/router/router.cpp
```diff
@@ -666,39 +666,24 @@ namespace llarp
...
+    // that no longer exists
+    for (auto rc_itr = b_list.begin(); rc_itr != b_list.end();)
+    {
+      if (rc_itr->IsObsoleteBootstrap())
+        b_list.erase(rc_itr);
+      else
+        rc_itr++;
+    }
+
     for (auto& rc : b_list)
```
jagerman: This for loop can be DRYed out a bit with the nearly identical loop below

llarp/router/router.cpp
```diff
@@ -711,6 +696,24 @@ namespace llarp
 
     if (bootstrapRCList.empty() and not conf.bootstrap.seednode)
     {
+#ifdef BOOTSTRAP_FALLBACK
```
jagerman: I would rather keep the `ifdef` as small as possible to minimize code compilation differences, perhaps via this outside the function:
```C++
static constexpr std::optional<std::string_view> bootstrap_fallback_path{
#ifdef BOOTSTRAP_FALLBACK
    std::string_view{BOOTSTRAP_FALLBACK}
#else
    std::nullopt;
#endif
};
```
Then the code here can do (without an ifdef):
```C++
if (bootstrap_fallback_path)
{
  ...
}
```

llarp/router/router.cpp
```diff
@@ -666,52 +666,57 @@ namespace llarp
...
 
     if (bootstrapRCList.empty() and not conf.bootstrap.seednode)
     {
-      throw std::runtime_error{"we have no bootstrap nodes"};
+      if (not bootstrap_fallback.empty())
+      {
+        b_list.clear();
+        b_list.AddFromFile(BOOTSTRAP_FALLBACK);
+
+        verifyRCs(false);
```
jagerman:
```suggestion
        verifyRCs(true);
```

llarp/router/router.cpp
```diff
@@ -666,52 +666,57 @@ namespace llarp
...
+      for (auto& rc : b_list)
       {
-        LogWarn("ignoring invalid RC: ", RouterID(rc.pubkey));
-        continue;
+        if (not rc.Verify(Now()))
+        {
+          LogWarn("ignoring invalid RC: ", RouterID(rc.pubkey));
+          continue;
+        }
+        if (ignoreObsolete and rc.IsObsoleteBootstrap())
```
jagerman: Should ~~LogWarn~~ `log::warning(logcat, ...)` here as well to let the admin know there is something being skipped in the bootstrap file.

llarp/router/router.cpp
```diff
@@ -1032,14 +1037,27 @@ namespace llarp
...
+        m_NextDecommissionWarn = now + DecommissionWarnInterval;
+      }
+      else if (isSvcNode)
+      {
+        constexpr int KnownPeerWarningThreshold = 5;
+        if (nodedb()->NumLoaded() < KnownPeerWarningThreshold)
+          LogError(
+              "We appear to be an active service node, but have fewer than ",
+              KnownPeerWarningThreshold,
+              " known peers.");
```
jagerman:
```suggestion
          log::error(logcat,
              "We appear to be an active service node, but have fewer than {} known peers", KnownPeerWarningThreshold);
```

llarp/router/router.cpp
```diff
@@ -666,52 +666,60 @@ namespace llarp
...
+    }
+
+    auto verifyRCs = [&]() {
+      for (auto& rc : b_list)
       {
-        LogWarn("ignoring invalid RC: ", RouterID(rc.pubkey));
-        continue;
+        if (rc.IsObsoleteBootstrap())
+        {
+          LogWarn("ignoring obsolete boostrap RC: ", RouterID(rc.pubkey));
```
jagerman:
```suggestion
          log::warning(logcat, "ignoring obsolete boostrap RC: {}", RouterID(rc.pubkey));
```

llarp/bootstrap.cpp
```diff
@@ -36,4 +36,35 @@ namespace llarp
...
+    }
+    else
+    {
+      RouterContact rc;
+      if (not rc.Read(fpath))
+      {
+        throw std::runtime_error{
+            fmt::format("failed to decode bootstrap RC, file='{}', rc={}", fpath, rc)};
+      }
+      this->insert(rc);
```
majestrate, line 67: since you're here, we really need to not make BootstrapList a subtype of std::vector

jagerman: ... the fuck?

majestrate: my bad. std::set

jagerman: Should be done, but seems a bit out of scope for this PR.

tewinget: std::set, and while that is something that would be good it's out of scope for this PR

majestrate (approved):

jagerman:
One more thing:

Let's make lokinet-router *not* ping oxend if it's active and doesn't know of at least (threshold) number of routers, because that generally indicates some sort of serious connectivity issue and we want the operator to notice right away (because of a lack of proofs).  (We *do* still want it to ping oxend while unfunded, though, because it's common for an unfunded router to not know about any routers).

jagerman (changes requested):
Two changes needed here:

- the "fallback" should just bake the bootstrap data into the binary at compile time rather than carrying a path around that we have to distribute.
- we need to support bootstraps for different networks, typically mainnet + testnet, so that compile time fallback should probably be a unordered_map of netid -> bootstrap data.

jagerman (approved):

PR #1997, opened 2022-09-28
omq rpc log subscription

tewinget:
respects whether RPC is enabled, removes the log sink otherwise

bumps oxen-mq and oxen-logging

### DIFF
--- a/llarp/rpc/rpc_server.cpp
+++ b/llarp/rpc/rpc_server.cpp
@@ -746,7 +746,7 @@ namespace llarp::rpc
     }
     else
     {
-      log::info(logcat, "Renewed logs subscription request from conn id {}@{}", m.conn, m.remote);
+      log::debug(logcat, "Renewed logs subscription request from conn id {}@{}", m.conn, m.remote);
       m.send_reply("ALREADY");
     }
   }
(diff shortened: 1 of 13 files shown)


llarp/rpc/rpc_server.cpp
```diff
@@ -12,10 +12,17 @@
...
+namespace
+{
+  static auto logcat = llarp::log::Cat("lokinet.rpc");
+}  // namespace
 
 namespace llarp::rpc
 {
-  RpcServer::RpcServer(LMQ_ptr lmq, AbstractRouter* r) : m_LMQ(std::move(lmq)), m_Router(r)
+  RpcServer::RpcServer(LMQ_ptr lmq, AbstractRouter* r)
+      : m_LMQ(std::move(lmq)), m_Router(r), log_subs(*m_LMQ, llarp::logRingBuffer)
```
majestrate: curly brace initialization

llarp/util/logging.hpp
```diff
@@ -25,6 +26,8 @@ namespace llarp
 // Deprecated loggers (in the top-level llarp namespace):
 namespace llarp
 {
+  inline std::shared_ptr<log::RingBufferSink> logRingBuffer = nullptr;
```
majestrate, line 29: is there a good reason to have this be a "global"?

jagerman: We talked about this in more depth off github; the answer is that the logging system is inherently (already) using static globals because of how logging needs to done easily from anywhere: needing to carry around an instance to get to the logger makes logging much more painful.  As a result, the sink will already be held in static global variables; this is just storing another pointer to it here because it's a bit awkward to extract it from the main logging system internals.

jagerman (approved):

PR #2002, opened 2022-10-03
Gui fix hardcoded path

Bilb:
made this include changes from  https://github.com/oxen-io/lokinet/pull/1997


jagerman:
Closing (the update here was included in #1997).

PR #2014, opened 2022-10-13
Fix multi-field ordering

jagerman:
Lots and lots of places in the code had broken < operators because they are returning something like:

    foo < other.foo or bar < other.bar;

but this breaks both the strict weak ordering requirements that are required for the "Compare" requirement for things like std::map/set/priority_queue.

For example:

    a = {.foo=1, .bar=3}
    b = {.foo=3, .bar=1}

does not have an ordering over a and b (both `a < b` and `b < a` are satisfied at the same time).

This needs to be instead something like:

    foo < other.foo or (foo == other.foo and bar < other.bar)

but that's a bit clunkier, and it is easier to use std::tie for tuple's built-in < comparison which does the right thing:

    std::tie(foo, bar) < std::tie(other.foo, other.bar)

(Initially I noticed this in SockAddr/sockaddr_in6, but upon further investigation this extends to the major of multi-field `operator<`'s.)

This fixes it by using std::tie (or something similar) everywhere we are doing multi-field inequalities.

The most significant implication of this looks to be that the OutboundMessage priority queue was not prioritizing correctly: messages would be prioritized by *either* msgID or priority, and so we could get fairly random prioritization rather than prioritizing by priority first, and then msgID only as a fallback among equal-priority messages.

### DIFF
--- a/llarp/net/net.hpp
+++ b/llarp/net/net.hpp
@@ -34,62 +34,16 @@
 
 namespace llarp
 {
-  inline bool
-  operator==(const in_addr& a, const in_addr& b)
+  inline int
+  cmp(const in_addr& a, const in_addr& b)
   {
-    return memcmp(&a, &b, sizeof(in_addr)) == 0;
+    return memcmp(&a, &b, sizeof(in_addr));
   }
 
-  inline bool
-  operator==(const in6_addr& a, const in6_addr& b)
+  inline int
+  cmp(const in6_addr& a, const in6_addr& b)
   {
-    return memcmp(&a, &b, sizeof(in6_addr)) == 0;
-  }
-
-  inline bool
-  operator==(const sockaddr_in& a, const sockaddr_in& b)
-  {
-    return a.sin_port == b.sin_port and a.sin_addr.s_addr == b.sin_addr.s_addr;
-  }
-
-  inline bool
-  operator==(const sockaddr_in6& a, const sockaddr_in6& b)
-  {
-    return a.sin6_port == b.sin6_port and a.sin6_addr == b.sin6_addr;
-  }
-
-  inline bool
-  operator==(const sockaddr& a, const sockaddr& b)
-  {
-    if (a.sa_family != b.sa_family)
-      return false;
-    switch (a.sa_family)
-    {
-      case AF_INET:
-        return reinterpret_cast<const sockaddr_in&>(a) == reinterpret_cast<const sockaddr_in&>(b);
-      case AF_INET6:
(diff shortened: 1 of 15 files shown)


llarp/net/net.hpp
```diff
@@ -74,22 +74,35 @@ namespace llarp
...
+  inline int
+  cmp(const in_addr& a, const in_addr& b)
+  {
+    return memcmp(&a, &b, sizeof(in_addr));
+  }
+
+  inline int
+  cmp(const in6_addr& a, const in6_addr& b)
+  {
+    return memcmp(&a, &b, sizeof(in_addr));
```
tewinget: `sizeof(in6_addr)`?

llarp/net/sock_addr.cpp
```diff
@@ -206,8 +206,8 @@ namespace llarp
   bool
   SockAddr::operator<(const SockAddr& other) const
   {
-    return (m_addr.sin6_addr < other.m_addr.sin6_addr)
```
majestrate, line 209: why does this hunk need to be changed?

jagerman: Way more than just this need to be fixed as well, I am amending the PR.

llarp/net/uint128.hpp
```diff
@@ -141,25 +141,25 @@ namespace llarp
...
     {
-      return upper > b.upper || (upper == b.upper && lower > b.lower);
+      return std::tie(upper, lower) > std::tie(b.upper, b.lower);
     }
 
     constexpr bool
     operator>=(const uint128_t& b) const
     {
-      return upper > b.upper || (upper == b.upper && lower >= b.lower);
+      return std::tie(upper, lower) >= std::tie(b.upper, b.lower);
```
jagerman, line 162: (These changes aren't changing anything, but rather just making it consistent with all the other inequality comparisons in this PR).

llarp/service/intro.hpp
```diff
@@ -55,8 +55,8 @@ namespace llarp
       bool
       operator<(const Introduction& other) const
       {
-        return expiresAt < other.expiresAt || pathID < other.pathID || router < other.router
-            || version < other.version || latency < other.latency;
+        return std::tie(expiresAt, pathID, router, version, latency)
+            < std::tie(other.expiresAt, other.pathID, other.router, other.version, other.latency);
```
jagerman, line 59: I'm not certain that this is the right priority for all the fields here.

jagerman: I don't think the priority here matters as currently used; the only place that is currently using this is path/pathset.cpp, and it appears to only be using it for de-duplication, not (specifically) for ordering.

tewinget (approved):

PR #2018, opened 2022-10-17
add sysvinit init script provided by cheif36

majestrate:


jagerman:
This script doesn't look like it does anything to deal with the capabilities lokinet needs to run, so as-is I'm not sure how this is meant to work.  Is the admin expected to `setcap` the binary first?

majestrate:
> This script doesn't look like it does anything to deal with the capabilities lokinet needs to run, so as-is I'm not sure how this is meant to work. Is the admin expected to `setcap` the binary first?

this would also require a postint that would setcap the binary

PR #2019, opened 2022-10-18
Config setting documentation edits

### LINKED ISSUE #2012: Generated config is malformatted
On an upgrade to 0.9.10~rc1-1 the replacement lokinet.ini (generated by lokinet during post-inst) contains this text, which has multiple formatting problems:
```
# When in exit mode announce we allow a private range in our introsetexmaple:
# owned-range=10.0.0.0/24
# List of ip traffic whitelist, anything not specified will be dropped by us.examples:
# tcp for all tcp traffic regardless of port
# 0x69 for all packets using ip protocol 0x69udp/53 for udp port 53
# tcp/smtp for smtp port
# Specify a `.loki` address and an optional ip range to use as an exit broker.
# Example:
# exit-node=whatever.loki # maps all exit traffic to whatever.loki
# exit-node=stuff.loki:100.0.0.0/24 # maps 100.0.0.0/24 to stuff.loki
# Specify an optional authentication code required to use a non-public exit node.
# For example:
#     exit-auth=myfavouriteexit.loki:abc
# uses the authentication code `abc` whenever myfavouriteexit.loki is accessed.
# Can be specified multiple times to store codes for different exit nodes.
# Enable / disable automatic route configuration.
# When this is enabled and an exit is used Lokinet will automatically configure operating system routes to route traffic through the exit node.
# This is enabled by default, but can be disabled to perform advanced exit routing configuration manually.
#auto-routing=true
```

jagerman:
Rewords/reformats the documentation of various configuration options.

Fixes #2012

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -459,9 +459,8 @@ namespace llarp
         "owned-range",
         MultiValue,
         Comment{
-            "When in exit mode announce we allow a private range in our introset"
-            "exmaple:",
-            "owned-range=10.0.0.0/24",
+            "When in exit mode announce we allow a private range in our introset.  For example:",
+            "    owned-range=10.0.0.0/24",
         },
         [this](std::string arg) {
           IPRange range;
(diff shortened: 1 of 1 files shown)


majestrate (approved):
looks correct, if it has the desired format, shipit.

Issue #2029, opened 2022-10-24
cannot create keyfile for .loki keys


majestrate:
reported on windows and linux, creation of .loki privkeys fails on dev branch

PR #2030, opened 2022-10-24
llarp_buffer_t idiocy 

majestrate:
when read/writing a .loki privkey file we dont rewind a llarp_buffer_t after use. this is an argument in favor of just removing that type from the code entirely.

fixes by using 2 distinct locally scoped llarp_buffer_t, one for read, one for write.


majestrate:
fixes #2029

tewinget (approved):

PR #2035, opened 2022-10-26
try fixing weird drop in windows startup

majestrate:
call some dark magic for windows that "resets" the "ipv4" or something.


jagerman:
didn't work :(

majestrate:
ok

PR #2037, opened 2022-10-26
Use rsvg instead of imagemagick to make windows .ico; split up windows build

jagerman:
imagemagick is messing up the conversion, so just avoid it entirely and use rsvg-convert directly to do it instead.


majestrate (approved):

jagerman:
Without this, currently the installer icon looks like this:

image

which is the same crap macos was doing before rebuilding it to use rsvg.  (We can't do this same bypass on mac, though, because there we also have to use imagemagick to paint pixels in the corner to avoid the R Kelly effect).

majestrate:
also update the readme to note these new deps in the windows build

PR #2040, opened 2022-10-27
wait until actually stopped to tell windows we are

tewinget:
We should send STOP_PENDING rather than STOPPED while we aren't yet actually stopped; STOPPED is already sent when we actually finish stopping.

Also fixes some silly argc/argv shenanigans, I think.

TODO:
- [x] delay windows service notification until after we are fully initialized so that if we fail (e.g. can't start wintun, config error, etc.) we won't have told windows that we started (and it can properly report a failure to start the service).


daemon/lokinet.cpp
```diff
@@ -681,10 +681,12 @@ win32_daemon_entry(DWORD argc, LPTSTR* argv)
...
   // Report initial status to the SCM
   ReportSvcStatus(SERVICE_START_PENDING, NO_ERROR, 3000);
-  // SCM clobbers startup args, regenerate them here
-  argc = 2;
-  argv[1] = strdup("c:\\programdata\\lokinet\\lokinet.ini");
-  argv[2] = nullptr;
+  // SCM calls this function with different args than a normal "main" expects,
+  // but lokinet_main expects normal args, so set them here instead.  At the
+  // moment we are not passing any args to lokinet_main this way anyway though.
+  argc = 1;
```
majestrate: i would suggest using a totally separate array here so that we arent screwing with argc and argv

majestrate (approved):
if it works now it looks fine

llarp/linux/service_manager.cpp
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /// report our current state to the system layer
+    void
+    report_our_state() override
+    {
+      if (m_State == ServiceState::Starting)
+      {
+        // TODO: maybe request more time?
+        ::sd_notify(0, "READY=0");
```
jagerman: This isn't right; from sd_notify docs:

> Since there is little value in signaling non-readiness, the only value services should send is "READY=1" (i.e. "READY=0" is not defined).

llarp/linux/service_manager.cpp
```diff
@@ -0,0 +1,61 @@
+#include <llarp/util/service_manager.hpp>
+
+#if defined(WITH_SYSTEMD)
+#include <systemd/sd-daemon.h>
+#include <cassert>
+#endif
```
jagerman: I feel like this should get split into two separate files: A no-op version, and a systemd version and we conditionally compile base on whether we have systemd support.

llarp/linux/service_manager.cpp
```diff
@@ -1,13 +1,10 @@
 #include <llarp/util/service_manager.hpp>
```
jagerman: I think this file should be removed now?

llarp/router/router.cpp
```diff
@@ -874,6 +874,60 @@ namespace llarp
     m_LastStatsReport = now;
   }
 
+  std::string
+  Router::status_line()
+  {
+    std::string status;
+#if defined(WITH_SYSTEMD)
```
jagerman: Should nuke this ifdef guard; this status line doesn't depend on systemd anymore.

llarp/router/router.cpp
```diff
@@ -874,6 +874,60 @@ namespace llarp
...
+    auto out = std::back_inserter(status);
+    fmt::format_to(out, "WATCHDOG=1\nSTATUS=v{}", llarp::VERSION_STR);
```
jagerman: The return value here should be starting from "v", but not including the `WATCHDOG=1\nSTATUS=` prefix (which should be the responsibility of sd_service_manager to add).

llarp/linux/service_manager.cpp
majestrate: yes it should be removed

llarp/router/router.cpp
majestrate: done

llarp/router/router.cpp
majestrate: done

llarp/router/router.cpp
```diff
@@ -870,6 +874,58 @@ namespace llarp
     m_LastStatsReport = now;
   }
 
+  std::string
+  Router::status_line()
+  {
+    std::string status;
+    auto out = std::back_inserter(status);
+    fmt::format_to(out, "v{}", llarp::VERSION_STR);
```
jagerman: This appears to be the only place that uses `VERSION_STR` in the entire code base; should we just drop it entirely and use fmt here to join the numeric array like this?
```suggestion
    fmt::format_to(out, "v{}", fmt::join(llarp::VERSION, "."));
```

daemon/lokinet.cpp
```diff
@@ -84,9 +86,6 @@ install_win32_daemon()
     llarp::LogError("Cannot install service ", GetLastError());
     return;
   }
-  // just put the flag here. we eat it later on and specify the
-  // config path in the daemon entry point
-  StringCchCat(szPath.data(), 1024, " --win32-daemon");
```
jagerman, line 87: This seems wrong.  If removed here it seems there is nowhere that sets the service up with the `--win32-daemon` flag, but main() still relies on it?

daemon/lokinet.cpp
```diff
@@ -82,9 +84,6 @@ install_win32_daemon()
...
```
jagerman, line 87: This seems wrong.  If removed here it seems there is nowhere that sets the service up with the `--win32-daemon` flag, but main() still relies on it?

daemon/lokinet.cpp
majestrate: yea. i also dont know.

jagerman: Why did you remove it?

majestrate: it was using a function called `StringCchCat` from some windows header that does god knows what. i forget to figure out wtf it did and emulate the desired behavior.

tewinget:
supplanted by #2045

PR #2048, opened 2022-11-03
simplify llarp::service::Endpoint::ReadyToDoLookup()

majestrate:
previously we had a checking style function that passes in an optional defaulting to nullopt as a micro optimzation, this makes the code unnessarily obtuse.

simplify this by splitting up into 2 functions,
one for getting the unique endpoints and one for checking if the number of them is above the minimum.

### DIFF
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -972,7 +972,7 @@ namespace llarp
     bool
     Endpoint::ReadyToDoLookup(size_t num_paths) const
     {
-      // Currently just checks the number of paths, but could do more checks in the future. (jason)
+      // Currently just checks the number of paths, but could do more checks in the future.
       return num_paths >= MIN_ENDPOINTS_FOR_LNS_LOOKUP;
     }
 
--- a/llarp/service/endpoint.cpp
+++ b/llarp/service/endpoint.cpp
@@ -40,16 +40,10 @@
 #include <uvw.hpp>
 #include <variant>
 
-namespace
-{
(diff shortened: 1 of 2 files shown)


llarp/service/endpoint.cpp
```diff
@@ -992,14 +981,9 @@ namespace llarp
...
-      path::Path::UniqueEndpointSet_t paths;
-      ForEachPath([&paths](auto path) {
-        if (path and path->IsReady())
-          paths.insert(path);
-      });
-
+      auto paths = GetUniqueEndpointsForLookup();
       // not enough paths
-      if (not ReadyToDoLookup(paths.size()))
+      if (paths.size() < MIN_ENDPOINTS_FOR_LNS_LOOKUP)
```
jagerman: This should still be using the function, not duplicating the condition here.

llarp/service/endpoint.cpp
```diff
@@ -958,20 +952,28 @@ namespace llarp
...
-        numPaths = paths.size();
-      }
+      return IsReady() and ReadyToDoLookup(GetUniqueEndpointsForLookup().size());
+    }
 
-      return numPaths >= MIN_ENDPOINTS_FOR_LNS_LOOKUP;
+    bool
+    Endpoint::ReadyToDoLookup(size_t num_paths) const
+    {
+      // Currently just checks the number of paths, but could do more checks in the future. (jason)
```
jagerman:
```suggestion
      // Currently just checks the number of paths, but could do more checks in the future.
```

jagerman (approved):

PR #2055, opened 2022-11-15
Fix missing option names, make [lokid]:rpc required

jagerman:
Three changes:

1. At some point between 0.9.9 and 0.9.10 we removed the printing of option names when a value doesn't have a default, but this means the config is littered with things like:

    ```ini
        # This option sets the greater foo value.
    ```

    with no actual option name printed out when there is no default.

    This fixes it by always printing the option name in such a case, just with an empty value, e.g.:

    ```ini
        # This option sets the greater foo value.
        #big-foo=
    ```
    
2. This makes the `[lokid]:rpc` setting required, because we cannot actually start up in service node mode if it is missing.  The debs, which do have a sensible default, replace the rpc setting in the generated config in postinst (and broke because the commented-out field was missing, because of the above).

3. Default and Required don't make sense together, so we now fail to compile if used together.  Similarly for Required and Hidden.  The test code had a bunch of places that *did* use them together anyway: the behaviour of Default+Required was to write the uncommented default value into the config file, but we decided a long time ago that empty configs was our default so supporting that doesn't make sense.

### DIFF
--- a/llarp/config/definition.cpp
+++ b/llarp/config/definition.cpp
@@ -152,10 +152,8 @@ namespace llarp
       const std::string& section, std::vector<std::string> comments)
   {
     auto& sectionComments = m_sectionComments[section];
-    for (size_t i = 0; i < comments.size(); ++i)
-    {
-      sectionComments.emplace_back(std::move(comments[i]));
-    }
+    for (auto& c : comments)
+      sectionComments.emplace_back(std::move(c));
   }
 
   void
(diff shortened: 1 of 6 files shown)


llarp/config/definition.cpp
```diff
@@ -152,10 +152,8 @@ namespace llarp
       const std::string& section, std::vector<std::string> comments)
   {
     auto& sectionComments = m_sectionComments[section];
-    for (size_t i = 0; i < comments.size(); ++i)
-    {
-      sectionComments.emplace_back(std::move(comments[i]));
-    }
+    for (auto& c : comments)
+      sectionComments.emplace_back(std::move(c));
```
majestrate, line 156: i feel like move constructing here is a mistake and we should be doing a copy.

majestrate: no, nevermind i am wrong, comments is a local.

majestrate (approved):

kraniet888:
Hi

PR #2076, opened 2022-11-22
Fix Windows DNS issues

### LINKED ISSUE #2073: windows dns on 0.9.10 build is buggy
there is intermittent weirdness in dns replies on windows where the replies simply are not sent back correctly

tewinget:
Fixes a couple of bugs with Windows DNS handling:

- OS DNS config matching lokinet upstream config will work now

- Windivert `WINDIVERT_ADDRESS.Outbound` flag will now be set correctly, i.e. packet injection will be in the correct direction.

These two points together should fix #2073

Also adds a bunch of debug logging

### DIFF
--- a/llarp/win32/windivert.cpp
+++ b/llarp/win32/windivert.cpp
@@ -201,7 +201,7 @@ namespace llarp::win32
       }
 
       void
-      send_packet(Packet&& w_pkt) const
+      send_packet(Packet w_pkt) const
       {
         auto& pkt = w_pkt.pkt;
         auto* addr = &w_pkt.addr;
--- a/llarp/win32/windivert.cpp
+++ b/llarp/win32/windivert.cpp
@@ -14,6 +14,8 @@ extern "C"
 
 namespace
 {
+  using namespace oxen::log::literals;
+
   std::string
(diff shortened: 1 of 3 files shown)


llarp/win32/windivert.cpp
```diff
@@ -106,18 +192,29 @@ namespace llarp::win32
...
+
+        log::trace(logcat, "got packet of size {}B", sz);
+        log_windivert_addr(addr);
+
         return Packet{std::move(pkt), std::move(addr)};
       }
 
       void
-      send_packet(const Packet& w_pkt) const
+      send_packet(Packet&& w_pkt) const
```
majestrate: this does not need to be an R value reference. or any other reference for that matter.

llarp/win32/windivert.cpp
```diff
@@ -55,6 +57,90 @@ namespace llarp::win32
       WINDIVERT_ADDRESS addr;
     };
 
+    void
+    log_windivert_addr(const WINDIVERT_ADDRESS& addr)
```
majestrate: this really should be a to_string kind of function instead.

majestrate (approved):

PR #2089, opened 2022-12-12
Liblokinet fixes

tewinget:
draft PR to trigger Mac CI


llarp/quic/tunnel.cpp
```diff
@@ -538,7 +538,7 @@ namespace llarp::quic
           [this,
            after_path = std::move(after_path),
            pport = pport,
-           remote_addr = std::move(remote_addr)](auto maybe_remote) {
+           remote_addr = remote_addr](auto maybe_remote) {
```
jagerman: No need for the `x = x` here, just `x` does the same.

tewinget: yeah, fair point.  Was a lazy/quick fix, will amend.

majestrate:
perhaps making a quic packet buffer dumper type, which shows a full breakdown of what is in the quic packet like we do with hex viewing memory would help debug the issue we are having.

contrib/liblokinet/CMakeLists.txt
```diff
@@ -1,14 +1,13 @@
...
-
-add_executable(tcp_listen
-    tcp_listen.cpp)
-target_link_libraries(tcp_listen PUBLIC lokinet-embedded-api)
-
-add_executable(tcp_connect
-    tcp_connect.cpp)
-target_link_libraries(tcp_connect PUBLIC lokinet-embedded-api)
+
+option(LOKINET_SRC_ROOT "" "root directory with lokinet source code")
```
jagerman: `option` is only for TRUE/FALSE values

llarp/quic/endpoint.cpp
```diff
@@ -318,9 +333,14 @@ namespace llarp::quic
...
       {
         Connection& conn = **conn_ptr;
-        auto exp = ngtcp2_conn_get_idle_expiry(conn);
-        if (exp >= now_ts || conn.draining)
+        auto exp = ngtcp2_conn_get_expiry(conn);
+
+        // a bit of buffer on the expiration time in case the last call to
+        // ngtcp2_conn_get_expiry() returned ~0ms from now and the connection
+        // hasn't had time to handle it yet.  5ms should do.
+        if (exp >= (now_ts - 5'000'000) || conn.draining)
```
majestrate, line 341: make this `5ms` not `5'000'000`

llarp/quic/tunnel.cpp
```diff
@@ -231,21 +242,21 @@ namespace llarp::quic
         // destroy the whole thing.
         if (ct.conns.empty() and (not ct.tcp or not ct.tcp->active()))
         {
-          LogDebug("All sockets closed on quic:", port, ", destroying tunnel data");
+          log::debug(logcat, "All sockets closed on quic:{}, destroying tunnel data", port);
+          if (ct.close_cb)
```
majestrate, line 246: should we null out `ct.close_cb` after closing?

llarp/quic/tunnel.cpp
```diff
@@ -591,9 +606,34 @@ namespace llarp::quic
     auto conn = tunnel.client->get_connection();
 
     conn->on_stream_available = [this, id = row.first](Connection&) {
-      LogDebug("QUIC connection :", id, " established; streams now available");
+      log::debug(logcat, "QUIC connection :{} established; streams now available", id);
       if (auto it = client_tunnels_.find(id); it != client_tunnels_.end())
+      {
         flush_pending_incoming(it->second);
```
majestrate, line 612: maybe change to `this->flush_pending_incoming` to match the style of `this->close` below?

jagerman: Other way around; we should avoid `this->whatever` unless `whatever` won't work for some reason (e.g. needing to disambiguate some local variable).

jagerman: (Possibly that's why `this->close` was used instead of just `close`?)

majestrate: `close` is ambiguous with `::close` in this context

llarp/service/endpoint.cpp
```diff
@@ -1054,26 +1057,27 @@ namespace llarp
     {
       if (not NameIsValid(name))
       {
-        handler(ParseAddress(name));
+        log::warning(logcat, "\"{}\" is not a valid ONS name", name);
```
majestrate, line 1060: feel free to use single quotes in the error message:

` log::warning(logcat, "'{}' is not a valid ONS name", name); `

llarp/quic/tunnel.cpp
jagerman: Not needed; the very next line is removing the item from the container.

PR #2090, opened 2022-12-15
Fedora/36

necro-nemesis:
- Merge upstream/stable
- Bump SPEC file to ver 0.9.11 with dns patch removal and comment
- checkout specific submodules commits as follows (commits as per upstream/stable):
- - oxen-encoding to a869ae2b0152ad70855e3774a425c39a25ae1ca6
- - oxen-mq to ac6ef82ff6fd20437b7d073466dbef82a95a2173
- - cpr to f88fd7737de3e640c61703eb57a0fa0ce00c60cd

lokinet.ini required uncomment of 'no-resolvconf=' and value set to '1' to connect to .loki sites. Resolving is unstable when exits are set.


nosfirlabs:
merge when?

majestrate:
when annual leave ends for me (January 4) it'll get reviewed and merged by then

jagerman:
> merge when?

Right now.  I've pushed the rpms for Fedora 37 (based on this branch) and will be pushing the F36 rpms shortly.

jagerman:
> lokinet.ini required uncomment of 'no-resolvconf=' and value set to '1' to connect to .loki sites. Resolving is unstable when exits are set.

This should go into a new issue if still happens.  I don't *think* the `no-resolvconf` can actually do anything in Fedora as I don't think we ever shipped the resolvconf script in the Fedora branches that looks for that setting (it's a sort of fake setting in that lokinet ignores it and it only gets used in the resolvconf script).

PR #2104, opened 2023-01-03
add case for openwrt aarch64 in static deps

### LINKED ISSUE #2103: Support OpenWRT cross compiling of static dependencies
StaticBuild.cmake falls short with handling OpenWRT tuples when building openssl for various arches when using musl. When parsing "triplets" it is unable to effectively handle "${ARCH}-openwrt-linux-musl" when building on the SDK for RPi4. ARCH in this instance = aarch64 based on SDK target selection of RPi4. Could be other arch's... probably should be others arch's ;).

I was able to compile with the addition of:

  elseif(ARCH_TRIPLET STEQUAL aarch64-openwrt-linux-musl)
    #cross_compile_arm64
    set(openssl_arch_linux-aarch64)

to StaticBuild.cmake in the openssl section and successfully build to test this theory.

majestrate:


jagerman (approved):

necro-nemesis:
This can be similarly applied for building MIPS32 on OpenWRT. 

```
elseif(ARCH_TRIPLET STREQUAL mips-linux-gnu OR ARCH_TRIPLET STREQUAL mips-openwrt-linux-musl) 
    set(openssl_arch linux-mips32)
```

PR #2108, opened 2023-01-09
[internals refactor] docs and reorganize compilation units

majestrate:
begin piecewise refactor

rearrange compilation units in cmake to have unambiguous component separation, with documentation notes on what the responsibilities of each are.

the new library structure:

* creates `lokinet-base` interface library that sets up common include paths and compile flags
* makes `lokinet-amalgum` an interface library with everything in it.
* marks the `lokinet-service-deprecated-kitchensink` component (the thing we are splitting up)
* defines the desired component layout going forward, adds stub libraries but no code or headers
* very brief documentation of component relations

to get the refactor smaller a few concessions have been made:

* have all our headers be referenced in a consistent albeit dumb way (consistency is preferred)

these commits in order are:

1) adding high level plan docs
2) redoing header conversion
3) cmake restructure
4) clang format comment


majestrate:
blocked on review from @jagerman

CMakeLists.txt
```diff
@@ -49,8 +49,8 @@ endif()
 option(USE_AVX2 "enable avx2 code" OFF)
 option(USE_NETNS "enable networking namespace support. Linux only" OFF)
 option(NATIVE_BUILD "optimise for host system and FPU" ON)
-option(EMBEDDED_CFG "optimise for older hardware or embedded systems" OFF)
-option(BUILD_LIBLOKINET "build liblokinet.so" ON)
+option(BUILD_LIBLOKINET "same as WITH_EMBEDDED_LOKINET" ON)
```
jagerman: Why keep this?

docs/internals_overview.md
```diff
@@ -0,0 +1,382 @@
...
+
+the link layer has side effects that bubble up to upper layers lokinet has no ability to be aware of.
+the congestion strategy in lokinet happens as a coincidental side effect in the implementation detail of how link layer frames are transported.
+elevated packet loss causes retransmissions which will delay traffic to the next hop. 
+this behavior is desirable but because it is done at the link layer and not the onion layer, it will affect all paths going across that node.
+this behavior should be reimplemented on an upper level, propagating link layer session info upwards would help inform how and when to do this.
+
+iwp is a homegrown protocol that nothing else speaks, it needs to be replaced.
+DTLS + SCP would be a prime choice as it is what webrtc speaks, this would make a WASM port a possiblity.
+QUIC would be great for durable transmission of messages but the capabilities of it to transmit lossy data is unclear at this time.
```
jagerman: Not unclear: https://datatracker.ietf.org/doc/html/rfc9221

docs/internals_overview.md
```diff
@@ -0,0 +1,382 @@
...
+#### llarp::Router
+
+the contextualized god object is supposed to be `llarp::Context` but well over 90% of the functionality is done in `llarp::Router` and the former provides a psuedo wrapper with a semi-stable C++ api. this sucks, alot. this needs to be split up and refactored.
+
+#### the service endpoint amalgum
+
+this is the 1 ½ kitchen sink, a signle giant component that strattles the `llarp::handlers` and `llarp::service` that is in charge of the flow layer.
+none of the flow layer was designed before hand and this code has had 2 or 3 partial failed refactors laying inside of it.
+this code is the result of an additive organic dev process with zero design, foresight or rigor. 
+it is a jumbled clusterfuck of obtuse partially completed components that merge into itself, fractally, with ambiguous names of functions, types and methods, with no useful documentation of any kind. (minus occational comments peppered with explitives).
```
jagerman: Please ditch the hyperbole, keep this doc professional.

docs/refactor_notes.md
```diff
@@ -0,0 +1,139 @@
...
+
+### new interface types (1)
+
+the component to refactor is `llarp::service::Context`
+
+the types that will get a new interface type are:
+
+`llarp::service::Context` -> `llarp::service::I_ServiceContext`
+
+`llarp::service::Endpoint` -> `llarp::service::I_Endpoint`
```
jagerman: Is there ever going to be more than one implementer of these interfaces?  I feel like we already have too many abstract interfaces that serve little purpose right now because they only have (and likely only ever will have) one single implementation.

jagerman: Additionally, for performance reasons, we ought to avoid unnecessary virtual inheritance, particularly for anything that gets used in hot code: there is a run-time cost (albeit small) of virtual lookup table access.

docs/refactor_notes.md
```diff
@@ -0,0 +1,139 @@
...
+
+`llarp::handlers` namespace will be deprecated and no longer publicly exposed
+
+an amalgum interface for obtaining all platform specific details, like event loop, vpn platform, etc: `llarp::service::I_PlatformDetails`
```
jagerman: FWIW I'm not a fan of the "I" prefix, partly because there really is no such thing as a pure interface in C++ (i.e. it's purely convention).

Proper interface classes (by convention in C++, or concrete in other languages) are awkwardly restricted in many classes.  For instance, I could easily imagine this `I_PlatformDetails` having some protected member variables/functions that de-duplicate common functionality from platform code, but as soon as you have any member or non-pure-virtual base implementation it isn't really an interface anymore.

(And many of our current `I_` classes *do* violate this principle, because having base implementations or members is both common and useful).

docs/refactor_notes.md
```diff
@@ -0,0 +1,139 @@
...
+`llarp::path::PathSet` owns many `llarp::path::Path`, this association itself is held in `llarp::path::PathContext`.
+
+`llarp::path::PathSet` it is the original component that was ambigious in responsibilities and scope.
+it is the core pain point causing previous refactors to explode and fail to complete.
+this is mainly due to `llarp::path::PathSet` implementing the `llarp::routing::IMessageHandler` interface, this makes the boundary between components start to blur.
+**EVERY** other refactor was to refactor pathset, but all of them failed as it explodes in scope when you do that.
+we will have `llarp::path::PathSet` be deprecated but will be retained as a consession to limit the scope of the refactor.
+
+`llarp::routing::IMessageHandler` and below CAN NOT have their public interfaces modified at all.
+this includes ALL `llarp::path::*` types.
```
jagerman: Why not?

llarp/CMakeLists.txt
```diff
@@ -1,5 +1,21 @@
...
+)
+
+# libcrypt interface library to optionally use libcrypt if we can find it
+add_library(lokinet-libcrypt INTERFACE)
+pkg_check_modules(CRYPT libcrypt IMPORTED_TARGET)
+if(CRYPT_FOUND AND NOT CMAKE_CROSSCOMPILING)
+  add_definitions(-DHAVE_CRYPT)
+  target_link_libraries(lokinet-libcrypt INTERFACE PkgConfig::CRYPT)
+  message(STATUS "using libcrypt ${CRYPT_VERSION}")
+endif()
```
jagerman: I feel like this belongs in `external/`

docs/internals_overview.md
```diff
@@ -0,0 +1,382 @@
...
+* there is no internal documentation of any kind.
+
+### kitchen sinks
+
+* `llarp::Router` 
+* the service endpoint amalgum
+
+#### llarp::Router
+
+the contextualized god object is supposed to be `llarp::Context` but well over 90% of the functionality is done in `llarp::Router` and the former provides a psuedo wrapper with a semi-stable C++ api. this sucks, alot. this needs to be split up and refactored.
```
jagerman: Refactored how?

docs/internals_overview.md
```diff
@@ -0,0 +1,382 @@
...
+
+* clients are always initiators and never are the recipient.
+* Servers are able to be the initiator and a recipient.
+
+If a server connects to a server this is a linguistic ambiguity that **MUST** be resolved to properly disambiguiate clients connecting to a server and a server connecting to a server as these are a fundamentally different type of interaction.
+
+so when referring to alice (A) and bob (B):
+
+* alice is **always** the initiator.
+* bob is **always** the recipient.
```
jagerman: This section feels needlessly verbose.  Would it not be sufficient to say:
"Lokinet uses a peer-based architecture, not a client-server architecture.  Thus we refer to interactions solely in terms of initiators and recipients, not clients and servers."
and *just* that.  All the other stuff feels irrelevant to the rest of this doc (e.g. alice and bob are never mentioned again), and I don't understand the need to bother justifying so heavily not using client-server terminology.  It feels like you're trying to make an argument against someone who strongly wants to use client-server terminology but such a person doesn't exist?

CMakeLists.txt
majestrate: so i break less things in CI. (lazy)

docs/internals_overview.md
majestrate: gotcha

docs/internals_overview.md
majestrate: yup, this was cherrypicked from a local branch

docs/refactor_notes.md
majestrate: fair

docs/refactor_notes.md
majestrate: so a `WhateverThing_Base` as an abstract class would be better then?

docs/refactor_notes.md
majestrate: i need to rewrite this section

llarp/CMakeLists.txt
majestrate: maybe, it is an optional system dep rather than a submodule so i didnt put it in external

majestrate: moved to external.

docs/refactor_notes.md
```diff
@@ -0,0 +1,102 @@
...
+data is some datum we are onion routing that we would apply symettric encryption as needed before propagating to upper or lower layers.
+
+
+## Link Layer
+
+the link layer is responsbile for transmission of frames between nodes.
+this layer will handle queuing and congestion control between wire proto sessions between nodes.
+the link layer is will initate and recieve wire session to/from remote nodes.
+
+the link layer PDU is (data, src-router-id, dst-router-id) tuples.
```
jagerman, line 87: I think src and dst need not be routers?

jagerman: i.e. maybe "src-peer-id" / "dst-peer-id" would be better terminology (unless I'm misunderstanding where talking to clients goes in this stack).

CMakeLists.txt
jagerman: I feel like this will come back to bite us in the ass, let's nuke the old name.

docs/refactor_notes.md
majestrate: src and dst are router id as that is what we care about here, we dont care about how we send it to them that is the wire layer's responsibility. really what it is is, "send this frame to router X" where X is addressed by their ed25519 public key.

majestrate: the link layer would then pick a wire protocol session that is established and then transfer it.

CMakeLists.txt
majestrate: yea, okay

majestrate:
@jagerman looking to merge this today, needs approval.

llarp/CMakeLists.txt
```diff
@@ -197,27 +245,61 @@ add_library(lokinet-amalgum
...
+  link/link_manager.cpp
   router/outbound_message_handler.cpp
   router/outbound_session_maker.cpp
   router/rc_lookup_handler.cpp
   router/rc_gossiper.cpp
   router/router.cpp
   router/route_poker.cpp
+  rpc/lokid_rpc_client.cpp
+  rpc/rpc_server.cpp
+  rpc/endpoint_rpc.cpp
```
jagerman, line 251: I think we should split these off into a distinct `lokinet-rpc` library.

external/CMakeLists.txt
```diff
@@ -140,3 +140,12 @@ if(WITH_BOOTSTRAP)
...
 endif()
+
+# libcrypt interface library to optionally use libcrypt if we can find it
+add_library(lokinet-libcrypt INTERFACE)
+pkg_check_modules(CRYPT libcrypt IMPORTED_TARGET)
+if(CRYPT_FOUND AND NOT CMAKE_CROSSCOMPILING)
+  add_definitions(-DHAVE_CRYPT)
+  target_link_libraries(lokinet-libcrypt INTERFACE PkgConfig::CRYPT)
+  message(STATUS "using libcrypt ${CRYPT_VERSION}")
+endif()
```
jagerman, line 165: This is currently enable/disabling a feature based on the presence of a dev package on your system.  I think we should feature-gate this behind a cmake option, and if enabled, require libcrypt.  We probably want that disabled by default for any of:
- static build
- Windows
- Android

Also probably worth leaving a "TODO" here to add it to StaticBuild.cmake (i.e. we should do it but defer that for now).

llarp/CMakeLists.txt
```diff
@@ -242,65 +324,206 @@ add_library(lokinet-amalgum
...
+target_link_libraries(lokinet-context PUBLIC
+  lokinet-cryptography
+  lokinet-platform
+  lokinet-config
+  lokinet-peerstats
+  lokinet-layers
+  lokinet-consensus)
+
+target_link_libraries(lokinet-consensus PUBLIC lokinet-util)
+target_link_libraries(lokinet-consensus PRIVATE lokinet-context)
```
jagerman: You can combine these PUBLIC/PRIVATE linking into the same command which makes it a bit easier to read:
```cmake
target_link_libraries(lokinet-consensus
    PUBLIC lokinet-util
    PRIVATE lokinet-context)
```

llarp/CMakeLists.txt
```diff
@@ -242,65 +324,206 @@ add_library(lokinet-amalgum
...
-    oxenc::oxenc
+target_link_libraries(lokinet-cryptography PUBLIC lokinet-libcrypt)
+target_link_libraries(lokinet-cryptography PRIVATE lokinet-util)
+
+target_link_libraries(lokinet-platform PUBLIC lokinet-util)
+
+target_link_libraries(lokinet-layer-link PUBLIC lokinet-peerstats)
+
+target_link_libraries(lokinet-dht PUBLIC lokinet-nodedb lokinet-util)
+target_link_libraries(lokinet-dht PRIVATE lokinet-context)
```
jagerman: Likewise.

llarp/CMakeLists.txt
```diff
@@ -242,65 +324,206 @@ add_library(lokinet-amalgum
...
+
+target_link_libraries(lokinet-dns PUBLIC lokinet-platform lokinet-dns-platform)
+target_link_libraries(lokinet-dns PRIVATE lokinet-config)
+
+target_link_libraries(lokinet-nodedb PUBLIC lokinet-util lokinet-cryptography)
+target_link_libraries(lokinet-nodedb PRIVATE lokinet-platform)
+
+target_link_libraries(lokinet-util PRIVATE lokinet-nodedb lokinet-platform)
+# XXX: i hate this
+target_include_directories(lokinet-util PUBLIC ${PROJECT_SOURCE_DIR} ${PROJECT_SOURCE_DIR}/include)
```
jagerman: I think rather than making `lokinet-util` the "god" base target here (the the includes) we should instead of a `lokinet-common` (or `lokinet-base`?) INTERFACE library which carries the include path (and maybe other compiler flags) that every other lokinet-whatever need.

jagerman: Also I think with that change, most of the `PUBLIC` linking here (e.g. `lokinet-nodedb` -> `lokinet-cryptography`) can become `PRIVATE` instead.

jagerman (approved):
Left some comments; once those are addressed this looks good to merge.

majestrate:
there was so many parts repeated that i ended up adding 2 helper functions to do the linking.
this will need a quick review again tomorrow.

(i will push this in the morning)

llarp/CMakeLists.txt
```diff
@@ -242,65 +306,218 @@ add_library(lokinet-amalgum
...
+
+# helper function to link a library to lokinet-base, enable lto, add to lokinet-amalgum and then link to other libs
+function(lokinet_link_lib)
+  set(libname ${ARGV0})
+  message(DEBUG "created target: ${libname}")
+  list(POP_FRONT ARGV)
+  enable_lto(${libname})
+  target_link_libraries(${libname} PUBLIC lokinet-base ${ARGV})
+  target_link_libraries(lokinet-amalgum INTERFACE ${libname})
+endfunction()
```
jagerman: CMake functions can have arguments which would save the ARGV manipulation here; any argument passed beyond the listed ones then go into ${ARGN}.  So this would simplify a bit to:
```suggestion
function(lokinet_link_lib libname)
  message(DEBUG "created target: ${libname}")
  enable_lto(${libname})
  target_link_libraries(${libname} PUBLIC lokinet-base ${ARGN})
  target_link_libraries(lokinet-amalgum INTERFACE ${libname})
endfunction()
```

majestrate:
okay, ready for one final review

PR #2109, opened 2023-01-09
CLI Fixes

### LINKED ISSUE #2057: lokinet-vpn does not work with ipc sockets + ipc by default
when lokinet uses an ipc socket and we have the ability to access it just fine

`/usr/bin/lokinet-vpn --rpc ipc:///var/lib/lokinet/rpc.sock --down`

prints `std::bad_cast` and exits with status 2, instead of calling the rpc endpoint it needed to.

dr7ana:
### CLI Flags

Single-letter flags (-v, -h), will sometimes flip CLI::IncorrectConstruction error. Documentation on git and gitbook have conflicting statements as to which of the following is preferable:

`cli.add_flag("v,--verbose", options.verbose, "Verbose");`

`cli.add_flag("-v,--verbose", options.verbose, "Verbose");`

The first triggers error, but the second does not. Going forward, the second convention will be followed.

### CLI Logic

Fixed error generating .ini configuration files and logic handling

### IPC Socket

Fixes: #2057 (pending reproduction)

### DIFF
--- a/daemon/lokinet-vpn.cpp
+++ b/daemon/lokinet-vpn.cpp
@@ -108,11 +108,11 @@ main(int argc, char* argv[])
   command_line_options options{};
 
   // flags: boolean values in command_line_options struct
-  cli.add_flag("v,--verbose", options.verbose, "Verbose");
+  cli.add_flag("-v,--verbose", options.verbose, "Verbose");
   cli.add_flag("--up", options.vpnUp, "Put VPN up");
   cli.add_flag("--down", options.vpnDown, "Put VPN down");
   cli.add_flag("--status", options.printStatus, "Print VPN status and exit");
-  cli.add_flag("k,--kill", options.killDaemon, "Kill lokinet daemon");
+  cli.add_flag("-k,--kill", options.killDaemon, "Kill lokinet daemon");
 
   // options: string values in command_line_options struct
   cli.add_option("--exit", options.exitAddress, "Specify exit node address")->capture_default_str();
@@ -156,6 +156,10 @@ main(int argc, char* argv[])
     case 0:
       return exit_error(3, "One of --up/--down/--status/--kill must be specified");
     case 1:
+      break;
+    case 2:
+    case 3:
+    case 4:
       return exit_error(3, "Only one of --up/--down/--status/--kill may be specified");
     default:
       break;
--- a/daemon/lokinet-vpn.cpp
+++ b/daemon/lokinet-vpn.cpp
@@ -157,12 +157,8 @@ main(int argc, char* argv[])
       return exit_error(3, "One of --up/--down/--status/--kill must be specified");
     case 1:
       break;
-    case 2:
-    case 3:
-    case 4:
-      return exit_error(3, "Only one of --up/--down/--status/--kill may be specified");
     default:
-      break;
+      return exit_error(3, "Only one of --up/--down/--status/--kill may be specified");
   }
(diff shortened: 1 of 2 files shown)


majestrate (approved):

daemon/lokinet.cpp
```diff
@@ -34,11 +34,12 @@ namespace
     bool generate = false;
     bool router = false;
     bool force = false;
+    bool config = false;
     bool configOnly = false;
     bool overwrite = false;
 
     // string options
-    std::string configPath;
+    std::string configPath = "";
```
majestrate: this really should be an `fs::path`

`fs::path configPath{};`

dr7ana: I was attempting to follow the logic of the previous implementation:
- Original: [the path is extracted from 'result' and then stored in the fs::path 3 lines later](https://github.com/oxen-io/lokinet/blob/bf0dc52df7c05d4a14b5d5140615c1d547430055/daemon/lokinet.cpp#L454)
- Updated: path is stored in struct as string, and if it is updated then it is stored into the [fs::path here](https://github.com/drouhana/lokinet/blob/06b2c58a7dc35d05fe250bc428a00c13209cf0e1/daemon/lokinet.cpp#L397)

Would you like me to change it so that the struct stores the fs::path right off the bat?

jagerman: I don't think we should change it because, while CLI11 supports std::filesystem::path, but our `fs` is sometimes `ghc::filesystem` which wouldn't work on those platforms.

So perhaps leave a comment along the lines of:
```C++
// TODO: change this to use a std::filesystem::path once we stop using ghc::filesystem on some platforms
```

jagerman: However the addition of the `= "";` initializer here is unnecessary: strings always initialize to an empty string (when not using one of the other specific constructors).

dr7ana: I only added the `= ""` so I could compare [here](https://github.com/drouhana/lokinet/blob/06b2c58a7dc35d05fe250bc428a00c13209cf0e1/daemon/lokinet.cpp#L426), as it gave me an error that the expression using a string has to evaluate to bool-type. I can fix that

majestrate: no it's not worth it then. just use `std::string` and value initialize it

daemon/lokinet.cpp
```diff
@@ -417,7 +418,12 @@ namespace
         }
       }
 
-      if (options.configOnly)
+      if (options.generate)
+      {
+        options.configOnly = true;
+      }
+
+      if (options.configPath != "")
```
majestrate: replace with `if(not options.configPath.empty())`

daemon/lokinet-vpn.cpp
```diff
@@ -156,6 +156,10 @@ main(int argc, char* argv[])
     case 0:
       return exit_error(3, "One of --up/--down/--status/--kill must be specified");
     case 1:
+      break;
+    case 2:
+    case 3:
+    case 4:
       return exit_error(3, "Only one of --up/--down/--status/--kill may be specified");
     default:
       break;
```
majestrate: this `return exit_error` should replace the default case here instead of adding 2,3,4 and keeping the existing default case.

dr7ana: fixed

jagerman (approved):

PR #2111, opened 2023-01-10
IPC Socket Fixes

### LINKED ISSUE #2110: Lokinet default to RPC socket
### Current Functionality
lokinet.ini (under the [api] section) defaults to `bind=tcp://127.0.0.1:1190`

#2057 adds the ability to access an IPC socket by changing the line to `bind=ipc:///some/path/my.sock`

### Desired Functionality
Lokinet should be able to bind to multiple listening addresses instead of one, by repeating the `bind=...` line such as

> bind=tcp://127.0.0.1:1190
> bind=ipc:///some/path/my.sock

### TODO (so far)

- Update oxenmq interface to call `listen_plain` for each bind address
- Update config file to take list instead of single value
- Update config parsing logic to chdir to data directory to accommodate IPC socket relative paths

dr7ana:
Currently, lokinet.ini (under the [api] section) defaults to `bind=tcp://127.0.0.1:1190`

#2057 adds the ability to access an IPC socket by changing the line to `bind=ipc:///some/path/my.sock`

Lokinet should be able to bind to multiple listening addresses instead of one, by repeating the `bind=...` line such as

> bind=tcp://127.0.0.1:1190
> bind=ipc:///some/path/my.sock

Fixes: #2110

### DIFF
--- a/daemon/lokinet.cpp
+++ b/daemon/lokinet.cpp
@@ -588,6 +588,9 @@ namespace
         return;
       }
 
+      // change cwd to dataDir to support relative paths in config
+      fs::current_path(conf->router.m_dataDir);
+
       ctx = std::make_shared<llarp::Context>();
       ctx->Configure(std::move(conf));
 
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -1135,7 +1135,7 @@ namespace llarp
         "bind",
         DefaultRPCBind,
         MultiValue,
-        [=, first = true](std::string arg) mutable {
+        [this, first = true](std::string arg) mutable {
           if (first)
           {
             m_rpcBindAddresses.clear();
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -1114,7 +1114,12 @@ namespace llarp
   void
   ApiConfig::defineConfigOptions(ConfigDefinition& conf, const ConfigGenParameters& params)
   {
-    constexpr Default DefaultRPCBindAddr{"tcp://127.0.0.1:1190"};
+    constexpr std::array DefaultRPCBind{
+        Default{"tcp://127.0.0.1:1190"},
+#ifndef _WIN32
+        Default{"ipc://rpc.sock"},
+#endif
+    };
 
     conf.defineOption<bool>(
         "api",
@@ -1128,20 +1133,22 @@ namespace llarp
     conf.defineOption<std::string>(
         "api",
         "bind",
-        DefaultRPCBindAddr,
-        [this](std::string arg) {
-          if (arg.empty())
+        DefaultRPCBind,
+        MultiValue,
+        [=, first = true](std::string arg) mutable {
+          if (first)
           {
-            arg = DefaultRPCBindAddr.val;
+            m_rpcBindAddresses.clear();
+            first = false;
           }
           if (arg.find("://") == std::string::npos)
           {
             arg = "tcp://" + arg;
           }
-          m_rpcBindAddr = std::move(arg);
+          m_rpcBindAddresses.emplace_back(arg);
         },
         Comment{
-            "IP address and port to bind to.",
+            "IP addresses and ports to bind to.",
             "Recommend localhost-only for security purposes.",
         });
 
--- a/llarp/router/router.hpp
+++ b/llarp/router/router.hpp
@@ -298,9 +298,6 @@ namespace llarp
     void
     PumpLL();
 
-    const oxenmq::address DefaultRPCBindAddr = oxenmq::address::tcp("127.0.0.1", 1190);
-    bool enableRPCServer = false;
-    oxenmq::address rpcBindAddr = DefaultRPCBindAddr;
     std::unique_ptr<rpc::RpcServer> m_RPCServer;
 
     const llarp_time_t _randomStartDelay;
--- a/llarp/rpc/rpc_server.cpp
+++ b/llarp/rpc/rpc_server.cpp
@@ -25,13 +25,13 @@ namespace llarp::rpc
   RpcServer::RpcServer(LMQ_ptr lmq, AbstractRouter* r)
       : m_LMQ{std::move(lmq)}, m_Router{r}, log_subs{*m_LMQ, llarp::logRingBuffer}
   {
-    for (auto addr : r->GetConfig()->api.m_rpcBindAddresses)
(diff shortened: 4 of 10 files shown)


llarp/config/config.cpp
```diff
@@ -1114,7 +1114,10 @@ namespace llarp
   void
   ApiConfig::defineConfigOptions(ConfigDefinition& conf, const ConfigGenParameters& params)
   {
-    constexpr Default DefaultRPCBindAddr{"tcp://127.0.0.1:1190"};
+    constexpr std::array DefaultRPCBind{
+        Default{"tcp://127.0.0.1:1190"},
+        Default{"ipc://rpc.sock"},
```
jagerman, line 1120: I think we probably want to not include the latter one here as a default for windows.  I know there is, in theory, support for unix sockets in Windows 10 since sometime in 2018 or so, but I have no idea if it actually *works* via the rest of our software stack (zmq, etc.).

So I think let's just remove it from the defaults for windows, at least until someone has the time and energy to test it and figure out how to get windows to tell us whether it actually supports it yet.  (The only docs I can find on how to tell involve opening a windows command prompt as admin and running some "sc" comment, which is just... no.)
```suggestion
#ifndef _WIN32
        Default{"ipc://rpc.sock"},
#endif
```

jagerman (approved):

daemon/lokinet.cpp
```diff
@@ -588,6 +588,9 @@ namespace
         return;
       }
 
+      // change cwd to dataDir to support relative paths in config
+      fs::current_path(conf->router.m_dataDir);
```
majestrate, line 592: make this happen in `llarp::Context::Configure` instead so that it's not just lokinet binary that does this but also embedded lokinet and macos net ext

llarp/config/config.cpp
```diff
@@ -1128,20 +1133,22 @@ namespace llarp
     conf.defineOption<std::string>(
         "api",
         "bind",
-        DefaultRPCBindAddr,
-        [this](std::string arg) {
-          if (arg.empty())
+        DefaultRPCBind,
+        MultiValue,
+        [=, first = true](std::string arg) mutable {
```
majestrate: dont capture with `=` here, use `this` as the former will copy construct things we dont want to like the parent function's arguments.

llarp/router/router.hpp
```diff
@@ -298,9 +298,6 @@ namespace llarp
     void
     PumpLL();
 
-    const oxenmq::address DefaultRPCBindAddr = oxenmq::address::tcp("127.0.0.1", 1190);
-    bool enableRPCServer = false;
```
majestrate, line 302: i like this kind of cleanup here a lot

llarp/rpc/rpc_server.cpp
```diff
@@ -23,7 +24,15 @@ namespace llarp::rpc
 {
   RpcServer::RpcServer(LMQ_ptr lmq, AbstractRouter* r)
       : m_LMQ{std::move(lmq)}, m_Router{r}, log_subs{*m_LMQ, llarp::logRingBuffer}
-  {}
+  {
+    for (auto addr : r->GetConfig()->api.m_rpcBindAddresses)
```
majestrate: `const auto &` will prevent an extra copy construction here, a pattern i like is to only use `auto &` or `const auto &` unless a copy construction is explicitly desired in each iteration every time. using `auto &` will let you `std::move` int (steal ownership) in the loop body, and `const auto &` will ensure any copy constriction you do is on purpose, not a side effect of the loop.

majestrate: const correctness is also here too as unless you explicitly want to modify things in place with the loop, it should be a `const auto &`

llarp/rpc/rpc_server.cpp
```diff
@@ -23,7 +24,15 @@ namespace llarp::rpc
...
+    {
+      m_LMQ->listen_plain(addr.zmq_address());
+      LogInfo("Bound RPC server to ", addr.full_address());
+    }
+
+    this->AddRPCCats();
```
majestrate: better name suggestion: `AddRPCCategories`

daemon/lokinet.cpp
jagerman: Scratch those; we don't want the embedded lib to do this (and it won't, because it doesn't call this function), and mac is... a special snowflake that runs as root and data-dir is some obscure random path.

We can fix mac in a later PR if it isn't working.

llarp/config/config.cpp
jagerman: It won't copy things you don't use, but it's also not the best practice just because you aren't being explicit about the things you want to use and might unintentionally copy things outside the lambda you didn't mean to.

jagerman: I'll use `=` (or `&`) sometimes, but generally only if it's a tiny lambda of one or two lines where the copies/references are obvious.

llarp/rpc/rpc_server.cpp
jagerman: Yeah, it's good practice to always start from `for (const auto& x : loop)`, and then if you specifically need to modify remove the `const`, and if you specifically need to copy each element, remove the `&`.

llarp/rpc/rpc_server.cpp
jagerman: Nibbler disagrees.  :black_cat: (But he doesn't actually get a vote).

llarp/config/config.cpp
dr7ana: fixed

llarp/rpc/rpc_server.cpp
dr7ana: fixed

jagerman:
Something else that would be very useful is to be able to expose an encrypted/authenticated but *public* listener.  This needs two config additions:

1. We need to be able to specify the address on which we `listen_curve()`.
2. We need to be able to specify one (or more) public keys that will be accepted.

Something like this in the config:
```ini
[api]
bind_curve = tcp://0.0.0.0:1234
curve_pubkey = abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789
```

This would then use `listen_curve()` on that address (in addition to any of the `listen_plain()`s that happen from the `bind` options currently).

`listen_curve()` takes an "allow" callback: the allow callback would be looking at the public key and return `AuthLevel::admin` if it is in the `curve_pubkey` list, otherwise `AuthLevel::denied` to refuse the connection.

PR #2113, opened 2023-01-10
testnet fixes

dr7ana:
-- add ability to pass .ini path without --config flag -
- -r flag runs lokinet in router mode


daemon/lokinet.cpp
```diff
@@ -419,6 +419,11 @@ namespace
         }
       }
 
+      if (options.router)
```
majestrate: why not `opts.isSNode = options.router;` ?

dr7ana: resolved and merged into previous commit

jagerman (approved):

PR #2120, opened 2023-01-17
Fix incorrect usage of oxen-logging syslog

tewinget:
Previously oxen-logging was erroneously hard-coded to use the target "lokinet" for system logs.  Obviously this is wrong for anything else which uses oxen-logging and the system log.  This changes our call to add_sink to pass "lokinet" as the target rather than the config filename, and updates oxen-logging to use that argument correctly.

### DIFF
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -411,7 +411,7 @@ namespace llarp
     if (log::get_level_default() != log::Level::off)
       log::reset_level(conf.logging.m_logLevel);
     log::clear_sinks();
-    log::add_sink(log_type, conf.logging.m_logFile);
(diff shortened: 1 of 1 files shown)


majestrate:
you need to bump the required oxen-logging version in `external/CMakeLists.txt`

tewinget:
I don't see that as one of the submodules which has such a requirement, i.e. it appears it's always built from source?

majestrate:
ah, nevermind i thought we had one for that like oxenc

jagerman:
Yeah, oxen-logging doesn't have releases (so far); we always build it from source in all three projects using it.

jagerman (approved):

PR #2121, opened 2023-01-19
RPC refactor

### LINKED ISSUE #2058: redo lokinet rpc
we need to totally redo the rat's nest that is lokinet's zmq rpc

the desired outcome:

* documented endpoints that can give generated markdown
* api actually fit the needs that we have, not just be a collection of random junk
* get rid of the 5 level deep mutable lambdas

### LINKED ISSUE #2114: exit hotswap via lokinet-vpn tool
right now to change exits on runtime you would:

```bash
lokinet-vpn --down
lokinet-vpn --up --exit newexit
```

but this leaks packets.
we should add a --swap flag to add new exit and remove old exit on runtime

alternatively, make `--up` a deprecated alias to `--add` which is the actual behavior now in practice.

### LINKED ISSUE #2115: rpc refactor
current rpc code is horrible

we want to:

* redefine rpc endpoints
* make endpoint documented and add tooling to auto generate docs for api.oxen.io
* define what the new interfaces from the endpoint layer will need to expose for rpc

dr7ana:
Changes:
- Current endpoints refactored
- Implemented communication architecture based on oxen-core
- New endpoints added (list_exits, map_exit, unmap_exit, swap_exits)

Fixes: #2115, #2058, #2114


llarp/config/ini.cpp
```diff
@@ -168,4 +168,36 @@ namespace llarp
...
+      return;
+    }
+    if (m_Config.empty())
+    {
+      LogError("New config not loaded when attempting new .ini save");
+      return;
+    }
+    if (m_FileName.empty())
+    {
+      LogError("New config cannot be saved with filepath specified");
```
majestrate: these LogError should be a throw

llarp/rpc/rpc_server.cpp
```diff
@@ -98,627 +57,538 @@ namespace llarp::rpc
...
+      return;
+    }
+
+    if (quiclistener.request.port)
+    {
+      auto id = 0;
+      try
+      {
+        SockAddr addr{
+            quiclistener.request.remoteHost + ":" + std::to_string(quiclistener.request.port)};
```
majestrate: `SockAddr addr{quiclistener.request.remoteHost, net::port_t::from_host(quiclistener.request.port)};`

llarp/rpc/rpc_server.cpp
```diff
@@ -98,627 +57,538 @@ namespace llarp::rpc
...
+    }
+
+    if (quicconnect.request.closeID)
+    {
+      quic->forget(quicconnect.request.closeID);
+      quicconnect.response = CreateJSONResponse("OK");
+      return;
+    }
+
+    SockAddr laddr{};
```
majestrate: `SockAddr laddr{quicconnect.request.bindAddr};`

llarp/config/ini.cpp
```diff
@@ -82,8 +83,8 @@ namespace llarp
       while (!line.empty() && whitespace(line.back()))
         line.remove_suffix(1);
 
-      // Skip blank lines and comments
-      if (line.empty() or line.front() == ';' or line.front() == '#')
+      // Skip blank lines but NOT comments
+      if (line.empty() or line.front() == ';')
```
majestrate: `;` is considered comment delimiter just like `#`

llarp/config/ini.cpp
```diff
@@ -168,4 +169,33 @@ namespace llarp
...
+    for (const auto& [section, values] : m_Config)
+    {
+      ofs << std::endl << "[" << section << "]" << std::endl;
+      for (const auto& [key, value] : values)
+      {
+        ofs << key << "=" << value << std::endl;
+      }
+    }
+
+    m_Config.clear();
```
majestrate: why do we want to call `clear()` here?

llarp/rpc/param_parser.hpp
```diff
@@ -0,0 +1,359 @@
+#pragma once
+
+#include "json_binary_proxy.hpp"
+#include "oxenc/bt_serialize.h"
```
majestrate: `#include <oxenc/bt_serialize.h>`

llarp/rpc/json_binary_proxy.hpp
```diff
@@ -0,0 +1,182 @@
+#pragma once
+
+#include "crypto/crypto.hpp"
```
majestrate: we dont need this include

majestrate: (if we do use it use `#include <llarp/crypto/crypto.hpp>`)

llarp/config/ini.cpp
dr7ana: I was following the same convention as the other functions that write to file, which is to [clear](https://github.com/oxen-io/lokinet/blob/4cbfaf1d2fe19845b5f3061916844c25d86d88ba/llarp/config/ini.cpp#L169) after writing

dr7ana: though that's a good point... i imagine the other functions call clear() to make the object re-usable, but this configparser object is destroyed after the call to config endpoint is completed

majestrate: this also makes this able to be const

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +57,570 @@ namespace llarp::rpc
...
       return;
     }
+
+    if (quicconnect.request.remoteHost.empty() and quicconnect.request.closeID == 0)
+    {
+      quicconnect.response = CreateJSONError("Host not provided");
+      return;
+    }
+
+    auto endpoint = (quicconnect.request.endpoint.empty())
```
majestrate, line 195: race condition

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +57,570 @@ namespace llarp::rpc
...
+        ? GetEndpointByName(m_Router, "default")
+        : GetEndpointByName(m_Router, quicconnect.request.endpoint);
+
+    if (not endpoint)
+    {
+      quicconnect.response = CreateJSONError("No such local endpoint found.");
+      return;
+    }
+
+    auto quic = endpoint->GetQUICTunnel();
```
majestrate, line 205: race condition

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +57,570 @@ namespace llarp::rpc
...
+      return;
+    }
+
+    SockAddr laddr{quicconnect.request.bindAddr};
+
     try
     {
-      handleRequest(
-          *maybe, [defer = msg.send_later()](std::string result) { defer.reply(result); });
+      auto [addr, id] = quic->open(
```
majestrate, line 226: race condition

llarp/config/ini.hpp
```diff
@@ -24,6 +24,12 @@ namespace llarp
     bool
     LoadFile(const fs::path& fname);
 
+    /// load new .ini file from string (calls ParseAll() rather than Parse())
+    /// return true on success
+    /// return false on error
+    bool
```
majestrate, line 30: this should (eventually) use void return and only throw on error

llarp/net/ip_range.hpp
```diff
@@ -24,6 +24,14 @@ namespace llarp
         : addr{std::move(address)}, netmask_bits{std::move(netmask)}
     {}
 
+    static IPRange
+    StringInit(std::string _range)
```
majestrate: this should be called `from` or `from_string` imo

llarp/rpc/rpc_server.hpp
```diff
@@ -1,35 +1,159 @@
...
+        log::info(logcat, "RPC request 'rpc.{}' failed with: {}", rpc.name, e.what());
+        rpc.response = CreateJSONError(
+            fmt::format("RPC request 'rpc.{}' failed with: {}", rpc.name, e.what()));
+      }
+      catch (const std::exception& e)
+      {
+        log::info(logcat, "RPC request 'rpc.{}' raised an exception: {}", rpc.name, e.what());
+        rpc.response = CreateJSONError(
+            fmt::format("RPC request 'rpc.{}' raised an exception: {}", rpc.name, e.what()));
+      };
```
majestrate: stray `;` ?

llarp/rpc/rpc_server.hpp
```diff
@@ -1,35 +1,159 @@
...
+    void (*invoke)(oxenmq::Message&, RPCServer&);
+  };
 
-    void
-    AddRPCCategories();
+  //  RPC request registration
+  //    Stores references to RPC requests in a unordered map for ease of reference
+  //    when adding to server. To add endpoints, define in rpc_request_definitions.hpp
+  //    and register in rpc_server.cpp
+  extern const std::unordered_map<std::string, std::shared_ptr<const rpc_callback>> rpc_request_map;
```
majestrate: why `shared_ptr` and not `unique_ptr`?

llarp/net/ip_range.hpp
```diff
@@ -24,6 +25,27 @@ namespace llarp
         : addr{std::move(address)}, netmask_bits{std::move(netmask)}
     {}
 
+    explicit IPRange(std::string _range)
```
majestrate, line 28: i dont think this constructor needs to be `explicit`

jagerman: Single argument constructors should *always* be `explicit` by default and only made non-explicit if you are sure you want that.  Not being explicit means you want:
```C++
void somefunc(IPRange x) { ... }

std::string x = "...";
somefunc(x);
```
to be allowed.

daemon/lokinet.cpp
```diff
@@ -525,7 +525,7 @@ namespace
               "⠄⠄⠄⠄⠄⠿⠿⠟⠛⡹⠉⠛⠛⠿⠿⣿⣿⣿⣿⣿⡿⠂⠄⠄⠄⠄⠄⠄⠄⠄⠄⠄⠄",
               "⠠⠤⠤⠄⠄⣀⠄⠄⠄⠑⠠⣤⣀⣀⣀⡘⣿⠿⠙⠻⡍⢀⡈⠂⠄⠄⠄⠄⠄⠄⠄⠄⠄",
               "⠄⠄⠄⠄⠄⠄⠑⠠⣠⣴⣾⣿⣿⣿⣿⣿⣿⣇⠉⠄⠻⣿⣷⣄⡀⠄⠄⠄⠄⠄⠄⠄⠄",
-              "file a bug report now or be cursed with this "
+              "file a bug report now or be cursed with this ",
```
jagerman: I don't think the comma is supposed to be there -- that last line is meant to be one long line (i.e. we're wrapping the string in source code only, not in the output).

dr7ana: fixed

llarp/rpc/rpc_server.hpp
```diff
@@ -1,35 +1,167 @@
...
+  {
+   public:
+    RPCServer& server;
+    RPC rpc{};
+
+    EndpointHandler(RPCServer& _server, DeferredSend _replier) : server{_server}
+    {
+      log::critical(logcat, "init tag: {}", oxenc::to_hex(_replier.reply_tag));
+      rpc.replier.emplace(std::move(_replier));
+      log::critical(logcat, "now tag: {}", rpc.replier->reply_tag);
```
jagerman: Debugging

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +61,518 @@ namespace llarp::rpc
...
+      }
+    }
+
+    m_Router.hiddenServiceContext().GetDefault()->map_exit(
+        map_request.request.address,
+        map_request.request.token,
+        map_request.request.ip_range,
+        [unmap = std::move(unmap_request),
+         &r = m_Router,
+         old_exit = swapexits.request.exit_addresses[0]](bool success, std::string result) mutable {
```
majestrate, line 463: ah yes, you discovered how easy it is to fall down into 5 levels of recursive lambdas.

majestrate: a better way i have been trying out can be seen here: https://github.com/oxen-io/lokinet/pull/2119/files#diff-048b8cbbc9425cec6b3ef996ec840de9d93a23249a65b13eb66f0b103209ac11R110

contrib/omq-rpc.py
```diff
@@ -0,0 +1,99 @@
...
+                else:
+                    print(x.decode(), end="\n\n")
+
+else:
+    print("Request timed out", file=sys.stderr)
+    socket.close(linger=0)
+    sys.exit(1)
+
+
+# ./lmq-rpc.py ipc://$HOME/.oxen/testnet/oxend.sock 'llarp.get_service_nodes' | jq
```
tewinget: `s/lmq/omq/`

daemon/lokinet-vpn.cpp
```diff
@@ -225,6 +235,15 @@ main(int argc, char* argv[])
     }
     return 0;
   }
+
+  if (not options.swapExits.empty())
+  {
+    nlohmann::json opts{{"exit_addresses", std::move(options.swapExits)}};
+
+    if (not OMQ_Request(omq, connectionID, "llarp.swap_exits", std::move(opts)))
```
majestrate: we dont actually need a new endpoint on the other side for this.
we can just do two requests  where the second one doesn't fire unless the first one is successful.

llarp/rpc/rpc_server.cpp
jagerman: I don't see any lambda nesting at all here?

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +61,518 @@ namespace llarp::rpc
...
+  //  to new exit. Similar to how map_exit steals the oxenmq deferredsend object, swapexit
+  //  moves the replier object to the unmap_exit struct, as that is called second. Rather than
+  //  the nested lambda within map_exit making the reply call, it instead calls the unmap_exit logic
+  //  and leaves the message handling to the unmap_exit struct
+  void
+  RPCServer::invoke(SwapExits& swapexits)
+  {
+    MapExit map_request;
+    UnmapExit unmap_request;
+    auto current_exits = m_Router.hiddenServiceContext().GetDefault()->ExtractStatus()["exitMap"];
```
tewinget:
```suggestion
    auto endpoint = m_Router.hiddenServiceContext().GetDefault();
    auto current_exits = endpoint->ExtractStatus()["exitMap"];
```

and then also replace `m_Router.hiddenServiceContext().GetDefault()` the 2 places below with `endpoint` (and capture `endpoint` in the lambda by value)

daemon/lokinet-vpn.cpp
jagerman: That would push complexity down into the caller which doesn't seem desirable: the client would have to have logic about the proper order and handling of the swap, whereas a `swap_exits` endpoint provides an atomic (from the caller's perspective) operation.

llarp/rpc/rpc_server.cpp
majestrate: ah yes, nvm, i am used to this already being 2 or 3  levels deep.

llarp/config/ini.hpp
```diff
@@ -47,13 +53,20 @@ namespace llarp
     void
     Save();
 
+    /// save new .ini config file to path
+    void
+    SaveNew() const;
```
tewinget, line 58: if only called by this class, make it private

llarp/net/ip_range.hpp
tewinget: any reason this isn't just a constructor?

llarp/net/ip_range.hpp
```diff
@@ -24,6 +25,27 @@ namespace llarp
...
+
+    static IPRange
+    StringInit(std::string _range)
+    {
+      IPRange range{};
+      range.FromString(_range);
+      return range;
+    }
+
+    static bool
```
tewinget: appears unused

llarp/net/ip_range.hpp
tewinget: actually, it *is* a constructor just above and I'm blind.  probably just remove this function.

dr7ana: done

llarp/rpc/rpc_request_definitions.hpp
```diff
@@ -0,0 +1,311 @@
...
+    {
+      std::string bindAddr;
+      int closeID;
+      std::string endpoint;
+      uint16_t port;
+      std::string remoteHost;
+    } request;
+  };
+
+  //  RPC: quick_listener
```
tewinget: `s/quick/quic/`

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +61,518 @@ namespace llarp::rpc
...
+        endpoint->PutSRVRecord(std::move(srvData));
+      }
+
+      SetJSONResponse(result, quiclistener.response);
+      return;
+    }
+  }
+
+  void
+  RPCServer::invoke(LookupSnode& lookupsnode)
```
tewinget, line 311: this RPC call should be revisited (i.e. what does/should it do and why?) but that may be out of scope for this PR

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +61,518 @@ namespace llarp::rpc
...
+  void
+  RPCServer::invoke(SwapExits& swapexits)
+  {
+    MapExit map_request;
+    UnmapExit unmap_request;
+    auto current_exits = m_Router.hiddenServiceContext().GetDefault()->ExtractStatus()["exitMap"];
+
+    if (current_exits.empty())
+    {
+      SetJSONError("No exits", swapexits.response);
```
tewinget: Not a huge deal, but this error message is a bit sparse.  Maybe go with something like "Cannot swap to new exit: no current exits"

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +61,518 @@ namespace llarp::rpc
...
+    // populate map_exit request with old IP ranges
+    for (auto& [range, exit] : current_exits.items())
+    {
+      if (exit.get<std::string>() == swapexits.request.exit_addresses[0])
+      {
+        map_request.request.ip_range.push_back(IPRange::StringInit(range));
+        unmap_request.request.ip_range.push_back(IPRange::StringInit(range));
+      }
+    }
+
```
tewinget, line 456: if no mapped ranges found matching the requested swap, should error out here.

tewinget: This case *would* get handled by the ip_range member of the unmap request being empty in that handler, but the returned error would be ambiguous/misleading.

llarp/rpc/rpc_request_definitions.hpp
dr7ana: but what if it listens quickly

llarp/rpc/rpc_server.cpp
dr7ana: agreed. only included it as it was present pre-refactor

tewinget (approved):

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +61,527 @@ namespace llarp::rpc
...
-                    {
-                      reply(CreateJSONError(fmt::format(
-                          "failed to set [{}]:{}: value is not a string", section, key)));
-                      return;
-                    }
-                    r->GetConfig()->Override(section, key, value.get<std::string>());
-                  }
-                }
-              }
+              r.routePoker()->Down();
```
majestrate: this will 100% cause packets to leak out, which is what we want to prevent with this endpoint.

majestrate (changes requested):

llarp/service/endpoint.cpp
```diff
@@ -2114,6 +2191,20 @@ namespace llarp
...
+    {
+      // unmap all ranges that match the given exit when hot swapping
+      m_ExitMap.RemoveIf([&](const auto& item) -> bool {
+        if ((range.Contains(item.first)) and (item.second.ToString() == exit))
+        {
+          log::info(logcat, "{} unmap {} range mapping to exit node {}", Name(), item.first, exit);
+          return true;
+        }
+        return false;
+      });
```
majestrate, line 2208: we should put the route poker down after doing the `RemoveIf` call if we no longer have anything in our exit map.

llarp/rpc/rpc_server.cpp
```diff
@@ -107,626 +61,527 @@ namespace llarp::rpc
...
+
+    SetJSONResponse((status.empty()) ? "No exits" : status, listexits.response);
+  }
+
+  void
+  RPCServer::invoke(UnmapExit& unmapexit)
+  {
+    try
+    {
+      m_Router.routePoker()->Down();
```
majestrate: this can be removed if the call to `UnmapExitRange` puts the route poker down

majestrate (approved):

PR #2127, opened 2023-01-27
Update Documentation

majestrate:
stub pr for docs that dan is adding


majestrate:
random note for addition to this pr:

```
relay up/down has: 32 bytes (nounce) +3 bytes bencode, + 16 bytes (txid) + 3 bytes bencode + 16 bytes (rxid) + 3 bytes bencode + 3 bytes bencode for link layer message kind (1:a) + 2 bytes for 'de' + the data you are sending as ciphertext padded to nearest 128 bytes + 4 or 5 bytes bencode. 
the smallest possible relay up/down message is 210 plaintext then add the current wire protocol overhead of 108 bytes for the node to node transmission header for non fragmented transmission.
the smallest possible sized udp packet for node to node data via a path ends up being 328 bytes.
as long as your routing message is under 128 bytes it will be padded to 128 when inside the relay/up down message
```

majestrate:
i think we can merge this.

jagerman (approved):

PR #2129, opened 2023-01-30
Add Support for Encrypted/Authenticated Listeners

### LINKED ISSUE #2122: Add encrypted/authenticated IPC listener support
Something that would be very useful is to be able to expose an encrypted/authenticated but *public* listener.  This needs two config additions:

1. We need to be able to specify the address on which we `listen_curve()`.
2. We need to be able to specify one (or more) public keys that will be accepted.

Something like this in the config:
```ini
[api]
bind_curve = tcp://0.0.0.0:1234
curve_pubkey = abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789
```

This would then use `listen_curve()` on that address (in addition to any of the `listen_plain()`s that happen from the `bind` options currently).

`listen_curve()` takes an "allow" callback: the allow callback would be looking at the public key and return `AuthLevel::admin` if it is in the `curve_pubkey` list, otherwise `AuthLevel::denied` to refuse the connection.

_Originally posted by @jagerman in https://github.com/oxen-io/lokinet/issues/2111#issuecomment-1382254445_

dr7ana:
User can specify encrypted listener addresses and comma-delimited public keys to be accepted by exposed encrypted listener. Keys must be attached to a listener address.

Example: 

`bind_curve=tcp://0.0.0.0:1234|pubkeyA,pubkeyB`
`bind_curve=tcp://0.0.0.0:5678|pubkeyC,pubkeyD`
 
In the given example above, port 1234 is only accessible by whitelisted pubkeys A and B, while 5678 is accessible by C and D.

Fixes: #2122

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -3,6 +3,7 @@
 #include "ini.hpp"
 
 #include <oxenmq/address.h>
+#include <oxenmq/oxenmq.h>
 #include <llarp/constants/files.hpp>
 #include <llarp/constants/platform.hpp>
 #include <llarp/constants/version.hpp>
@@ -1174,9 +1175,10 @@ namespace llarp
             key = "tcp://" + key;
 
           auto pubkeys = split(values, ",", true);
+          oxenmq::address key_addr{key};
 
           for (auto& pk : pubkeys)
-            m_rpcEncryptedAddresses[oxenmq::address{key}].emplace(pk);
+            m_rpcEncryptedAddresses[key_addr].emplace(pk);
         },
         Comment{
             "Specify encrypted listener addresses and comma-delimited public keys to be accepted ",
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -2,6 +2,7 @@
 #include "definition.hpp"
 #include "ini.hpp"
 
+#include <oxenmq/address.h>
 #include <llarp/constants/files.hpp>
 #include <llarp/constants/platform.hpp>
 #include <llarp/constants/version.hpp>
@@ -1152,10 +1153,45 @@ namespace llarp
             "Recommend localhost-only for security purposes.",
         });
 
-    conf.defineOption<std::string>("api", "authkey", Deprecated);
+    conf.defineOption<std::string>(
+        "api",
+        "bind_curve",
+        Default{""},
+        MultiValue,
+        [this](std::string arg) mutable {
+          if (arg.empty())
+            return;
 
-    // TODO: this was from pre-refactor:
-    // TODO: add pubkey to whitelist
+          auto pipe = arg.find("|");
+
+          if (pipe == arg.npos)
+            throw std::invalid_argument(
+                "Addresses and whitelisted pubkeys must be pipe-delimited key:value pairs");
+
+          auto key = arg.substr(0, pipe), values = arg.substr(pipe + 1, arg.npos);
+
+          if (not starts_with(key, "tcp://"))
+            key = "tcp://" + key;
+
+          auto pubkeys = split(values, ",", true);
+
+          for (auto& pk : pubkeys)
+            m_rpcEncryptedAddresses[oxenmq::address{key}].emplace(pk);
+        },
+        Comment{
+            "Specify encrypted listener addresses and comma-delimited public keys to be accepted ",
+            "by exposed encrypted listener. Keys must be attached to a listener address.",
+            "",
+            "Example: ",
+            "   bind_curve=tcp://0.0.0.0:1234|pubkeyA,pubkeyB",
+            "   bind_curve=tcp://0.0.0.0:5678|pubkeyC,pubkeyD",
+            "",
+            "In the given example above, port 1234 is only accessible by whitelisted ",
+            "pubkeys A and B, while 5678 is accessible by C and D.",
+            "",
+            "Note: tcp addresses passed without \"tcp://\" prefix will have it prepended"});
+
+    conf.defineOption<std::string>("api", "authkey", Deprecated);
   }
 
   void
--- a/llarp/config/config.hpp
+++ b/llarp/config/config.hpp
@@ -2,8 +2,8 @@
 #include "ini.hpp"
 #include "definition.hpp"
 
+#include <oxenmq/auth.h>
 #include <chrono>
-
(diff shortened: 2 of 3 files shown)


llarp/config/config.cpp
```diff
@@ -1,6 +1,7 @@
 #include "config.hpp"
 #include "definition.hpp"
 #include "ini.hpp"
+#include "oxenmq/address.h"
```
majestrate: angle brackets

llarp/config/config.cpp
```diff
@@ -1152,10 +1153,45 @@ namespace llarp
...
+        "bind_curve",
+        Default{""},
+        MultiValue,
+        [this](std::string arg) mutable {
+          if (arg.empty())
+            return;
+
+          auto pipe = arg.find("|");
+
+          if (pipe == arg.npos)
```
majestrate, line 1168: `std::string::npos`

llarp/config/config.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 #include "ini.hpp"
 #include "definition.hpp"
+#include "oxenmq/auth.h"
```
majestrate: angle brackets

llarp/config/config.cpp
majestrate: ~~actually remove this one include~~

llarp/config/config.cpp
```diff
@@ -1152,10 +1153,45 @@ namespace llarp
...
 
-    // TODO: this was from pre-refactor:
-    // TODO: add pubkey to whitelist
+          if (not starts_with(key, "tcp://"))
+            key = "tcp://" + key;
+
+          auto pubkeys = split(values, ",", true);
+
+          for (auto& pk : pubkeys)
+            m_rpcEncryptedAddresses[oxenmq::address{key}].emplace(pk);
```
jagerman: Should do a `oxenmq::address key_addr{key}` outside the loop (rather than needing to re-parse it for every iteration).

Also, if it throws, does the user see an appropriate exception message?

llarp/rpc/rpc_server.cpp
```diff
@@ -109,6 +109,17 @@ namespace llarp::rpc
       LogInfo("Bound RPC server to ", addr.full_address());
     }
 
+    for (const auto& [address, allowed_keys] : r->GetConfig()->api.m_rpcEncryptedAddresses)
+    {
+      m_LMQ->listen_curve(address.zmq_address(), [allowed_keys = allowed_keys](auto pk, ...) {
```
jagerman: `pubkey` is the second argument to the allow callback (first is the remote address).

llarp/rpc/rpc_server.cpp
```diff
@@ -109,6 +109,17 @@ namespace llarp::rpc
...
+        if (std::find(allowed_keys.begin(), allowed_keys.end(), pk) != allowed_keys.end())
+          return oxenmq::AuthLevel::admin;
```
jagerman: `std::find` does a linear scan to find a match; you want `allowed_keys.find()` instead for a set/map.  But actually, in this case, `allow_keys.count(pk)` will be simpler still.

llarp/rpc/rpc_server.cpp
```diff
@@ -109,6 +109,17 @@ namespace llarp::rpc
...
+
+        LogInfo("Curve pubkey not found in whitelist");
```
jagerman: I think this should be elevated to Warning instead of Info

jagerman: Would probably be also be useful to include the remote addr here (the first argument), something like:

```C++
log::warning(logcat, "Curve pubkey not in whitelist, denying incoming RPC connection from {}", addr);
```

jagerman:
This is going to need some deeper changes in lokinet to make it work w.r.t. to lokinet-side "server" pubkey, so for right now let's put this on hold until we can come back to it (likely after some other refactors are merged).  Some notes on what is required to make this work follow.

--- 

The main issue is where we specify the OMQ listener keypair: right now, we default-construct the `OxenMQ` instance during Router construction, which results in OxenMQ generating a new random keypair at every startup.  For a curve listener this is fairly useless because it means you'd have to go get this pubkey (which right now we don't even log, much less expose in any useful way) every time you want to use the client: you couldn't, for instance, set up a script that periodically polls lokinet for stats because every time the server restarts you'd have to (manually) obtain the new server pubkey and update it in the script.

What we need here is two pieces:

1. For a service node, we want to use the service node's x25519 keypair itself.  This is the same key as the ed25519 keypair we use for the router already, just converted into x25519 keyspace.  (We could call on libsodium to convert, but it's probably easier to just use the pre-converted x25519 keypair that oxend gives us through the `get_service_privkey` endpoint that we already call to get the ed25519 keypair).

This gets complicated, though, because we have to re-create the OxenMQ instance *after* we get this keypair and give it the new keypair we got from oxend.

In storage server, we solve this problem by using *two* OxenMQ instances: we create a temporary one during instantiation that only has one job: connecting to oxend and getting the keys.  This basically blocks startup until it succeeds, at which point we destroy that object and reconstruct a new OxenMQ using the just-fetched private key.

Lokinet likely needs to do something similar, but currently that is difficult because of how the OxenMQ instance is embedded in the Router instance (and this is the instance that we use to get the keys from oxend).  Likely a solution here is going to require us setting up a temporary OxenMQ instance to extract the keys *very early* in the startup (before we construct the router god object) and passing those keys into Router construction so that the OxenMQ instance that Router owns can get constructed with the keys.

2. For lokinet clients, we want to store a persistent x25519 private key in a generated file in the data directory.  We can generate and store a new key if this doesn't already exist, and (when curve listening is enabled) we should log the x25519 pubkey (not privkey!) in the startup logging.  Then it's just a matter of getting this once, putting into your logging script (or whatever is connecting to your lokinet rpc) and you're done.

This file-backed private key we then feed into the Router construction (similar to the snode case) so that it creates the OxenMQ instance with this key.

jagerman:
An alternative here would be to update OxenMQ to support a per-socket key in the `listen_curve` call.  Currently it just uses a master key, but there's no ZMQ requirement for that: we could set up the listened ZMQ socket with a distinct keypair.

PR #2130, opened 2023-02-05
fix openwrt mips32 static build

XutaxKamay:
We did talk some months ago about it on IRC, just asking to pull request this for now, I'll figure out the rest for the blackhole thing maybe its due to kernel config idk, I'll let you know

### DIFF
--- a/cmake/StaticBuild.cmake
+++ b/cmake/StaticBuild.cmake
@@ -242,6 +242,7 @@ set(openssl_system_env "")
 set(openssl_arch "")
 set(openssl_configure_command ./config)
 set(openssl_flags "CFLAGS=${deps_CFLAGS}")
+set(unbound_ldflags "")
 if(CMAKE_CROSSCOMPILING)
   if(ARCH_TRIPLET STREQUAL x86_64-w64-mingw32)
     set(openssl_arch mingw64)
@@ -257,6 +258,9 @@ if(CMAKE_CROSSCOMPILING)
   elseif(ARCH_TRIPLET STREQUAL mips64-linux-gnuabi64)
     set(openssl_arch linux-mips64)
   elseif(ARCH_TRIPLET STREQUAL mips-linux-gnu)
+    set(openssl_arch linux-mips32)
+  elseif(ARCH_TRIPLET STREQUAL mips-openwrt-linux)
+    set(unbound_ldflags "-latomic")
     set(openssl_arch linux-mips32)
   elseif(ARCH_TRIPLET STREQUAL mipsel-linux-gnu)
     set(openssl_arch linux-mips)
@@ -319,7 +323,7 @@ build_external(unbound
   --enable-static --with-libunbound-only --with-pic
   --$<IF:$<BOOL:${WITH_LTO}>,enable,disable>-flto --with-ssl=${DEPS_DESTDIR}
   --with-libexpat=${DEPS_DESTDIR}
-  "CC=${deps_cc}" "CFLAGS=${deps_CFLAGS}"
+  "CC=${deps_cc}" "CFLAGS=${deps_CFLAGS}" "LDFLAGS=${unbound_ldflags}"
 )
 add_static_target(libunbound unbound_external libunbound.a)
 if(NOT WIN32)


XutaxKamay:
Ah wait hold on, need to fix something

cmake/StaticBuild.cmake
```diff
@@ -258,6 +259,10 @@ if(CMAKE_CROSSCOMPILING)
     set(openssl_arch linux-mips64)
   elseif(ARCH_TRIPLET STREQUAL mips-linux-gnu)
     set(openssl_arch linux-mips32)
+  elseif(ARCH_TRIPLET STREQUAL mips-openwrt-linux)
+    # needed to fix unbound so that it links correctly with openssl (crypto, EVP_sha256)
```
majestrate: this comment seems like it'd be best in a commit message :^D

XutaxKamay: okay xD

XutaxKamay: just I leave the comment just in case though?

majestrate: i think having it in the commit only is better so i git blame preserves the info

XutaxKamay: oki

cmake/StaticBuild.cmake
```diff
@@ -319,7 +323,7 @@ build_external(unbound
   --enable-static --with-libunbound-only --with-pic
   --$<IF:$<BOOL:${WITH_LTO}>,enable,disable>-flto --with-ssl=${DEPS_DESTDIR}
   --with-libexpat=${DEPS_DESTDIR}
-  "CC=${deps_cc}" "CFLAGS=${deps_CFLAGS}"
+  "CC=${deps_cc}" "CFLAGS=${deps_CFLAGS}" "LDFLAGS=${deps_latomic}"
```
majestrate: better name: `deps_LDFLAGS`

majestrate:
a better commit message would be

```
fix openwrt mips32 static build

libunbound needs to link against libatomic because of openssl
```

cmake/StaticBuild.cmake
XutaxKamay: actually it was named like this before, but i remember seeing deps_ld somewhere

XutaxKamay: one sec

XutaxKamay: okay I think it should be fine to use deps_ld cause it's only used for android stuffs which is before where it sets the variable, i'll just re-use it https://github.com/oxen-io/lokinet/blob/bf0dc52df7c05d4a14b5d5140615c1d547430055/cmake/StaticBuild.cmake#L254

XutaxKamay:
> a better commit message would be
> 
> ```
> fix openwrt mips32 static build
> 
> libunbound needs to link against libatomic because of openssl
> ```

okay, let me know if it's good now

majestrate (approved):
looks good waiting for ci

majestrate:
ci failed.

cmake/StaticBuild.cmake
majestrate: no, that's the path to the linker itself.

majestrate: (we want to set additional linker flags)

XutaxKamay:
ah, got link? can't see it here i think

majestrate:
> ah, got link? can't see it here i think

ci.oxen.rocks

majestrate:
https://ci.oxen.rocks/oxen-io/lokinet

XutaxKamay:
> https://ci.oxen.rocks/oxen-io/lokinet

 LDFLAGS=/usr/lib/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi-ld

okay yeah that's what you said earlier, let me fix it

majestrate (commented):
okay this looks right, waiting for ci

majestrate:
oh ci done. let's merge.

majestrate (approved):

majestrate:
(ci status delivery hooks are borked yay)

XutaxKamay:
ah great sorry went to eat

PR #2134, opened 2023-02-10
reduce number of jobs pushed through omq proxy thread.

majestrate:
when we queue cpu heavy work in lokinet to worker threads we make 1 job per function call. we call a lot of jobs so this coleuses the jobs into 1 job that we push off at the end of the event loop cycle, reducing the number of jobs going across the omq proxy thread, in theory reducing cpu usage.

@jagerman 
i am curious how much perf increase we get on snodes, so i'd like to have this tested on testnet

### DIFF
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -72,11 +72,21 @@ namespace llarp
     _lastTick = llarp::time_now_ms();
     m_NextExploreAt = Clock_t::now();
     m_Pump = _loop->make_waker([this]() { PumpLL(); });
+    m_Work = _loop->make_waker([this]() { submit_work(); });
   }
 
   Router::~Router()
   {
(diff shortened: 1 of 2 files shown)


majestrate:
metric for measuring perf improvement: 

ratio of cumulative cpu usage across workers to cpu usage in the omq proxy thread.

higher ratio is better.

majestrate:
@jagerman please test this on testnet.
make sure to measure for perf before and after flipping.

majestrate:
it has a noticeable improvement on exit.loki. less context switches, lower system load average and higher throughput. 


if_enp1s0-day (1)
interrupts-day (1)
load-day (1)

i'll test it on my service node this week to see what it improves.

majestrate:
getting this into a point release would probably speed the network up a bit.

PR #2141, opened 2023-02-27
[WIP] liblokinet changes

majestrate:
placeholder for dan's branch for liblokinet used for code review.


llarp/quic/connection.cpp
```diff
@@ -897,35 +897,42 @@ namespace llarp::quic
       fprintf(stderr, "add_stream_data for non-stream returned [%ld,%ld]\n", nwrite, consumed);
       assert(consumed <= 0);
 
-      switch (nwrite)
+      if (nwrite == -240)
```
majestrate: there isn't really a good reason to use switch case for anything not an `enum class`. the compiler is smart enough to optimize it.

dr7ana: agreed. reverted to a if/else block after this for better control over breaking/continuing

llarp/lokinet_shared.cpp
```diff
@@ -1,4 +1,5 @@
 #include <lokinet.h>
+#include <filesystem>
```
majestrate, line 4: this should be `llarp/util/fs.hpp` as it contains compat shims for macos and such.

llarp/quic/endpoint.hpp
```diff
@@ -111,7 +111,7 @@ namespace llarp::quic
 
     // Called to handle an incoming packet
     void
-    handle_packet(const Packet& p);
+    handle_packet(Packet p);
```
jagerman: This would be better as an rvalue reference `Packet&& p` so that we are still passing it by reference rather than needing to copy it (there is internal ngtcp2 C stuff in here that likely isn't moveable, so taking it by value will involve at least *some* copying).  Whereas if you take it as a `Packet&&` you *have* to move it when calling, but we still take it as a reference so don't have to copy anything.

majestrate: hah, i knew it

dr7ana:
### Addressing Fix

1. Add an `std::optional<std::variant<llarp::service::Address, llarp::RouterID>>` member to `llarp::quic::Address`
2. Set it in `.remote` when we create initialize the `llarp::quic::Path` for `Connection` (I think `GetEndpointWithConvoTag` will give us what we want).
3. Hard-code `.remote` address to be `::1` (IPv6 localhost).
4. Chase through the code that constructs a `Connection` in `llarp::quic::Server` and `llarp::quic::Client` to stop using a `ConvoTag` and instead set the address variant.
5. Remove the `ConvoTag` stuff from `llarp::quic::Endpoint`
6. Where we call `SendToOrQueue` in `llarp::quic::Endpoint` we now pass the variant (instead of, currently, the `ConvoTag` via implicit conversion)

llarp/CMakeLists.txt
majestrate, line 1: we will need to make this a compile time flag so we can not compile in the libunbound parts in our dns subsystem. this wont really do anything right now.

however, we can also remove libssl and libexpat as a result which will make the static targets smaller by a lot probably

dr7ana: sorry jeff just saw this! sounds good to me, will implement if it works in reducing the size as we think it will

majestrate: There are a few things we need to do to make this compile time configurable:

- move libunbound resolver to its own compilation unit. 
- only add that new compiliation unit when libunbound is enabled
- in header for new compilation unit declare a single function that will create a shared_ptr of resolver_base for the libunbound resolver, add nop implementation of function in dns/server.cpp when libunvound is disabled on compile time.

majestrate: There are a few things we need to do to make this compile time configurable:

- move libunbound resolver to its own compilation unit. 
- only add that new compiliation unit when libunbound is enabled
- in header for new compilation unit declare a single function that will create a shared_ptr of resolver_base for the libunbound resolver, add nop implementation of function in dns/server.cpp when libunvound is disabled on compile time.

llarp/endpoint_base.hpp
```diff
@@ -122,7 +122,7 @@ namespace llarp
     virtual bool
     EnsurePathTo(
         AddressVariant_t addr,
-        std::function<void(std::optional<service::ConvoTag>)> hook,
+        std::function<void(std::optional<std::variant<service::Address, RouterID>>)> hook,
```
majestrate, line 125: `std::function<void(bool)>`

llarp/net/ip_address.hpp
```diff
@@ -24,7 +24,7 @@ namespace llarp
     /// Empty constructor.
     IpAddress() = default;
     /// move construtor
-    IpAddress(IpAddress &&) = default;
```
majestrate: bin this entire file if you can

Goop-arsh: https://github.com/oxen-io/lokinet/releases/tag/v0.9.11

.drone.jsonnet
```diff
@@ -315,6 +315,34 @@ local mac_builder(name,
...
+  name: name,
+  platform: { os: 'darwin', arch: 'amd64' },
+  steps: [
+    { name: 'submodules', commands: submodule_commands },
+    {
+      name: 'build',
+      environment: { SSH_KEY: { from_secret: 'SSH_KEY' } },
+      commands: [
+        'echo "Building on ${DRONE_STAGE_MACHINE}"',
+        'export CMAKE_BUILD_PARALLEL_LEVEL=6',
```
Goop-arsh, line 338:
```suggestion
        'export CMAKE_BUILD_PARALLEL_LEVEL=6',
```

PR #2150, opened 2023-03-27
add python script which generates an ini config snippet to pin edges

majestrate:
takes in multiple wallet addresses as arguments and outputs to stdout a config snippet that manually pins all snodes from those opers.

### DIFF
--- a/contrib/snode-pin.py
+++ b/contrib/snode-pin.py
@@ -32,13 +32,7 @@ for snode in resp.json().get("result").get("service_node_states", []):
         snodes[addr].add(snode.get("pubkey_ed25519"))
 
 # print the config snippet to stdout
-for oper, addrs in snodes:
-    print(f"# pin edges to use oper {oper}")
-    print("[network]")
-    for addr in addrs:
-        # convert base16 (hex) to zbase32 and add .snode suffix
-        print(
-            "strict-connect={}.snode".format(
-                oxenc.to_base32z(binascii.unhexlify(addr.strip()))
-            )
-        )
+for oper, addrs in snodes.items():
+    print(f"# pin edges to use oper {oper}") 
+    print("[network]") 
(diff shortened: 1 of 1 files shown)


contrib/snode-pin.py
```diff
@@ -0,0 +1,40 @@
...
+    snode_pubkeys = set()
+    jreq = {
+        "jsonrpc": "2.0",
+        "id": "0",
+        "method": "get_service_nodes",
+        "params": {"service_node_pubkeys": []},
+    }
+
+    # collect all snodes for this oper
+    resp = requests.post("https://public.loki.foundation/json_rpc", json=jreq)
```
frtget: The request can be moved outside of the for loop

majestrate: you're right

frtget (commented):
can't approve but it looks good

contrib/snode-pin.py
```diff
@@ -0,0 +1,44 @@
...
+for snode in resp.json().get("result").get("service_node_states", []):
+    addr = snode.get("operator_address", None)
+    if addr in addrs:
+        snodes[addr].add(snode.get("pubkey_ed25519"))
+
+# print the config snippet to stdout
+for oper in snodes:
+    print(f"# pin edges to use oper {oper}")
+    print("[network]")
+    for addr in snodes.get(oper):
```
dr7ana: This is trivial and won't likely make any realistic difference, but if you wanted to cut out the call to `snodes.get` and use list comprehension instead:
```
for oper, addrs in snodes.items():
    print(f"# pin edges to use oper {oper}")
    print("[network]")
    print("\n".join(
        f"strict-connect={oxenc.to_base32z(binascii.unhexlify(addr.strip()))}.snode"
        for addr in addrs
    ))
```

dr7ana (approved):

contrib/snode-pin.py
majestrate: true

PR #2164, opened 2023-04-27
Redo wire protocol race condition fix.

majestrate:
In commit a76acd49560f9845c2ad9a839364f2834afc7756 we attempted to fix issues on service nodes that related to spamming service nodes with traffc, but this caused a regression where clients could not connect to the network.

However, we also did something idiotic with the protocol handshake. The handshake logic was modified and caused client connections to fail. This was due to a change in how we decided what a timed out session is. We redo the close bug and revert the change to the protocol handshake.

### DIFF
--- a/llarp/iwp/session.cpp
+++ b/llarp/iwp/session.cpp
@@ -141,6 +141,8 @@ namespace llarp
     void
     Session::EncryptAndSend(ILinkSession::Packet_t data)
     {
+      if (m_State == State::Closed)
+        return;
       m_EncryptNext.emplace_back(std::move(data));
       TriggerPump();
       if (!IsEstablished())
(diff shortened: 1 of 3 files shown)


jagerman:
~~I'm flipping testnet to this branch for testing.~~ Will flip it tomorrow.

@dr7ana - can you `git cherry-pick b0b6e2b198ba32e8c5641e99494d2874b2243b8f` into your liblokinet-dev branch (which testnet currently follows)?

dr7ana (approved):
cherry-picked and merged to liblokinet-dev

llarp/iwp/session.cpp
```diff
@@ -352,7 +351,7 @@ namespace llarp
     bool
     Session::TimedOut(llarp_time_t now) const
     {
-      if (m_State == State::Ready)
```
majestrate, line 355: this line here is the most important part of the change.

`State::LinkIntro` means we are finalizing the protocol handshake, we are in a kind of half established state were we are an encrypted but unauthenticated session. this is where we are sending the other end a "link intro" which identifies who we are (by snode address), including the ephemeral pubkey that we were challenged with for establishing the encryption. it's stupid af and i hate it.

`State::Ready` means we are authenticated and ready to exchange frames for inter node communication. 

the state we hit here in this regression that one side thinks the other side is timed out when they really aren't, and are trying to finish the handshake. they can still send/recv but they just have no idea who it is they are talking to in terms of their .snode addr but the other end disagrees about where we are in that process and it gets stuck. this is because snode and client timeouts are different (because of course they are). someone wanted to add that requirement onto the pile and so it was added, even though it was something i did object to for complexity reasons. 
as usual i gave up fighting on that because people have strong opinions here and i just want stuff to get done, rocking the boat is a waste of time.

PR #2165, opened 2023-04-30
document behavior of dns section.

majestrate:
Before we were just shoving all options into libunbound and assume they end with a ':', now we ensure they do and throw if they do not.

This adds a check and documents this behavior in config section.


majestrate:
copy editing desired too.

majestrate:
@tewinget i will correct some spelling mistakes in this and force push that change. merging this would be great.

PR #2168, opened 2023-05-05
[docs] node to node protocol.

majestrate:
rough draft outline of lokinet's node to node protocol.

### DIFF
--- a/docs/node2node.md
+++ b/docs/node2node.md
@@ -1,0 +1,105 @@
+# node to node messages
+
+
+[summary placeholder]
+
+(todo: fully describe pop and shift mechanism)
+(todo: calcuate optimal data sizes)
+
+
+## streamed requests
+
+direct node to node request/response messages sent on a bidi quic stream.
+
+the stream starts by sending 2 bytes identifying the stream kind of which there are the following:
+
+* network data lookup
+* network data upsert
+* path build
+* path update
+* path stream
+
+### network data lookup stream
+
+used for fetching of:
+
+* pubkey list with unlock times if they are applicable.
+* pubkey list delta given last hash of full list.
+* router contact by snode identity pubkey.
+* router contacts by operator wallet address.
+* look up wallet addresses staked into snode.
+* look up snodes staked by a wallet address.
+
+
+### network data upsert stream
+
+used to upsert (insert/update) records for:
+
+* gossiped router contacts.
+* encrypted introsets.
+* signed service descriptors. (aka the "exit market" metadata).
+
+
+### path builds
+
+#### path build request
+
+we have 8 path build records in a path build, we pop and shift the records and decrypt the record meant for us.
+
+the plainext record contains:
+
+* next hop's pubkey
+* path lifetime
+* signing pubkey used by the path creator to sign updates
+* encryption pubkey to generate shared secret at the path
+* additional extra fields used in the future.
+
+
+(todo: i want to permit multiple kinds of pubkeys for forwards compat)
+
+we could also add an additional list of nonces on the path build that lets us open multiple path streams as we build our path.
+given we provide N nonces in the build we are opening up N path streams as we make our path build. see the path streams section below.
+
+(todo: shrink tx/rx path id to be 8 bytes each?)
+
+#### path build result
+
+(todo: finish this)
+
+### path streams
+
+onion routed persisting stream connection to the pivot router for sending anonymised requests destined for the pivot router.
+
+streams are created for an existing path by opening a "path stream" stream providing the path's tx/rx id and 24 bytes nonce.
+on accepting the stream creation the node will open a "path stream" to their upstream router but with a nonce that is deterministically mutated `new_nonce = nonce ^ hash(nonce)`, this mutation reduces stream corrilation across hops.
+on rejecting the stream creation the node will shutdown the read end of the stream, sending a reply signed by our identity key and then shutting down the write end of the stream (which fully closes the stream).
+if the accepting node is the pivot route it will write. 
+
+a "request stream path result" message has a flag for the result, zero for success and non zero for failure.
+
+(todo: if we use a ring signature on the "request path stream result" message can do a pop and shift on the pubkeys in the ring so that the original requestor can know who rejected/accepted without the intermediate hops knowing who did. would we have to generate derived keys here?)
+(todo: if we use normal signatures here would we want to blind the signing keys some how? rotating blinded keys rachet style)
+
+after built we have a quic substream which will do xchacha20 stream cipher at each hop using the nonce provided in the request. this substream can be used to send consensus data requests to the pivot router.
+
+over this stream once created we will use xchacha20 stream cipher at each hop on all stream data with the nounce provided in the request.
+
+#### path stream requests
+
+we can use path stream to request introsets, publish introsets, look up name hashes, look up services via service discovery, publish service discovery data, and any kinds of future anonymous requests we need to 
+
+### path update
+
+sent on an existing path to update metadata about the path.
+
+lets us close a path prematurely. could let us extend the lifetime of a path. could let us rewire an upstream router and path id at a hop ( this can be used to route around a bad snode without building a full new path ).
+
+(todo: if done wrong path rewiring introduces infinite path extension, we need to figure out how to ensure that cannot happen if we do hop rewriring).
+(todo: path lifetime extension limits are yet to be explored).
+
+lets us possibly set qos on a path (todo: think about per hop qos vs all hop qos, specifically hiding qos from each hop or using the same qos by having it be signed plaintext). 
+
+## lossy traffic
+
+one shot datagrams with a 24 byte nonce and tx/rx path id are sent best effort on an established path with the nonce mutated deterministically and symettric encryption applied. 
+    
--- a/docs/node2node.md
+++ b/docs/node2node.md
@@ -11,6 +11,8 @@
 
 direct node to node request/response messages sent on a bidi quic stream.
 
+a stream carries 1 type of request/response messages. we maintain 1 stream per type to each remote node with the exception of path stream which requires multiple instances of itself.
+
 the stream starts by sending 2 bytes identifying the stream kind of which there are the following:
 
 * network data lookup
--- a/docs/node2node.md
+++ b/docs/node2node.md
@@ -73,7 +73,7 @@ given we provide N nonces in the build we are opening up N path streams as we ma
 onion routed persisting stream connection to the pivot router for sending anonymised requests destined for the pivot router.
 
 streams are created for an existing path by opening a "path stream" stream providing the path's tx/rx id and 24 bytes nonce.
-on accepting the stream creation the node will open a "path stream" to their upstream router but with a nonce that is deterministically mutated `new_nonce = nonce ^ hash(nonce)`, this mutation reduces stream corrilation across hops.
+on accepting the stream creation the node will open a "path stream" to their upstream router but with a nonce that is deterministically mutated `new_nonce = nonce ^ hash(hop_ephemeral_secret)`, this mutation reduces stream corrilation across hops.
 on rejecting the stream creation the node will shutdown the read end of the stream, sending a reply signed by our identity key and then shutting down the write end of the stream (which fully closes the stream).
 if the accepting node is the pivot route it will write. 
 


docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+
+direct node to node request/response messages sent on a bidi quic stream.
+
+the stream starts by sending 2 bytes identifying the stream kind of which there are the following:
+
+* network data lookup
+* network data upsert
+* path build
+* path update
+* path stream
```
jagerman: It is not clear to me from this description how requests and streams interact.

Is the idea here that there is one stream per request?  Or one stream per *type* of request?

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+
+### network data lookup stream
+
+used for fetching of:
```
jagerman, line 26: Used how?  Right now this reads a bit like a wishlist, but to properly document the protocol, this page needs to evolve into something that completely describes the messages.  Ultimately, I want me or you or some other dev to be sit down with *only* this document and have enough information to produce an implementation that can interact with a remote Lokinet via the wire protocol.

docs/node2node.md
majestrate: the idea is that streams are for one type of request, but there is no
requirement that you only have one per type. this is especially true
with path stream type.

looking at this again i think it would be best to define which streams
would allow multiple instances per remote. e.g. a path stream
connection would have many instances to each remote, and per path
per remote, but for network data lookup/upsert you'd really only need
one per remote.

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+the plainext record contains:
+
+* next hop's pubkey
+* path lifetime
+* signing pubkey used by the path creator to sign updates
+* encryption pubkey to generate shared secret at the path
+* additional extra fields used in the future.
+
+
+(todo: i want to permit multiple kinds of pubkeys for forwards compat)
```
jagerman, line 60: I don't understand this todo: I don't see how you could be forwards compatible with a type of key you don't understand how to use.  To the extent that future versions of the protocol evolve to add other keys, you would either need to do that in a backwards-compatible way (i.e. providing alternatives, but still supporting the old key type for backwards compat), or in a breaking protocol change.

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+#### path build result
+
+(todo: finish this)
+
+### path streams
+
+onion routed persisting stream connection to the pivot router for sending anonymised requests destined for the pivot router.
+
+streams are created for an existing path by opening a "path stream" stream providing the path's tx/rx id and 24 bytes nonce.
+on accepting the stream creation the node will open a "path stream" to their upstream router but with a nonce that is deterministically mutated `new_nonce = nonce ^ hash(nonce)`, this mutation reduces stream corrilation across hops.
```
jagerman: > `new_nonce = nonce ^ hash(nonce)`

Just a hash is enough; the extra XOR of the nonce adds nothing.

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+on rejecting the stream creation the node will shutdown the read end of the stream, sending a reply signed by our identity key and then shutting down the write end of the stream (which fully closes the stream).
+if the accepting node is the pivot route it will write.
```
jagerman, line 78: It will write what?

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+### path streams
+
+onion routed persisting stream connection to the pivot router for sending anonymised requests destined for the pivot router.
+
+streams are created for an existing path by opening a "path stream" stream providing the path's tx/rx id and 24 bytes nonce.
+on accepting the stream creation the node will open a "path stream" to their upstream router but with a nonce that is deterministically mutated `new_nonce = nonce ^ hash(nonce)`, this mutation reduces stream corrilation across hops.
+on rejecting the stream creation the node will shutdown the read end of the stream, sending a reply signed by our identity key and then shutting down the write end of the stream (which fully closes the stream).
+if the accepting node is the pivot route it will write. 
+
+a "request stream path result" message has a flag for the result, zero for success and non zero for failure.
```
jagerman, line 80: What else does the message have?  What does it look like, how is it encoded, etc?

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+
+(todo: shrink tx/rx path id to be 8 bytes each?)
+
+#### path build result
+
+(todo: finish this)
+
+### path streams
+
+onion routed persisting stream connection to the pivot router for sending anonymised requests destined for the pivot router.
```
jagerman, line 73: I don't really understand why this is restricted to pivot routers: most of the requests listed in "path stream requests" seems like they would be invoked on any old remote router, not exclusively a pivot router.

docs/node2node.md
majestrate: clarification added

majestrate:
@jagerman i fell down an analytical rabbit hole last week on this so i forget if this was meant to be a full on technical protocol spec or just a high level thing.

remind me which one this was so i can dig into this again and so that that can be tracked.

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+on rejecting the stream creation the node will shutdown the read end of the stream, sending a reply signed by our identity key and then shutting down the write end of the stream (which fully closes the stream).
+if the accepting node is the pivot route it will write. 
+
+a "request stream path result" message has a flag for the result, zero for success and non zero for failure.
+
+(todo: if we use a ring signature on the "request path stream result" message can do a pop and shift on the pubkeys in the ring so that the original requestor can know who rejected/accepted without the intermediate hops knowing who did. would we have to generate derived keys here?)
+(todo: if we use normal signatures here would we want to blind the signing keys some how? rotating blinded keys rachet style)
+
+after built we have a quic substream which will do xchacha20 stream cipher at each hop using the nonce provided in the request. this substream can be used to send consensus data requests to the pivot router.
```
jagerman, line 85: It's not clear to me how this encryption mechanism will work:
- xchacha20 requires not just a nonce but also a key
- this is only describing how forward encryption works; where do the keys + nonces for the response stream encryption come from?
- how to deal with xchacha20's 64-byte block size?
- where does authentication happen (to detect/reject tampering)?
- can we recover from lost state?  (i.e. when we lose the IC counter for the stream)
- how and when do stream errors get detected and signaled?  (e.g. suppose intermediate router 2 has gone away).

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+we could also add an additional list of nonces on the path build that lets us open multiple path streams as we build our path.
+given we provide N nonces in the build we are opening up N path streams as we make our path build. see the path streams section below.
+
+(todo: shrink tx/rx path id to be 8 bytes each?)
+
+#### path build result
+
+(todo: finish this)
+
+### path streams
```
jagerman, line 71: This section left me with a lot of questions and took me a considerable amount of time to try to get my head around.  I feel like there is a lot of omitted details here in terms of describing how this gets initiated, how intermediate sub-streams get established, how data moves along it, how it gets authenticated, and so on.  There are *pieces* of all of those in here, but I feel a bit like I'm reading a book with every other sentence removed and trying to guess what the big picture is.

I need a diagram or a detailed description or both of how these paths work conceptually before you start in on the weeds of path ids, nonces, and encryption, because the picture I have in my head right now is a guess more than anything.

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+lets us close a path prematurely. could let us extend the lifetime of a path. could let us rewire an upstream router and path id at a hop ( this can be used to route around a bad snode without building a full new path ).
+
+(todo: if done wrong path rewiring introduces infinite path extension, we need to figure out how to ensure that cannot happen if we do hop rewriring).
+(todo: path lifetime extension limits are yet to be explored).
+
+lets us possibly set qos on a path (todo: think about per hop qos vs all hop qos, specifically hiding qos from each hop or using the same qos by having it be signed plaintext). 
+
+## lossy traffic
+
+one shot datagrams with a 24 byte nonce and tx/rx path id are sent best effort on an established path with the nonce mutated deterministically and symettric encryption applied.
```
jagerman, line 106: Elaboration needed.

docs/node2node.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### path stream requests
+
+we can use path stream to request introsets, publish introsets, look up name hashes, look up services via service discovery, publish service discovery data, and any kinds of future anonymous requests we need to 
+
+### path update
+
+sent on an existing path to update metadata about the path.
+
+lets us close a path prematurely. could let us extend the lifetime of a path. could let us rewire an upstream router and path id at a hop ( this can be used to route around a bad snode without building a full new path ).
```
jagerman, line 97: All of these capabilities, requests, responses, etc. needs to be fleshed out.

docs/node2node.md
```diff
@@ -0,0 +1,107 @@
...
+
+(todo: fully describe pop and shift mechanism)
+(todo: calcuate optimal data sizes)
+
+
+## streamed requests
+
+direct node to node request/response messages sent on a bidi quic stream.
+
+a stream carries 1 type of request/response messages. we maintain 1 stream per type to each remote node with the exception of path stream which requires multiple instances of itself.
```
jagerman, line 14: I don't really understand the point of using one stream per message type; since it isn't one stream per request, we need a request/response framework on top of the stream at which point you might as well just put the request type into the request itself.

docs/node2node.md
jagerman: Maybe a term such as "target router" would make more sense?

docs/node2node.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### path build result
+
+(todo: finish this)
+
+### path streams
+
+onion routed persisting stream connection to the pivot router for sending anonymised requests destined for the pivot router.
+
+streams are created for an existing path by opening a "path stream" stream providing the path's tx/rx id and 24 bytes nonce.
```
jagerman, line 75: The topology is not clear.

Suppose client A talks directly to 2 entry routers (E₁, E₂); through each entry router she builds random paths to three "target" routers (T₁, T₂, T₃ through E₁; and T₄, T₅, T₆ through E₂) to which she wants to make path stream requests; we'll suppose paths have two intermediate routers (X₁ through X₁₂).

Thus, from A's point of view, there are 6 paths:

- A-E₁-X₁-X₂-T₁
- A-E₁-X₃-X₄-T₂
- A-E₁-X₅-X₆-T₃
- A-E₂-X₇-X₈-T₄
- A-E₂-X₉-X₁₀-T₅
- A-E₂-X₁₁-X₁₂-T₆

What are all of the corresponding QUIC streams on the network, and how much state is attached to each?

docs/node2node.md
```diff
@@ -0,0 +1,107 @@
...
+* path build
+* path update
+* path stream
+
+### network data lookup stream
+
+used for fetching of:
+
+* pubkey list with unlock times if they are applicable.
+* pubkey list delta given last hash of full list.
```
jagerman, line 29: The purpose of these two endpoints is not obvious to me.  (Not saying there isn't one, just that it isn't obvious from just this description).

jagerman (commented):
This PR is a good start of the new protocol design.  There is, however, still considerable augmentation, elaboration, and details needed (both high-level and low-level) because it becomes an actionable plan.

I had a good bit of difficulty reviewing this document.  Right now it seems a bit confused in its presentation: parts of it are more like an outline or wishlist of features that don't exist yet (e.g. "look up wallet addresses staked into snode"), and other dive deep into the details (e.g. the "path streams" section) without enough description of the big picture of the feature to be able to fully understand and assess the details described.

I've left various suggestions below, as well as a handful of comments in the PR.

> @jagerman i fell down an analytical rabbit hole last week on this so i forget if this was meant to be a full on technical protocol spec or just a high level thing.

It needs to be both, but it *also* needs to be a proposal of how/why we are designing it the way it is.  (It's possible that this will eventually need to be split into multiple documents for different levels of understanding, but that is an editing task to be worried about (much) later in the process.)  More on this below.

# Suggestions

## Avoid scope creep

There's already too much scope creep in this document: for example, most of the "network data lookup stream" items are adding new features.  This PR is not the place for that.

That doesn't mean those features are not useful, it means those features are out of scope for the *initial* reimplementation of the network protocol.  The protocol needs to be able to be easily extensible to handle them, of course, and part of the protocol design needs to allow for protocol expansion, but for features that we don't have yet the protocol should only describe a generic request mechanism that is suitable for future request types.

## (Much) more elaboration

For the concrete requests that we do have, we need a concrete description of what the message looks like, all the data it contains, and so on.  For instance, what does a gossiped RC look like and what are the overall request, response, and error responses for updating one?  Or a path build request.  Or a snode lookup request.

## Avoid background knowledge

This document is currently heavily dependent on background knowledge of the existing protocol.  It should not be.  If pieces of the current protocol will be reused then write it out again, in full, because, unless there is a good reason why it must remain unchanged, it is just as much subject to critique as any *new* mechanisms being proposed here.  (And even if it must remain unchanged, it is still required here for this to be a complete reference of the protocol).

As I said above, we should aim to write this such that someone with *no* Lokinet knowledge can write a conforming client without looking at the current Lokinet implementation.

## What but also Why

The protocol design here needs more connection to high-level concepts so that this document can serve a purpose as not only the *what* of the protocol, but also the *why* of the protocol.  For instance, I have a lot of questions that I've left here in the path stream section, but I still don't have the big picture of *why* the stream state complexity of path streams is needed or what it allows that couldn't be done via another approach.

Abstractly:

1. We need to support feature X.
2. X requires an implementation of Y.
3. Here's Z.
4. Z is a proposed implementation of Y.

Reading this document right now I feel like I'm getting mostly just part 3, but understanding the protocol requires part 4, and proper discussion of the *design* of the protocol requires all of 2-4 (and maybe a bit of 1).

For a concrete example, see the pubkey list and pubkey list delta: in the current document there is no indication of what or who these are for or what problem having these endpoints solves.  I was going to comment on potential issues with download the entire list, or with the hash-diff mechanism, but then I realized my comments would only be based on my own guess of what these are here for.

## Review of current feature sets

I don't see anything about ONS lookups.

I don't know if there are other omissions, but would be worth going through again everything we need to support for *current* functionality to see if there are other pieces that are missing.

docs/node2node.md
majestrate: this operation needs to be commutative as it is done in both directions. when we get a nonce as the path builder from the pivot router we need to be able to do each operation in reverse order.

jagerman: I'm not understanding.

If I compute:

    X = N xor hash(N)

then how do I get N from X?

majestrate: that was my point, you can't.

jagerman: So then I'm back to my original comment: `X = N xor hash(N)` offers nothing over just `X = hash(N)`

majestrate: they both suffer from the same problem. it's a one way operation. when the sender of the original message is not the path creator we cannot obtain the original value of X because it's gone through a hash function. what we do right now with our existing scheme is `N = N ^ hash(path_shared_secret_symettric_key)`, note that the operation applied is `^`, and the `hash()` part of it is from the frame of reference of the next hop is such that they dont really care that a hash even happens. this is a tangent, i'm redoing this entire pr as i misunderstood the level of detail required.

PR #2170, opened 2023-05-08
[DRAFT] add embedded lokinet api for configuration.

majestrate:
draft proposal for a c api for dealing with lokinet's configs when in embedded mode.

feedback on what functionality is exposed would be greatly helpful.

cc @mpretty-cyro 
cc @jagerman 
cc @Bilb

### DIFF
--- a/include/lokinet/lokinet_config.h
+++ b/include/lokinet/lokinet_config.h
@@ -1,0 +1,93 @@
+#pragma once
+
+#include "lokinet_export.h"
+#include <stdio.h>
+#include <errno.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <string.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+  struct lokinet_config;
+
+  /// load a lokinet configuration from an open file handle.
+  struct lokinet_config* EXPORT
+  lokinet_config_from_file(FILE* f);
+
+  /// free a lokinet config we loaded.
+  void EXPORT
+  lokinet_config_free(struct lokinet_config*);
+
+  struct lokinet_config_iter;
+
+  /// open an iterator to iterate over all config values in a lokinet config in the section `sect`
+  struct lokinet_config_iter* EXPORT
+  lokinet_config_iterate_section(struct lokinet_config*, const char* sect);
+
+  /// fetch the next config value on this iterator.
+  /// returns false if there are no more items left to iterate over.
+  /// fills up opt_name and opt_val with option name and option value, or sets both to hold NULL if
+  /// at the last entry.
+  bool EXPORT
+  lokinet_config_iter_next(
+      struct lokinet_config_iter*, const char** opt_name, const char** opt_val);
+
+  /// close a previously openend lokinet_config_iter
+  void EXPORT
+  lokinet_config_iter_free(struct lokinet_config_iter*);
+
+  /// add to the lokinet config a config value for the section `sect` `opt_name=val`
+  void EXPORT
+  lokinet_config_add_opt(
+      struct lokinet_config*, const char* sect, const char* opt_name, const char* val);
+
+  /// write current config to an open file.
+  /// returns number of bytes written or -1 on error.
+  ssize_t EXPORT
+  lokinet_config_to_file(struct lokinet_config*, FILE* f);
+
+  /// helper function for reading config from file.
+  static struct lokinet_config* EXPORT
+  lokinet_config_read(const char* fname)
+  {
+    struct lokinet_config* conf;
+    FILE* f;
+    f = fopen(fname, "rb");
+    if (f == NULL)
+    {
+      fprintf(stderr, "failed to open %s: %s", fname, strerror(errno));
+      return NULL;
+    }
+    conf = lokinet_config_from_file(f);
+    if (conf == NULL)
+      fprintf(stderr, "failed to parse config file %s", fname);
+    fclose(f);
+    return conf;
+  }
+
+  /// helper function for writing config to file.
+  static bool EXPORT
+  lokinet_config_write(struct lokinet_config* conf, const char* fname)
+  {
+    ssize_t n;
+    FILE* f;
+    f = fopen(fname, "wb");
+    if (f == NULL)
+    {
+      fprintf(stderr, "failed to open %s: %s", fname, strerror(errno));
+      return NULL;
+    }
+    n = lokinet_config_to_file(conf, f);
+    if (n == -1)
+      fprintf(stderr, "failed to write config file %s", fname);
+    fclose(f);
+    return conf;
+  }
+
+#ifdef __cplusplus
+}
+#endif
--- a/include/lokinet/lokinet_context.h
+++ b/include/lokinet/lokinet_context.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lokinet_config.h"
 #include "lokinet_export.h"
 
 #include <stdbool.h>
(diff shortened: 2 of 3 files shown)


include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+  struct lokinet_config;
+
+  /// load a lokinet configuration from an open file handle.
+  struct lokinet_config* EXPORT
+  lokinet_config_from_file(FILE* f);
```
jagerman, line 19: Should be renamed `..._read_fd`

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+
+  /// free a lokinet config we loaded.
+  void EXPORT
+  lokinet_config_free(struct lokinet_config*);
+
+  struct lokinet_config_iter;
+
+  /// open an iterator to iterate over all config values in a lokinet config in the section `sect`
+  struct lokinet_config_iter* EXPORT
+  lokinet_config_iterate_section(struct lokinet_config*, const char* sect);
```
jagerman, line 29: This is a weird API:
- I *maybe* want an iterator over all config sections/settings.
- I definitely want to retrieve a single-valued option.
- I definitely want a way to retrieve multiple values for a multi-value option.

but what's the use case for providing an iterator over all the items of just one section?

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+
+  /// add to the lokinet config a config value for the section `sect` `opt_name=val`
+  void EXPORT
+  lokinet_config_add_opt(
+      struct lokinet_config*, const char* sect, const char* opt_name, const char* val);
+
+  /// write current config to an open file.
+  /// returns number of bytes written or -1 on error.
+  ssize_t EXPORT
+  lokinet_config_to_file(struct lokinet_config*, FILE* f);
```
jagerman, line 51: Rename `to_file` -> `write_fd`

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+  /// helper function for reading config from file.
+  static struct lokinet_config* EXPORT
+  lokinet_config_read(const char* fname)
+  {
+    struct lokinet_config* conf;
+    FILE* f;
+    f = fopen(fname, "rb");
+    if (f == NULL)
+    {
+      fprintf(stderr, "failed to open %s: %s", fname, strerror(errno));
```
jagerman, line 62: The caller needs to be able to know the error; having written it to stderr is not a nice way to communicate to a library caller.

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+    return conf;
+  }
+
+  /// helper function for writing config to file.
+  static bool EXPORT
+  lokinet_config_write(struct lokinet_config* conf, const char* fname)
+  {
+    ssize_t n;
+    FILE* f;
+    f = fopen(fname, "wb");
```
jagerman, line 78: Why is this "b"?

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+  ssize_t EXPORT
+  lokinet_config_to_file(struct lokinet_config*, FILE* f);
+
+  /// helper function for reading config from file.
+  static struct lokinet_config* EXPORT
+  lokinet_config_read(const char* fname)
+  {
+    struct lokinet_config* conf;
+    FILE* f;
+    f = fopen(fname, "rb");
```
jagerman, line 59: Why "b"?

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+  static bool EXPORT
+  lokinet_config_write(struct lokinet_config* conf, const char* fname)
+  {
+    ssize_t n;
+    FILE* f;
+    f = fopen(fname, "wb");
+    if (f == NULL)
+    {
+      fprintf(stderr, "failed to open %s: %s", fname, strerror(errno));
+      return NULL;
```
jagerman, line 82: `(bool) NULL` sounds ick.

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+    }
+    n = lokinet_config_to_file(conf, f);
+    if (n == -1)
+      fprintf(stderr, "failed to write config file %s", fname);
```
jagerman, line 86: More library stderr printing.

include/lokinet/lokinet_config.h
```diff
@@ -0,0 +1,93 @@
...
+
+  /// write current config to an open file.
+  /// returns number of bytes written or -1 on error.
+  ssize_t EXPORT
+  lokinet_config_to_file(struct lokinet_config*, FILE* f);
+
+  /// helper function for reading config from file.
+  static struct lokinet_config* EXPORT
+  lokinet_config_read(const char* fname)
+  {
```
jagerman, line 56: Why is this static + definition in the header?

include/lokinet/lokinet_context.h
```diff
@@ -13,10 +14,16 @@ extern "C"
...
-  struct lokinet_context* EXPORT
+  struct lokinet_config* EXPORT
   lokinet_context_new();
 
+  /// allocate a new lokinet context, passing in a loaded config, finalizing the config's loaded
+  /// values and injecting them into the lokinet context. if config is NULL we will load sensible
+  /// defaults. after calling lokinet_context_new the config can be released by calling
+  /// lokinet_config_free().
+  struct lokinet_context* EXPORT
+  lokinet_context_from_config(struct lokinet_config*);
```
jagerman, line 25: I think there should also be a convenience method taking a filename (to do the load + initialization from the file) for the likely most common use case.

llarp/lokinet_shared.cpp
```diff
@@ -193,17 +193,28 @@ namespace
   };
 }  // namespace
 
+struct lokinet_config
+{
+  std::shraed_ptr<llarp::Config> impl;
```
jagerman, line 198: typo

llarp/lokinet_shared.cpp
```diff
@@ -193,17 +193,28 @@ namespace
...
 
   std::shared_ptr<llarp::Context> impl = std::make_shared<Context>();
-  std::shared_ptr<llarp::Config> config = llarp::Config::EmbeddedConfig();
+  std::shared_ptr<llarp::Config> config;
 
   std::unique_ptr<std::thread> runner;
 
   int _socket_id = 0;
 
+  explicit lokinet_context(const std::shared_ptr<llarp::Config>& conf = nullptr) : config{conf}
```
jagerman, line 212: pass by value and move into `config`

llarp/lokinet_shared.cpp
```diff
@@ -525,6 +536,12 @@ extern "C"
     return new lokinet_context{};
   }
 
+  struct lokinet_context* EXPORT
+  lokinet_context_from_config(struct lokinet_config* cfg)
+  {
+    return new lokinet_context{cfg ? cfg->impl : nullptr};
```
jagerman, line 542: Since the only point of this function is to pass a config, passing NULL here seems like an error, not something that should be silently swallowed into a default embedded config.

jagerman (commented):
I left these a couple days ago but apparently forgot to click the button.

include/lokinet/lokinet_config.h
majestrate: comment was incomplete. the intended functionality would be passing in NULL to sect for all sections. but the iterator would need to also expose which section we are in. i'll add that in if it is not there.

```
/// if `sect` is NULL every section will be iterated over.
```



for single option vs multi option, an iterator approach covers both cases. how would the caller know that one specific config option didn't get multiple values provided? it doesn't really make sense to do a bunch of logic to handle each permutation of do we allow multiple config values, do we have multiple config values provided, do we want the first one if we only want one, etc. we just iterate over all config key/value parts provided and the caller can get what they want from it.

include/lokinet/lokinet_config.h
majestrate: i dont really like this function anyways as i feel like it's mostly pointless as the caller can just open a FILE themselves and propagate the error using their own mechanism with their own details they need.

PR #2174, opened 2023-05-11
make clients with strict-connect option explicitly connect to provided snode.

### LINKED ISSUE #2169: strict connect does not connect out
when we provide a `strict-connect` option we currently get stuck as we dont explicitly connect to those nodes on start.

majestrate:
make lokinet client with strict-connect option provided explicitly connect out to snodes provided. before it was just randomly chugging through peer selection until it found one we explicitly pinned. this rarely works.

* create new component for selecting edge snodes as client (llarp::consensus::EdgeSelector)
* require llarp::AbstractRouter own a llarp::consensus::EdgeSelector
* make outbound_session_maker.cpp use llarp::consensus::EdgeSelector owned by llarp::AbstractRouter

fixes #2169

### DIFF
--- a/llarp/CMakeLists.txt
+++ b/llarp/CMakeLists.txt
@@ -145,6 +145,7 @@ add_library(lokinet-config
 # lokinet-consensus is for deriving and tracking network consensus state for both service nodes and clients
 add_library(lokinet-consensus
   STATIC
+  consensus/edge_selector.cpp
   consensus/reachability_testing.cpp
 )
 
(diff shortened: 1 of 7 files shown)


majestrate:
this and #2164 should address the wire protocol regression introduced in 0.9.11, getting this merged would be a great thing to add to a point release before flipping the network into full on quic tls/1.3

majestrate:
@tewinget it may help in refactoring the existing path builder code to decompose the existing path selection code to have similar kinds of hop selectors as this PR does e.g. `aligned_hop_selector` which given an ordered set of previous hops, path length and the final aligned hop can select the next best hop.

PR #2204, opened 2023-08-29
Draft: Wire Protocol Overhaul, Message Handling Refactor

dr7ana:


llarp/util/buffer.hpp
```diff
@@ -20,7 +20,84 @@
...
+    {
+      return _bview.data();
+    }
+    const char*
+    vdata_at(size_t pos)
+    {
+      return _bview.data() + pos;
+    }
+
+    char
```
majestrate, line 102: should this be `char &` ? if it is not trying to get a reference it should be a `const` method maybe.

majestrate: but looks like it is just matching `llarp_buffer_t` so it's probably best to make it `const` is what my gut tells me.

llarp/router_contact.cpp
```diff
@@ -196,6 +235,7 @@ namespace llarp
     if (!bencode_write_uint64_entry(buf, "v", 1, version))
       return false;
 
+    // D We can delete this?
```
majestrate: force the new format of RCs that are encoded as a list not a dict and then you can.

majestrate: or only encode it when we do new style encoding (version not zero)

llarp/router/i_outbound_session_maker.hpp
```diff
@@ -45,10 +45,10 @@ namespace llarp
     virtual ~IOutboundSessionMaker() = default;
 
     virtual bool
-    OnSessionEstablished(ILinkSession* session) = 0;
+    OnSessionEstablished(AbstractLinkSession* session) = 0;
 
     virtual void
-    OnConnectTimeout(ILinkSession* session) = 0;
+    OnConnectTimeout(AbstractLinkSession* session) = 0;
```
majestrate: these should both use an l value reference not a pointer, the use of pointer is a code artifact.

llarp/router/abstractrouter.hpp
```diff
@@ -318,7 +320,8 @@ namespace llarp
 
     /// visit each connected link session
     virtual void
-    ForEachPeer(std::function<void(const ILinkSession*, bool)> visit, bool randomize) const = 0;
+    ForEachPeer(
+        std::function<void(const AbstractLinkSession*, bool)> visit, bool randomize) const = 0;
```
majestrate: this should use an l-value reference instead of pointer, use of pointer was a code artifact.

llarp/router/abstractrouter.hpp
```diff
@@ -100,7 +100,7 @@ namespace llarp
     virtual ~AbstractRouter() = default;
 
     virtual bool
-    HandleRecvLinkMessageBuffer(ILinkSession* from, const llarp_buffer_t& msg) = 0;
+    HandleRecvLinkMessageBuffer(AbstractLinkSession* from, const llarp_buffer_t& msg) = 0;
```
majestrate: ditto l value reference

llarp/messages/link_message.hpp
```diff
@@ -7,60 +7,56 @@
...
-    virtual bool
-    BEncode(llarp_buffer_t* buf) const = 0;
+    virtual ~AbstractLinkMessage() = default;
 
     virtual bool
-    HandleMessage(AbstractRouter* router) const = 0;
-
-    virtual void
-    Clear() = 0;
+    handle_message(AbstractRouter* router) const = 0;
```
majestrate: use of pointer here was a code artifact, best to use l-value referenece.

llarp/messages/link_message.hpp
```diff
@@ -7,60 +7,56 @@
...
   struct AbstractRouter;
 
+  inline std::array<std::string_view, 7> ABS_DICT_KEYS = {
+      "a"sv, "i"sv, "n"sv, "p"sv, "r"sv, "v"sv, "z"sv};
+
+  /// abstract base class for serialized messages
+  struct AbstractSerializable
+  {
+    virtual std::string
+    bt_encode() const = 0;
```
majestrate: i would highly suggest passing in a `std::vector<char> &` as an argument and have a non virtual helper method for returning `std::string` which just passes a local in and returns it. this way we can have a single vector that has reserved capacity to minimize heap thrashing. some of these messages, especially the relay up/down messages, are going to be insanely frequent, tens of thousands per event loop tick with potentially millions per second. having to do a heap alloc that many times is going to be noticable on the smaller end vps and small devices running as clients.

majestrate: actually, using a `llarp_buffer &` would be a way too.

jagerman: bt_{dict,list}_producer just needs a `.reserve()` to address this.

majestrate: fair. additionally my proposed change would explode scope way more than it needs to for something that is not needed yet.

majestrate: ....and i did not see that there is a `bt_encode(oxenc::bt_dict_producer&) const` method. this diff looks totally different on desktop web .... ugh

llarp/dht/messages/pubintro.cpp
```diff
@@ -9,194 +9,184 @@
...
+    return true;
+  }
+
+  void
+  PublishIntroMessage::bt_encode(oxenc::bt_dict_producer& btdp) const
+  {
+    try
+    {
+      btdp.append("A", "I");
+      btdp.append("T", introset.ToString());
```
majestrate: `btdp.append("I", introset.bt_encode());`

llarp/messages/common.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <llarp/util/buffer.hpp>
+
+#include <oxenc/bt.h>
+
+namespace llarp
+{
+  /// abstract base class for serialized messages
+  struct AbstractSerializable
```
majestrate, line 22: i'd love to have a bunch of free standing overloaded functions instead of using a pure virtual base class here. 

```

void
bt_encode(llarp_buffer&, const OurKindOfMessage & msg);
void
bt_encode(llarp_buffer&, const OurOtherMessage & msg);
```

majestrate: also a 
```
std::string bt_encode(const OurKindOfMessage & msg);
```

llarp/dht/messages/pubintro.cpp
majestrate: `ToString()` gives you printable information

dr7ana: `.bt_encode()` returns void and takes a `llarp_buffer_t` as a parameter

majestrate: a `llarp_buffer_t` or `llarp_buffer`? i feel like `llarp::llarp_buffer` should be `llarp::util::Buffer` or we use a different ideom in the naming scheme here, e.g. `EncodeDestBuffer` for encode and `DecodeSourceView` for decode.

i was podering if we could have a free standing helper function that would in combination with a template specialization return something we can pass into that `append` call.

something like a templated `constexpr bool llarp::is_bt_serializable` (that can become a c++20 concept later) and then have something like.

```
template<typename T, typename EncodeDestBuffer_t> 
EncodeDestBuffer_t bt_encode(const T & t)
{
    static_assert(llarp::is_bt_serializable<T>);
    static_assert(is_whatever_constraint_for_buffer_goes_here<Buffer_t>);
    EncodeDestBuffer_t tmp{};
    t.bt_encode(tmp);
    return tmp;
}
```

and then do a

```
btdp.append("I", bt_encode(introset))
```

where there is an overload for `append()` which will stick a raw buffer that is already encoded on to the end.

majestrate: having this distinct from `std::string` is how i like to prevent foot cannons with `ToString()` and why i like having `std::vector<std::byte>` used in stead for non printable binary serializable data. 

SSO would be a benefit too so a `std::basic_string<std::byte>` works too.

llarp/link/endpoint.cpp
```diff
@@ -0,0 +1,72 @@
+#include "endpoint.hpp"
+#include "link_manager.hpp"
+
+namespace llarp::link
+{
+  std::shared_ptr<link::Connection>
+  Endpoint::get_conn(const RouterContact& rc) const
+  {
+    for (const auto& [rid, conn] : conns)
```
tewinget: this could just be `conns.find(rc.pubkey)` I think

llarp/link/link_manager.cpp
```diff
@@ -9,27 +10,31 @@
...
+      : router{r}
+      , quic{std::make_unique<oxen::quic::Network>()}
+      , tls_creds{oxen::quic::GNUTLSCreds::make_from_ed_keys(
+            {reinterpret_cast<const char*>(router.encryption().data()), router.encryption().size()},
+            {reinterpret_cast<const char*>(router.encryption().toPublic().data()), size_t{32}})}
+      , ep{quic->endpoint(router.local), *this}
+  {}
+
+  std::shared_ptr<link::Connection>
+  LinkManager::get_compatible_link(const RouterContact& rc)
```
tewinget: This function was meant to get an endpoint which can connect to the given RC, but we decided we're just going to have one endpoint and it must be compatible, so this func goes away then.

dr7ana: yup its gone

llarp/link/endpoint.cpp
dr7ana: nice good point

llarp/link/link_manager.cpp
```diff
@@ -1062,14 +1062,178 @@ namespace llarp
   void
   LinkManager::handle_path_build(oxen::quic::message m)
   {
+    if (!router.path_context().AllowingTransit())
+    {
+      log::warning("got path build request when not permitting transit");
```
dr7ana: All these log statements without `link_cat` are going to error amigo :smile:

llarp/link/link_manager.cpp
```diff
@@ -1062,14 +1062,178 @@ namespace llarp
...
+        m.respond(serialize_response({{"STATUS", PathBuildMessage::BAD_FRAMES}}), true);
+        return;
+      }
+
+      const auto& hash = frame_list[0].at("hash");
+      auto& frame_str = frame_list[0].at("frame");
+
+      oxenc::bt_dict_consumer frame_dict{frame_str};
+
+      auto hop_info = frame_dict.require<ustring>("encrypted");
```
dr7ana: I like the idea of words as keys for clarity, but my one concern is uniformity with the other message types that use single letters. I could be convinced either way, but tbh my inclination is towards single letter keys. Wdyt?

llarp/link/link_manager.cpp
```diff
@@ -1062,14 +1062,178 @@ namespace llarp
...
+    }
     try
     {
-      oxenc::bt_dict_consumer btdc{m.body()};
+      // not using list_consumer here because we need to move the first (our) frame
+      // to the end and re-send, unless we're the terminal hop.  This could be done
+      // with list_consumer, but it would involve messy mucking around with the
+      // encoded list.  Revisit if optimization issue (shouldn't be).
+      auto frame_list = oxenc::bt_deserialize<oxenc::bt_list>(m.body());
+      if (frames.size() != path::MAX_LEN)
```
dr7ana: frame_list?

llarp/link/link_manager.cpp
```diff
@@ -1062,14 +1062,178 @@ namespace llarp
...
+            link_cat,
+            "Path build request has invalid number of records, ",
+            frame_list.size(),
+            "!=",
+            path::MAX_LEN);
+        m.respond(serialize_response({{"STATUS", PathBuildMessage::BAD_FRAMES}}), true);
+        return;
+      }
+
+      const auto& hash = frame_list[0].at("hash");
```
dr7ana: `Type 'std::list<oxenc::bt_value>' does not provide a subscript operator` at least on my editor?

llarp/link/link_manager.cpp
```diff
@@ -1062,14 +1062,178 @@ namespace llarp
...
+      auto& frame_str = frame_list[0].at("frame");
+
+      oxenc::bt_dict_consumer frame_dict{frame_str};
+
+      auto hop_info = frame_dict.require<ustring>("encrypted");
+      TunnelNonce nonce;
+      nonce.from_string_view(frame_dict.require<std::string_view>("nonce"));
+      Pubkey otherPubkey;
+      otherPubkey.from_string_view(frame_dict.require<std::string_view>("pubkey"));
```
dr7ana: Only the deserialize call, the dict consumer, and all the calls to `.require` or `.consume` need to be wrapped in a try/catch

llarp/link/link_manager.cpp
```diff
@@ -1062,14 +1062,178 @@ namespace llarp
...
+              digest.begin(), digest.end(), reinterpret_cast<const unsigned char*>(hash.data())))
+      {
+        log::info("HMAC mismatch on path build request");
+        m.respond(serialize_response({{"STATUS", PathBuildMessage::BAD_CRYPTO}}), true);
+        return;
+      }
+
+      // decrypt frame with our hop info
+      if (!crypto->xchacha20(
+              reinterpret_cast<unsigned char*>(hop_info.data()), hop_info.size(), shared, nonce))
```
dr7ana: This is already a ustring so luckily we dont need to reinterpret cast (since Jason's oxenc PR)

llarp/link/link_manager.cpp
```diff
@@ -1062,14 +1062,178 @@ namespace llarp
...
+      // extract pathIDs and check if zero or used
+      hop->info.txID.from_string_view(hop_dict.require<std::string_view>("txid"));
+      hop->info.rxID.from_string_view(hop_dict.require<std::string_view>("rxid"));
+      if (info.txID.IsZero() || info.rxID.IsZero())
+      {
+        log::info("Invalid PathID; PathIDs must be non-zero.");
+        m.respond(serialize_response({{"STATUS", PathBuildMessage::BAD_PATHID}}), true);
+        return;
+      }
+      hop->info.upstream.from_string_view(hop_dict.require<std::string_view>("next");
```
dr7ana: Could maybe just do:
`const auto& hop_info = hop->info`

Then every call is just 
`hop_info.txid` or `hop_info.rxid`, etc

llarp/util/aligned.hpp
```diff
@@ -242,7 +242,7 @@ namespace llarp
     }
 
     bool
-    from_string(std::string b)
+    from_string_view(std::string_view b)
```
dr7ana, line 245: Good idea!

jagerman:
Merging into `quic-wip` not because it's done, but because that's already a WIP branch, and ought to make new additions/updates/reviews easier of more code going into quic-wip.

PR #2213, opened 2023-10-16
Outbound/inbound message handling; compilation fixes

dr7ana:


daemon/lokinet.cpp
```diff
@@ -13,16 +12,38 @@
...
+#include <memory>
+#include <optional>
+#include <stdexcept>
+#include <utility>
+
+#include "CLI/Error.hpp"
+#include "CLI/Option.hpp"
+#include "CLI/impl/App_inl.hpp"
+#include "CLI/impl/Option_inl.hpp"
+#include "fmt/core.h"
```
jagerman: The includes above here (i.e. headers that aren't part of llarp) should be <...> instead of "...".

llarp/CMakeLists.txt
```diff
@@ -160,280 +243,140 @@ add_library(lokinet-layer-onion
...
 endif()
 
-# interface libraries for internal linkage
-add_library(lokinet-layers INTERFACE)
-add_library(lokinet-amalgum INTERFACE)
-
-# helper function to link a library to lokinet-base, enable lto, add to lokinet-amalgum and then link to other libs
-function(lokinet_link_lib libname)
-  message(DEBUG "created target: ${libname}")
-  enable_lto(${libname})
```
jagerman, line 268: This commit is currently dropping the `enable_lto` that gets applied to each library, but that needs to be brought back (LTO significantly reduces the size of release binaries, and optimizes performance as well).

We could do it in a different way, though.  Perhaps something something:

```cmake
# Add an internal lokinet static library target, enables LTO (if enabled) on the target,
# and links it to the common lokinet-base interface.
# Invoke with the target/library name (e.g. "lokinet-foo") and list of source files, e.g.
#     lokinet_add_library(lokinet-foo foo/source1.cpp foo/source2.cpp)
function(lokinet_add_library name)
    add_library(${libname} STATIC ${ARGN})
    target_link_libraries(${libname} PUBLIC lokinet-base)
    enable_lto(${libname})
endfunction()
```
and then changing all the `add_library(lokinet-foo STATIC sources...)` to `lokinet_add_library(lokinet-foo sources...)` to DRY out these two common additions.

daemon/lokinet.cpp
jagerman: Actually I think *most* of them should be.

The rule we decided on for this a couple years ago was to use `<a/b>` for a fully qualified headers -- certainly for external library headers, but also for fully-qualified llarp headers (e.g. `#include <llarp/util/logging.hpp>` rather than `#include "llarp/util/logging.hpp"`).

For includes in the same directory or a subdirectory, though, we use double-quotes, which try relative paths first (which angle includes do not).  So for instance, inside `llarp/foo.cpp` we would prefer `"#include "bar.hpp"` to `#include <llarp/bar.hpp>`, and likewise for includes from subdirectories, i.e. `#include "util/logging.hpp"` instead of `#include <llarp/util/logging.hpp>`.

jagerman: (Personally I'm fine with `#include "../whatever.hpp"` as well, e.g. so `llarp/foo/bar.hpp` could `#include "../util/logging.hpp"` rather than `#include <llarp/util/logging.hpp>` but Jeff and Tom vetoed that idea)

dr7ana: Yep you got it. These were all changed by IWYU haha, I was probably just going to write a little script to fix then manually verify rather than changing everything by hand

llarp/CMakeLists.txt
dr7ana: Cool, I will loop in `target_link_libraries(${libname} INTERFACE lokinet-amalgum)` as well

daemon/lokinet.cpp
```diff
@@ -350,8 +371,8 @@ namespace
   int
   lokinet_main(int argc, char** argv)
   {
-    if (auto result = Lokinet_INIT())
-      return result;
+    // if (auto result = Lokinet_INIT())
```
tewinget: Just remove; appears all this function did is `return 0`...

llarp/CMakeLists_old.txt
```diff
@@ -0,0 +1,421 @@
+include(Version)
```
tewinget: This file probably shouldn't be committed.

llarp/config/key_manager.cpp
```diff
@@ -93,23 +93,23 @@ namespace llarp
       // load identity key or create if needed
       auto identityKeygen = [](llarp::SecretKey& key) {
         // TODO: handle generating from service node seed
-        llarp::CryptoManager::instance()->identity_keygen(key);
+        llarp::crypto::identity_keygen(key);
```
tewinget, line 98: drop `llarp::`?

llarp/config/key_manager.cpp
```diff
@@ -93,23 +93,23 @@ namespace llarp
...
       };
       if (not loadOrCreateKey(m_idKeyPath, identityKey, identityKeygen))
         return false;
     }
 
     // load encryption key
     auto encryptionKeygen = [](llarp::SecretKey& key) {
-      llarp::CryptoManager::instance()->encryption_keygen(key);
+      llarp::crypto::encryption_keygen(key);
```
tewinget: drop `llarp::`?

llarp/crypto/crypto.cpp
```diff
@@ -113,28 +108,8 @@ namespace llarp
...
-      ntru_init(0);
-    }
-    int seed = 0;
-    randombytes(reinterpret_cast<unsigned char*>(&seed), sizeof(seed));
-    srand(seed);
-  }
-
   std::optional<AlignedBuffer<32>>
-  Crypto::maybe_decrypt_name(std::string_view ciphertext, SymmNonce nounce, std::string_view name)
+  crypto::maybe_decrypt_name(std::string_view ciphertext, SymmNonce nounce, std::string_view name)
```
tewinget: `s/nounce/nonce/`

llarp/crypto/crypto.hpp
```diff
@@ -17,12 +15,8 @@ namespace llarp
...
-  struct Crypto
+  namespace crypto
   {
-    Crypto();
-
-    ~Crypto() = default;
-
     /// decrypt cipherText given the key generated from name
     std::optional<AlignedBuffer<32>>
     maybe_decrypt_name(std::string_view ciphertext, SymmNonce nounce, std::string_view name);
```
tewinget: `s/nounce/nonce/`

llarp/crypto/types.cpp
```diff
@@ -2,14 +2,10 @@
...
 #include <llarp/util/file.hpp>
-
-#include <iterator>
-
 #include <oxenc/hex.h>
 
-#include <sodium/crypto_sign.h>
-#include <sodium/crypto_sign_ed25519.h>
-#include <sodium/crypto_scalarmult_ed25519.h>
+// #include <sodium/crypto_hash_sha512.h>
```
tewinget: remove if not needed?

llarp/dns/dns.hpp
```diff
@@ -1,30 +1,29 @@
 #pragma once
 
+#include <llarp/util/logging.hpp>
```
tewinget, line 3: I think IWYU got this one wrong...can't see how this header would need included here.

.clang-format
```diff
@@ -55,7 +55,20 @@ PointerAlignment: Left
...
+IncludeBlocks: Regroup
+IncludeCategories:
+  - Regex: '((")(.)+(.hpp"))'
+    Priority: 2
+  - Regex: '((<)(llarp)(/))'
+    Priority: 3
+  - Regex: '((<)[^(llarp)](.)+(.h))'
+    Priority: 4
+  - Regex: '((<)(\w)+(>))'
+    Priority: 5
```
jagerman:
```suggestion
  - Regex: '^".+\.h(pp)?"$'  # "quoted" includes of local headers
    Priority: 2
  - Regex: '^<llarp[./]'     # fully qualified llarp/ includes (and <llarp.hpp>)
    Priority: 3
  - Regex: '^<.*\.h(pp)?>$'  # external library includes
    Priority: 4
  - Regex: '^<\w+>$'         # stl includes
    Priority: 5
```

cmake/CMakeGraphVizOptions.cmake
```diff
@@ -0,0 +1,6 @@
+set(GRAPHVIZ_GRAPH_NAME "graph.dot" CACHE STRING "")
+# set(GRAPHVIZ_MODULE_LIBS OFF CACHE BOOL "")
+# set(GRAPHVIZ_EXTERNAL_LIBS OFF CACHE BOOL "")
```
jagerman: Should these two be removed?

crypto/libntrup/src/ntru.cpp
```diff
@@ -1,5 +1,7 @@
 #include <libntrup/ntru.h>
 
+#include "libntrup/ntru_api.h"
```
jagerman: Should be angle brackets

daemon/lokinet-bootstrap.cpp
```diff
@@ -1,26 +1,17 @@
...
 
+#include <iostream>
 #include <unordered_map>
 #include <unordered_set>
 
 #ifndef _WIN32
 #include <openssl/x509.h>
 #endif
 
-#include <sstream>
```
jagerman, line 18: Dropping this seems suspicious, because we're using a stringstream in this file

daemon/lokinet-vpn.cpp
```diff
@@ -1,25 +1,19 @@
...
-
-#include <CLI/App.hpp>
-#include <CLI/Formatter.hpp>
-#include <CLI/Config.hpp>
-#include "oxenmq/address.h"
+#include <vector>
 
 #ifdef _WIN32
 // add the unholy windows headers for iphlpapi
-#include <winsock2.h>
```
jagerman, line 17: `winsock2.h` is a special holy water include that changes other includes based on when it gets included, and so generally *has* to be before other Windows-specific includes.  Yay Windows.

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
 #include <llarp/config/config.hpp>  // for ensure_config
 #include <llarp/constants/version.hpp>
-#include <llarp.hpp>
-#include <llarp/util/lokinet_init.h>
 #include <llarp/util/exceptions.hpp>
-#include <llarp/util/fs.hpp>
-#include <llarp/util/str.hpp>
+#include <llarp/util/lokinet_init.h>
+
+#include <bits/chrono.h>
```
jagerman: Shouldn't be here; this is an stl-internal header; we should have `<chrono>` instead.

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
+#include "llarp/constants/files.hpp"
+#include "llarp/constants/platform.hpp"
+#include "llarp/ev/ev.hpp"
+#include "llarp/util/logging.hpp"
+#include "llarp/util/thread/threading.hpp"
+#include "oxen/log.hpp"
+#include "oxen/log/catlogger.hpp"
+#include "oxen/log/level.hpp"
+#include "oxen/log/ring_buffer_sink.hpp"
+#include "oxen/log/type.hpp"
```
jagerman: Angle includes

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
 #include <llarp/win32/service_manager.hpp>
+
 #include <dbghelp.h>
 #else
 #include <llarp/util/service_manager.hpp>
 #endif
 
+#include "CLI/Error.hpp"
+#include "CLI/impl/App_inl.hpp"
+#include "CLI/impl/Option_inl.hpp"
```
jagerman: Internal headers that shouldn't be here

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
+#include "CLI/Option.hpp"
+#include "llarp/constants/files.hpp"
+#include "llarp/constants/platform.hpp"
+#include "llarp/ev/ev.hpp"
+#include "llarp/util/logging.hpp"
+#include "llarp/util/thread/threading.hpp"
+#include "oxen/log.hpp"
+#include "oxen/log/catlogger.hpp"
+#include "oxen/log/level.hpp"
+#include "oxen/log/ring_buffer_sink.hpp"
```
jagerman: Shouldn't be here

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
+#include "CLI/impl/Option_inl.hpp"
+#include "CLI/Option.hpp"
+#include "llarp/constants/files.hpp"
+#include "llarp/constants/platform.hpp"
+#include "llarp/ev/ev.hpp"
+#include "llarp/util/logging.hpp"
+#include "llarp/util/thread/threading.hpp"
+#include "oxen/log.hpp"
+#include "oxen/log/catlogger.hpp"
+#include "oxen/log/level.hpp"
```
jagerman: These are implicit from `<oxen/log.hpp>`

daemon/lokinet.cpp
jagerman: `oxen/log/type.hpp` is implict from `oxen/log.hpp`

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
+#include "oxen/log/ring_buffer_sink.hpp"
+#include "oxen/log/type.hpp"
+
+#include <CLI/CLI.hpp>
+
+#include "fmt/core.h"
```
jagerman: <>

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
+#include "spdlog/common.h"
```
jagerman: Shouldn't be here

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
 #include <csignal>
-
-#include <string>
+#include <cstdlib>
+#include <exception>
+#include <filesystem>
+#include <future>
+#include <initializer_list>
```
jagerman: `-`

llarp/util/easter_eggs.cpp
```diff
@@ -1,39 +0,0 @@
...
-        MessageBoxA(NULL, text, title, MB_ICONHAND);
-        abort();
-      }
-    }
-  }
-};
-
-// i heckin love static initalization
-WineKiller lmao{};
-#endif
```
jagerman, line 33: The above should come back, but should move into `/daemon/lokinet.cpp` (so that it only fires on full lokinet, but not embedded lokinet).

jagerman: (It doesn't need to be in a struct; it could easily just be in the `main()` function)

daemon/lokinet.cpp
```diff
@@ -1,28 +1,52 @@
...
+#include "CLI/Error.hpp"
+#include "CLI/impl/App_inl.hpp"
+#include "CLI/impl/Option_inl.hpp"
+#include "CLI/Option.hpp"
+#include "llarp/constants/files.hpp"
+#include "llarp/constants/platform.hpp"
+#include "llarp/ev/ev.hpp"
+#include "llarp/util/logging.hpp"
+#include "llarp/util/thread/threading.hpp"
+#include "oxen/log.hpp"
```
jagerman: angle brackets

.clang-format
jagerman: Updated the second one so that `<llarp/...>` and `<llarp.hpp>` get grouped.

daemon/lokinet.cpp
dr7ana: fucking iwyu

llarp/crypto/types.hpp
```diff
@@ -1,10 +1,13 @@
...
 
 #include "constants.hpp"
+
 #include <llarp/router_id.hpp>
 #include <llarp/util/aligned.hpp>
-#include <llarp/util/types.hpp>
 #include <llarp/util/fs.hpp>
+#include <llarp/util/types.hpp>
+
+#include <sodium.h>
```
jagerman: This is putting sodium headers in nearly every compilation unit across the codebase, whereas sodium is meant to be just an internal implementation detail of the .cpp.

llarp/crypto/crypto.cpp
```diff
@@ -1,26 +1,13 @@
...
-#include <sodium/core.h>
-#include <sodium/crypto_generichash.h>
-#include <sodium/crypto_sign.h>
-#include <sodium/crypto_scalarmult.h>
-#include <sodium/crypto_scalarmult_ed25519.h>
-#include <sodium/crypto_stream_xchacha20.h>
-#include <sodium/crypto_core_ed25519.h>
-#include <sodium/crypto_aead_xchacha20poly1305.h>
-#include <sodium/randombytes.h>
-#include <sodium/utils.h>
```
jagerman, line 12: Breaking out the sodium headers this way was deliberate to only include the parts of libsodium we are using, rather than all of libsodium.

llarp/endpoint_base.hpp
```diff
@@ -1,29 +1,27 @@
...
-#include <set>
-#include "oxenc/variant.h"
-#include <quic.hpp>
 
 namespace llarp
 {
-  namespace quic
-  {
-    class TunnelManager;
-  }
```
jagerman, line 25: This should remain as a forward declaration (include of having the `#include <llarp/link/tunnel.hpp>`), so that lokinet-core-utils doesn't have to depend on lokinet-wire (where `link/tunnel.cpp` gets compiled).

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet.cpp
dr7ana: IWYU

daemon/lokinet-bootstrap.cpp
dr7ana: IWYU

daemon/lokinet-vpn.cpp
dr7ana: IWYU

crypto/libntrup/src/ntru.cpp
dr7ana: IWYU

daemon/lokinet.cpp
```diff
@@ -651,6 +632,17 @@ main(int argc, char* argv[])
...
   return lokinet_main(argc, argv);
 #else
+  if (auto hntdll = GetModuleHandle("ntdll.dll"))
+  {
+    if (GetProcAddress(hntdll, "wine_get_version"))
+    {
+      static const char* text = "Don't run lokinet in wine, aborting startup";
+      static const char* title = "Lokinet Wine Error";
+      MessageBoxA(NULL, text, title, MB_ICONHAND);
+      abort();
```
jagerman:
```suggestion
      std::abort();
```
(This is a bug from previous code this was copied from)

llarp/nodedb.cpp
```diff
@@ -243,7 +238,7 @@ namespace llarp
   size_t
   NodeDB::num_loaded() const
   {
-    return router.loop()->call_get([this]() { return entries.size(); });
+    return router.loop()->call_get([this]() -> size_t { return entries.size(); });
```
jagerman: This change doesn't seem necessary/wanted

llarp/pow.cpp
```diff
@@ -1,7 +1,6 @@
 #include "pow.hpp"
```
jagerman: I think we should just nuke this as I don't think it does anything useful.

llarp/service/outbound_context.hpp
```diff
@@ -10,29 +11,82 @@
...
+    /// for exits
+    void
+    send_packet_to_remote(std::string buf) override;
+
+    void
+    send_auth_async(std::function<void(std::string, bool)> resultHandler);
+
     void
-    Tick(llarp_time_t now) override;
+    Tick(std::chrono::milliseconds now) override;
```
jagerman: This (any many other in this file) should remain as `llarp_time_t` as they are representing timepoints (and even though `llarp_time_t` is a typedef to `ms`, it's useful also signals that a value is a time point, rather than a duration.  `llarp_time_t` isn't well implemented in that it really should be a proper time point, but that's a refactor for another day, and switching it to a plain duration is a step in the wrong direction here.

Looking at some of the others, it's possible that one or two of them should be converted when they are conceptually durations (and not conceptually time points), such as RTT().

llarp/service/protocol.cpp
```diff
@@ -39,26 +39,22 @@ namespace llarp::service
   ProtocolMessage::decode_key(const llarp_buffer_t& k, llarp_buffer_t* buf)
   {
     bool read = false;
-    if (!BEncodeMaybeReadDictInt("a", proto, read, k, buf))
-      return false;
+    // if (!BEncodeMaybeReadDictInt("a", proto, read, k, buf))
+    //   return false;
```
jagerman: Is this to be replaced or to be removed?

llarp/service/protocol.cpp
```diff
@@ -69,7 +65,7 @@ namespace llarp::service
 
     try
     {
-      btdp.append("a", static_cast<uint64_t>(proto));
+      // btdp.append("a", static_cast<uint64_t>(proto));
```
jagerman, line 67: replace or remove?

llarp/service/protocol.cpp
```diff
@@ -105,19 +98,18 @@ namespace llarp::service
...
```
jagerman, line 100: Replace or remove?

llarp/vpn/android.hpp
```diff
@@ -1,12 +1,13 @@
...
-#include <stdio.h>
-#include <unistd.h>
-
-#include "platform.hpp"
 #include "common.hpp"
+#include "platform.hpp"
+
 #include <llarp.hpp>
 
+#include <stdio.h>
```
jagerman:
```suggestion
#include <cstdio>
```

pybind/llarp/router.cpp
```diff
@@ -1,8 +1,8 @@
-#include <common.hpp>
-
 #include <llarp/router/router.hpp>
 #include <llarp/tooling/hive_router.hpp>
 
+#include <common.hpp>
```
jagerman:
```suggestion
#include "common.hpp"
```

pybind/llarp/router_id.cpp
```diff
@@ -1,7 +1,7 @@
-#include <common.hpp>
-
 #include <llarp/router_id.hpp>
 
+#include <common.hpp>
```
jagerman, line 3:
```suggestion
#include "common.hpp"
```

pybind/llarp/service/address.cpp
```diff
@@ -1,6 +1,7 @@
-#include <common.hpp>
 #include <llarp/service/address.hpp>
 
+#include <common.hpp>
```
jagerman, line 3:
```suggestion
#include "common.hpp"
```

pybind/llarp/tooling/router_event.cpp
```diff
@@ -1,14 +1,13 @@
...
 #include <llarp/tooling/dht_event.hpp>
 #include <llarp/tooling/path_event.hpp>
-#include <llarp/tooling/rc_event.hpp>
 #include <llarp/tooling/peer_stats_event.hpp>
+#include <llarp/tooling/rc_event.hpp>
+#include <llarp/tooling/router_event.hpp>
 
-#include <llarp/messages/relay_status.hpp>
-#include <llarp/path/path.hpp>
+#include <common.hpp>
```
jagerman, line 9:
```suggestion
#include "common.hpp"
```

pybind/llarp/tooling/router_hive.cpp
```diff
@@ -1,10 +1,10 @@
+#include <llarp.hpp>
+#include <llarp/router/router.hpp>
+#include <llarp/tooling/router_hive.hpp>
+
 #include <common.hpp>
```
jagerman, line 5:
```suggestion
#include "common.hpp"
```

jagerman (commented):
Done file-by-file review.

pybind/llarp/router.cpp
dr7ana: this change errors

pybind/llarp/router_id.cpp
dr7ana: this change errors

pybind/llarp/service/address.cpp
dr7ana: this change errors

pybind/llarp/tooling/router_event.cpp
dr7ana: this change errors

pybind/llarp/tooling/router_hive.cpp
dr7ana: this change errors

llarp/service/protocol.cpp
dr7ana: unsure, i thought we discussed this proto as a member variable being replaced or re-implemented in some way? also aren't these objects going the way of the dinosaurs like the other message objects

maybe i'm falsely remembering, will mention this tomorrow in standup

llarp/service/protocol.cpp
dr7ana: ^^ above

llarp/service/protocol.cpp
dr7ana: ^^ same as above

jagerman:
I pushed some changes here to bump the oxen-encoding submodule (to fix `require` not working depending on what's installed on the system), and a fix for the systemd linking.

Test code is still broken; I think we should probably just disable that from CI as a temporary measure.

PR #2214, opened 2023-10-17
Path build and onioned messages

tewinget:
WIP

### DIFF
--- a/llarp/link/link_manager.hpp
+++ b/llarp/link/link_manager.hpp
@@ -286,23 +286,21 @@ namespace llarp
 
    private:
     // DHT messages
-    void handle_find_name(oxen::quic::message);      // relay
-    void handle_find_intro(oxen::quic::message);     // relay
-    void handle_publish_intro(oxen::quic::message);  // relay
-    void handle_find_router(oxen::quic::message);    // relay + path
+    void handle_find_name(oxen::quic::message, const RouterID& from);      // relay
+    void handle_find_intro(oxen::quic::message, const RouterID& from);     // relay
+    void handle_publish_intro(oxen::quic::message, const RouterID& from);  // relay
+    void handle_find_router(oxen::quic::message, const RouterID& from);    // relay + path
 
     // Path messages
-    void handle_path_build(oxen::quic::message);     // relay
-    void handle_path_confirm(oxen::quic::message);   // relay
-    void handle_path_latency(oxen::quic::message);   // relay
-    void handle_path_transfer(oxen::quic::message);  // relay
-    void handle_relay_commit(oxen::quic::message);   // relay
-    void handle_relay_status(oxen::quic::message);   // relay
+    void handle_path_build(oxen::quic::message, const RouterID& from);     // relay
+    void handle_path_confirm(oxen::quic::message, const RouterID& from);   // relay
+    void handle_path_latency(oxen::quic::message, const RouterID& from);   // relay
+    void handle_path_transfer(oxen::quic::message, const RouterID& from);  // relay
 
     // Exit messages
-    void handle_obtain_exit(oxen::quic::message);  // relay
-    void handle_update_exit(oxen::quic::message);  // relay
-    void handle_close_exit(oxen::quic::message);   // relay
+    void handle_obtain_exit(oxen::quic::message, const RouterID& from);  // relay
+    void handle_update_exit(oxen::quic::message, const RouterID& from);  // relay
+    void handle_close_exit(oxen::quic::message, const RouterID& from);   // relay
 
     std::unordered_map<std::string, void (LinkManager::*)(oxen::quic::message)> rpc_commands = {
         {"find_name", &LinkManager::handle_find_name},
--- a/llarp/path/path.cpp
+++ b/llarp/path/path.cpp
@@ -107,8 +107,33 @@ namespace llarp::path
   Path::send_path_control_message(
       std::string method, std::string body, std::function<void(oxen::quic::message m)> func)
   {
+    oxenc::bt_dict_producer btdp;
+    btdp.append("METHOD", method);
+    btdp.append("BODY", body);
+    auto payload = std::move(btdp).str();
+    auto* payload_ptr = reinterpret_cast<unsigned char*>(payload.data());
+
+    // TODO: old impl padded messages if smaller than a certain size; do we still want to?
+
+    auto crypto = CryptoManager::instance();
+
+    TunnelNonce nonce;
+    outer_nonce.Randomize();
+    for (const auto& hop : hops)
+    {
+      // do a round of chacha for each hop and mutate the nonce with that hop's nonce
+      CryptoManager::instance()->xchacha20(payload_ptr, hop.shared, nonce);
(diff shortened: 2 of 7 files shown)


llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
...
+    btdp.append("BODY", body);
+    auto payload = std::move(btdp).str();
+    auto* payload_ptr = reinterpret_cast<unsigned char*>(payload.data());
+
+    // TODO: old impl padded messages if smaller than a certain size; do we still want to?
+
+    auto crypto = CryptoManager::instance();
+
+    TunnelNonce nonce;
+    outer_nonce.Randomize();
```
dr7ana, line 121: Where does this come from?

llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
   Path::send_path_control_message(
       std::string method, std::string body, std::function<void(oxen::quic::message m)> func)
   {
+    oxenc::bt_dict_producer btdp;
+    btdp.append("METHOD", method);
+    btdp.append("BODY", body);
+    auto payload = std::move(btdp).str();
```
dr7ana, line 113: Might be good to wrap the BTDP in braces before you std move it out of scope like you did in the other logic

llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
...
+    outer_nonce.Randomize();
+    for (const auto& hop : hops)
+    {
+      // do a round of chacha for each hop and mutate the nonce with that hop's nonce
+      CryptoManager::instance()->xchacha20(payload_ptr, hop.shared, nonce);
+      nonce ^= hop.nonceXOR;
+    }
+
+    oxenc::bt_dict_producer outer_dict;
+    outer_dict.append("PATHID", TXID().ToView());
```
dr7ana, line 130: Small note: would be best to alphabetize these keys (same with the ones at the top)

llarp/path/pathbuilder.cpp
```diff
@@ -465,17 +465,44 @@ namespace llarp
...
+      // Path build fail and success are handled poorly at best and changing how we
+      // handle these responses as well as how we store and use Paths as a whole might
+      // be worth doing sooner rather than later.  Leaving some TODOs below where fail
+      // and success live.
+      auto response_cb = [self](oxen::quic::message m) {
+        try
+        {
+          if (not m)
+          {
+            if (m.timed_out)
```
dr7ana, line 478: This is awkward

llarp/link/link_manager.hpp
```diff
@@ -286,23 +286,21 @@ namespace llarp
...
    private:
     // DHT messages
-    void handle_find_name(oxen::quic::message);      // relay
-    void handle_find_intro(oxen::quic::message);     // relay
-    void handle_publish_intro(oxen::quic::message);  // relay
-    void handle_find_router(oxen::quic::message);    // relay + path
+    void handle_find_name(oxen::quic::message, const RouterID& from);      // relay
+    void handle_find_intro(oxen::quic::message, const RouterID& from);     // relay
+    void handle_publish_intro(oxen::quic::message, const RouterID& from);  // relay
+    void handle_find_router(oxen::quic::message, const RouterID& from);    // relay + path
```
dr7ana, line 292: Why did you add this to all of them?

llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
   Path::send_path_control_message(
       std::string method, std::string body, std::function<void(oxen::quic::message m)> func)
   {
+    oxenc::bt_dict_producer btdp;
+    btdp.append("METHOD", method);
+    btdp.append("BODY", body);
+    auto payload = std::move(btdp).str();
+    auto* payload_ptr = reinterpret_cast<unsigned char*>(payload.data());
```
dr7ana, line 114: There's no benefit to doing this outside of the function call as its only used once

llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
...
+    // TODO: old impl padded messages if smaller than a certain size; do we still want to?
+
+    auto crypto = CryptoManager::instance();
+
+    TunnelNonce nonce;
+    outer_nonce.Randomize();
+    for (const auto& hop : hops)
+    {
+      // do a round of chacha for each hop and mutate the nonce with that hop's nonce
+      CryptoManager::instance()->xchacha20(payload_ptr, hop.shared, nonce);
```
dr7ana, line 125: This take 4 parameters

llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
...
+
+    oxenc::bt_dict_producer outer_dict;
+    outer_dict.append("PATHID", TXID().ToView());
+    outer_dict.append("NONCE", nonce.ToView());
+    outer_dict.append("PAYLOAD", payload);
+
+
     return router.send_control_message(
-        upstream(), std::move(method), std::move(body), std::move(func));
+        upstream(), "path_control", std::move(outer_dict.str()), std::move(func));
```
dr7ana, line 136: `std::move(outer_dict).str()`

llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
...
   {
+    oxenc::bt_dict_producer btdp;
+    btdp.append("METHOD", method);
+    btdp.append("BODY", body);
+    auto payload = std::move(btdp).str();
+    auto* payload_ptr = reinterpret_cast<unsigned char*>(payload.data());
+
+    // TODO: old impl padded messages if smaller than a certain size; do we still want to?
+
+    auto crypto = CryptoManager::instance();
```
dr7ana, line 118: Unused

llarp/path/path.cpp
dr7ana, line 136: Also, if you're going to create a new endpoint "path_control", whatever `std::function` is passed as `func` needs to be captured by a lambda in the call to `router.send_control_message(...)`

That way any path-specific logic doesn't have to be implemented every single place a function that calls `send_path_control_message` is called, like `find_router` or `find_name`

llarp/path/path.cpp
```diff
@@ -107,8 +107,33 @@ namespace llarp::path
   Path::send_path_control_message(
       std::string method, std::string body, std::function<void(oxen::quic::message m)> func)
   {
+    oxenc::bt_dict_producer btdp;
+    btdp.append("METHOD", method);
+    btdp.append("BODY", body);
```
jagerman, line 112: Alphabetical order!

llarp/path/path.cpp
jagerman: Not just best, it's required.  (It will fail an assert in a debug build)

llarp/path/path.cpp
tewinget: typo, needed 2 different nonces in path build message construction and messed up copying that bit over.

tewinget:
supplanted by #2216

PR #2215, opened 2023-10-17
Remove stuff

jagerman:
Remove stuff that we don't want or need and isn't worth supporting.

### DIFF
--- a/llarp/router/router.cpp
+++ b/llarp/router/router.cpp
@@ -571,39 +571,15 @@ namespace llarp
   }
 
   bool
-  Router::update_rc(bool rotateKeys)
+  Router::update_rc()
   {
     SecretKey nextOnionKey;
     RouterContact nextRC = router_contact;
-    if (rotateKeys)
-    {
-      CryptoManager::instance()->encryption_keygen(nextOnionKey);
-      std::string f = encryption_keyfile.string();
-      // TODO: use disk worker
-      if (nextOnionKey.SaveToFile(f.c_str()))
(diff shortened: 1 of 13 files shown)


tewinget (approved):

llarp/router_contact.hpp
```diff
@@ -165,18 +162,9 @@ namespace llarp
     bool
     decode_key(const llarp_buffer_t& k, llarp_buffer_t* buf);
 
-    bool
-    HasNick() const;
-
-    std::string
-    Nick() const;
```
dr7ana, line 172: RIP nick

llarp/router/router.cpp
```diff
@@ -1004,22 +998,22 @@ namespace llarp
...
 
-      const auto pk = conn.remote_rc.pubkey;
+      for_each_connection([this, &close_peers](link::Connection& conn) {
+        const auto& pk = conn.remote_rc.pubkey;
 
-      if (conn.remote_is_relay and not _rc_lookup_handler.is_session_allowed(pk))
-        close_peers.emplace(pk);
-    });
+        if (conn.remote_is_relay and not _rc_lookup_handler.is_session_allowed(pk))
+          close_peers.insert(pk);
```
dr7ana, line 986: insert instead of emplace because it's a const reference to an already a constructed object, right?

Would using emplace copy construct a new object? Or would emplace just copy-elision it into the unordered set?

dr7ana (approved):

PR #2216, opened 2023-10-23
Path build and onioned messages

tewinget:
WIP

depends #2213


llarp/link/link_manager.cpp
```diff
@@ -135,13 +136,31 @@ namespace llarp
...
   {
-    for (const auto& [name, func] : rpc_commands)
-    {
-      s->register_command(name, [this, f = func](oxen::quic::message m) {
-        _router.loop()->call([this, func = f, msg = std::move(m)]() mutable {
-          std::invoke(func, this, std::move(msg));
-        });
-      });
+    assert(ep.connid_map.count(s->conn_id()));
+    RouterID rid = ep.connid_map[s->conn_id()];
```
dr7ana: const reference?

llarp/link/link_manager.hpp
```diff
@@ -287,18 +286,22 @@ namespace llarp
 
    private:
     // DHT messages
-    void handle_find_name(oxen::quic::message);      // relay
-    void handle_find_intro(oxen::quic::message);     // relay
-    void handle_publish_intro(oxen::quic::message);  // relay
-    void handle_find_router(oxen::quic::message);    // relay + path
+    void
+    handle_find_name(std::string_view body, std::function<void(std::string)> respond);  // relay
```
dr7ana, line 285: Making a note to discuss: mis-matched function signatures for message and response handling (some are string/bool, some just string)

llarp/path/abstracthophandler.hpp
```diff
@@ -22,6 +22,14 @@ namespace llarp
...
+    std::string
+    make_onion_payload(
+        const TunnelNonce& nonce, const PathID_t& path_id, const std::string_view& inner_payload);
+    std::string
+    make_onion_payload(
+        const TunnelNonce& nonce, const PathID_t& path_id, const ustring_view& inner_payload);
+
     struct AbstractHopHandler
     {
       using TrafficEvent_t = std::pair<std::vector<byte_t>, TunnelNonce>;
```
dr7ana: I know you didn't do this, but can you get rid of this stupid typedef please

llarp/path/path.cpp
```diff
@@ -71,65 +69,93 @@ namespace llarp::path
...
-          if (m)
+        std::move(outer_payload),
+        [response_cb = std::move(func), weak = weak_from_this()](oxen::quic::message m) {
+          auto self = weak.lock();
+          if (not self)
+            return;
+          if (m.timed_out)
           {
-            // do path hop logic here
+            response_cb(""s, true);
```
dr7ana: Maybe we can find a meaningful error message here rather than an empty sequence. It would also save the superfluous creation of an empty string

llarp/path/path_context.hpp
```diff
@@ -25,149 +25,130 @@ namespace llarp
...
+    }
+  };
+}  // namespace std
 
-      /// called from router tick function
-      void
-      ExpirePaths(llarp_time_t now);
+namespace llarp::path
+{
+  using TransitHop_ptr = std::shared_ptr<TransitHop>;
```
dr7ana: Let's get rid of this typedef

llarp/path/path_context.hpp
```diff
@@ -25,149 +25,130 @@ namespace llarp
...
+    GetTransitHop(const RouterID&, const PathID_t&);
 
-      std::optional<std::weak_ptr<TransitHop>>
-      TransitHopByInfo(const TransitHopInfo&);
+    PathSet_ptr
+    GetLocalPathSet(const PathID_t& id);
 
-      std::optional<std::weak_ptr<TransitHop>>
-      TransitHopByUpstream(const RouterID&, const PathID_t&);
+    using EndpointPathPtrSet = std::set<Path_ptr, ComparePtr<Path_ptr>>;
```
dr7ana: and this one

llarp/path/path_context.hpp
```diff
@@ -25,149 +25,130 @@ namespace llarp
...
-      GetLocalPathSet(const PathID_t& id);
+    bool
+    HopIsUs(const RouterID& k) const;
 
-      using EndpointPathPtrSet = std::set<Path_ptr, ComparePtr<Path_ptr>>;
-      /// get a set of all paths that we own who's endpoint is r
-      EndpointPathPtrSet
-      FindOwnedPathsWithEndpoint(const RouterID& r);
+    void
+    AddOwnPath(PathSet_ptr set, Path_ptr p);
```
dr7ana, line 108: And definitely `Path_ptr` too

llarp/service/endpoint.cpp
```diff
@@ -139,63 +139,52 @@ namespace llarp::service
...
+                        poker->put_up();
 
-                        result_handler(true, result);
-                      }
+                      result_handler(true, result);
+                    }
 
-                      result_handler(false, result);
-                    };
+                    result_handler(false, result);
```
dr7ana: This needs to be in an `else` block so it doesn't roll from calling `result_handler` once into the second invocation

llarp/service/endpoint.cpp
```diff
@@ -878,10 +841,29 @@ namespace llarp::service
...
+      std::string name{};
+      try
+      {
+        oxenc::bt_dict_consumer btdc{resp};
+        name = btdc.require<std::string>("NAME");
+      }
+      catch (...)
+      {
+        log::warning(link_cat, "Failed to parse find name response!");
+        func(""s, false);
```
dr7ana: Ditto on the empty strings here

llarp/service/endpoint.cpp
```diff
@@ -897,18 +879,35 @@ namespace llarp::service
...
 
-    if (routers.find(rid) == routers.end())
-    {
-      auto path = GetEstablishedPathClosestTo(rid);
-      path->find_router("find_router", func);
-      return true;
-    }
+    auto response_cb = [func = std::move(func)](std::string resp, bool timeout) {
+      if (timeout)
+        func(RouterContact{}, false);
```
dr7ana, line 873: We have namespaced errors for timeouts and things like that

llarp/service/endpoint.cpp
```diff
@@ -1367,30 +1366,39 @@ namespace llarp::service
     // address once.
     bool hookAdded = false;
 
+    auto got_it = std::make_shared<bool>(false);
```
dr7ana, line 1355: I'm not clear what the point of this is

llarp/service/endpoint.hpp
```diff
@@ -235,12 +235,11 @@ namespace llarp
 
       // "find router" via closest path
       bool
-      lookup_router(RouterID router, std::function<void(oxen::quic::message)> func = nullptr);
+      lookup_router(RouterID router, std::function<void(RouterContact, bool)> func = nullptr);
```
dr7ana, line 238: This signature can match the others, string/bool

llarp/messages/common.hpp
```diff
@@ -18,6 +18,24 @@ namespace
...
+      return oxenc::bt_serialize(supplement);
+    }
+
+    // ideally STATUS is the first key in a bt-dict, so use a single, early ascii char
+    inline auto STATUS_KEY = "!"s;
+    inline auto STATUS_OK = "OK"sv;
+    inline auto STATUS_TIMEOUT = "TIMEOUT"sv;
+    inline auto STATUS_ERROR = "ERROR"sv; // generic, unknown error (e.g. onion decrypt throw)
+    inline auto TIMEOUT_BT_DICT = serialize_response({{STATUS_KEY, STATUS_TIMEOUT}});
+    inline auto ERROR_BT_DICT = serialize_response({{STATUS_KEY, STATUS_ERROR}});
```
dr7ana: I like the idea of doing this! What do you think of this namespacing?
```
namespace llarp
{
    namespace status
    {
        inline constexpr auto KEY = "!"sv;
        inline constexpr auto OK = "OK"sv;
        ...
    }
}
```
Then it would be 
```
m.respond(serialize_response({{status::KEY, CloseExitMessage::EXCEPTION}}, true);
```

It may also be useful to take `STATUS_OK` (or `status::OK` depending on how you namespace it) and do what you did with `TIMEOUT_BT_DICT` and `ERROR_BT_DICT`, since it's probably always key-paired as `"!":"OK"`

llarp/crypto/crypto.cpp
```diff
@@ -81,13 +81,13 @@ namespace llarp
...
   static bool
   dh_server_priv(
-      llarp::SharedSecret& shared, const PubKey& pk, const SecretKey& sk, const TunnelNonce& n)
+      llarp::SharedSecret& shared, const PubKey& pk, const SecretKey& sk, const SymmNonce& n)
   {
     llarp::SharedSecret dh_result;
 
     if (dh(dh_result, pk, sk.toPublic(), pk.data(), sk))
     {
-      return crypto_generichash_blake2b(shared.data(), 32, n.data(), 32, dh_result.data(), 32)
```
dr7ana, line 90: General thought in regards to most of these crypto functions that jason brought up the other day...

We should be checking sizes on the objects and throwing if it's not correct

llarp/path/path.cpp
```diff
@@ -454,73 +436,6 @@ namespace llarp::path
...
-        m_TXRate += msg.enc.size();
-      }
-      else
-      {
-        LogDebug("failed to send upstream to ", upstream());
-      }
-    }
-    r->TriggerPump();
-  }
-
```
dr7ana, line 440: love the swaths of red

llarp/path/transit_hop.cpp
```diff
@@ -110,157 +105,7 @@ namespace llarp::path
 
     return true;
   }
-
-  void
-  TransitHop::DownstreamWork(TrafficQueue_t msgs, Router* r)
```
dr7ana, line 87: goodbye downstream work aka upstream pump aka flush downstream aka handleallupstream

llarp/path/path_context.hpp
tewinget: agreed, but leaving this one specifically for a future PR as many of the places it's used should be a reference instead of copying a shared_ptr, and many others will likely go away entirely.  I can make that change in this PR if you want though.

llarp/messages/common.hpp
tewinget: I am indifferent to namespacing the statuses (since there are only a couple), but I can if you want.  STATUS_OK as a dict could be useful I guess; I was considering the case where there's more meat to the reply than that, but if a reply is a trivial "ok, done" rather than a response (e.g. an intro set) then I suppose having a pre-fab dict is ok.

llarp/crypto/crypto.cpp
tewinget: Agreed.  For most of the code this is achieved via type safety, but there may yet be bits which call the crypto functions which take bare pointers directly.  I don't know if it's in scope for this PR to seek out all of those cases, but I can if you want.

llarp/messages/common.hpp
dr7ana: I'm sortof indifferent, was more floating some ideas since I was doing something similar with storage server. I'm fine with altering these types of things as we go, depending on what we use the most/what we don't use at all?

llarp/crypto/crypto.cpp
dr7ana: Yeah i agree, though maybe we shouldn't rely too hard on the type safety lol. As to the pointers, we should const the ones that correspond to const references to objects in the other overloads, ex:
```C++
void do_crypto_thangs(const SecretKey& sk, const PubKey& pk, ...);

void do_crypto_thangs(const uint8_t* sk, size_t sk_size, const uint8_t* pk, size_t pk_size, ...);
```

I'll make a jira ticket RE the things we need to do to `crypto/` generally (make the uint8_t pointers const where needed, etc), and that can be it's own standalone PR? It's probably a solid 1-2 day task, so either you or i can knock it out when it makes sense between tasks

tewinget: I think it should be a separate PR, yeah.  Sounds good.

llarp/messages/common.hpp
tewinget: > I'm fine with altering these types of things as we go, depending on what we use the most/what we don't use at all?

You're starting to sound like me.  And yeah, that's how I see it as well.

llarp/link/link_manager.cpp
```diff
@@ -534,7 +536,7 @@ namespace llarp
     catch (const std::exception& e)
     {
       log::warning(link_cat, "Exception: {}", e.what());
-      respond(serialize_response({{"STATUS", FindNameMessage::EXCEPTION}}));
+      respond(serialize_response({{messages::STATUS_KEY, FindNameMessage::EXCEPTION}}));
```
dr7ana: Minor: Why do we (me included lol) still call `serialize_response` and pass two constexpr's in a dict as a parameter? Why dont we just
```C++
// messages.hpp
namespace messages {
    inline const auto STATUS_OK = serialize_response({{"!", "OK"}});
    inline const auto STATUS_EXCEPTION = serialize_response({{"!", "EXCEPTION"}});
    ....
}
```
To then invoke like:
```C++
// link_manager.cpp
respond(messages::STATUS_OK);
...
m.respond(messages::STATUS_EXCEPTION);
```
I had namespaced the messages individually when I first laid this out, but we really don't need the string `"EXCEPTION"` repeated, and the more I see `serialize_response` the more I don't want to see it lol.

Though for some cases, like `PublishIntroMessage::INVALID_ORDER`, the namespacing is useful so maybe there's a middle road where we bring as much as we can into `llarp/messages/common.hpp`.

Maybe we can even add a `bool` parameter "serialize_response" to the libquic `message::respond` method that defaults to `true`, where it does that automatically?

llarp/crypto/crypto.hpp
```diff
@@ -23,36 +23,30 @@ namespace llarp
 
     /// xchacha symmetric cipher
     bool
-    xchacha20(uint8_t*, size_t size, const SharedSecret&, const TunnelNonce&);
+    xchacha20(uint8_t*, size_t size, const SharedSecret&, const SymmNonce&);
```
dr7ana, line 26: Minor:
So since these nonce types are declared like this:
```C++
using TunnelNonce = AlignedBuffer<TUNNONCESIZE>;
using SymmNonce = AlignedBuffer<NONCESIZE>;
```
I feel like it could be a lot clearer if we did this:
```C++
inline constexpr uint32_t TUNNEL{32}; // this is TUNNONCESIZE
inline constexpr uint32_t SYMMETRIC{24}; // this is NONCESIZE

template <typename sz>
using Nonce = AlignedBuffer<sz>; // or we could add nonce-specfic methods and make this a struct
```
Then the type would be `Nonce<TUNNEL>` or `Nonce<SYMMETRIC>`. Not sure if this might cause issues with different types that have same sizes, but maybe we can figure out a better solution to the needless typedefing haha

llarp/link/link_manager.cpp
tewinget: Agreed, generic errors should probably be shared.  Should I chuck that in this PR?

llarp/crypto/crypto.hpp
tewinget: TunnelNonce isn't used any more anyway, but removing it cascaded into a lot of compiler whack-a-mole on code which is pending changes anyway,  so I was leaving this for until then (at which time we can just have one `Nonce` type, which should have been the case all along it seems).  I can include that change here, but I didn't follow the whack-a-mole rabbit hole to its conclusion so I'm not 100% sure where it terminates (might be really shallow and I ditched it early...)

PR #2217, opened 2023-10-25
SQLite (WIP)

jagerman:
This PR aims to replace the logic around RC handling (particularly w.r.t. timeouts), RC generation and acceptance, and RC storage (changing to use an sqlite database rather than files on disk).

PR #2218, opened 2023-10-30
RC Refactor + (some) BT-encode fixes

dr7ana:
Picking up on RC/BT work from #2217


dr7ana:
Accidentally closed when the commit was dropped, re-opening tomorrow with a re-working of routercontacts

llarp/router_contact_remote.cpp
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  void
+  RemoteRC::bt_encode(oxenc::bt_dict_producer& btdp) const
+  {
+    (void)btdp;
+
+    // TODO: implement append_encoded in oxenc so we can finish this implementation.
+    // It is almost identical to the implementation of LocalRC::bt_encode, except the
+    // call to append_signature is append_encoded.
```
jagerman: As described this won't work, because it will potentially break signatures (if the RC we received had extra keys we didn't understand, from a future Lokinet; or ones we ignore from a past lokinet).  We'll have to instead get the entire, untouched, encoded RC as we received it and push that out.

jagerman: For example, future lokinet might decide it doesn't care about ipv6 anymore and stops parsing the ipv6 field entirely, so then this:
```json
{"4": "abcdpp", "6": "whateverpp", "p": "...pubkey...", "~": "xyz..."}
```
would get round-trip re-encoded as:
```json
{"4": "abcdpp", "p": "...pubkey...", "~": "xyz..."}
```
but now the `xyz...` won't match anymore because the "6" will be missing from the serialized dict output.

llarp/crypto/types.cpp
```diff
@@ -32,14 +33,44 @@ namespace llarp
...
+  operator==(const PubKey& lhs, const RouterID& rhs)
+  {
+    return lhs.as_array() == rhs.as_array();
+  }
+
+  bool
+  operator==(const RouterID& lhs, const PubKey& rhs)
+  {
+    return lhs.as_array() == rhs.as_array();
+  }
```
jagerman: I'm pretty sure that we can get rid of both of these from here and the `.hpp` because the two-PubKey `operator==` should already properly match any pubkey+routerid pair (due to the added inheritance).

llarp/crypto/types.cpp
```diff
@@ -32,14 +33,44 @@ namespace llarp
...
+
   bool
   SecretKey::LoadFromFile(const fs::path& fname)
```
jagerman, line 61: I feel like we should get rid of these `LoadFromFile` methods and just have a single function something like:
```C++
template <typename T, typename = std::enable_if_t<is_aligned_buffer<T>>>
T load_from_file(const fs::path& fname) {
    ...
}

// ...
auto sk = load_from_file<SecretKey>("filename");
```
Having these nearly-identical `LoadFromFile` scatter all over the codebase (making `fs.hpp` included almost everywhere) seems sub-optimal.

llarp/router_contact_remote.cpp
dr7ana: I know it wont work! I remember your explanation on that. To clarify, I meant we need `append_encoded` for adding the signature to the end when we call `RemoteRC::bt_encode`, not the individual key/value pairs. For `LocalRC::bt_encode`, we will need `append_signature` (which we already have) at the very end. I can explain more over mumble when you're online

llarp/crypto/types.cpp
dr7ana: I agree, but question and a note:

Q: What other `LoadFromFile` methods are you referring to? I only see the one in `SecretKey` used only in two places
```
dan@demeter:~/oxen/lokinet/llarp$ grep -R "LoadFromFile" *
config/key_manager.cpp:    return key.LoadFromFile(path);
crypto/types.hpp:    LoadFromFile(const fs::path& fname);
crypto/types.cpp:  SecretKey::LoadFromFile(const fs::path& fname)
exit/session.cpp:    return exit_key.LoadFromFile(fname);
```

Note: We already have other free methods for reading files directly to buffers or string [here](https://github.com/oxen-io/lokinet/blob/07271f9ae7a9d0f39c3b866d06d625e218315419/llarp/util/file.hpp#L13). Their former name is absolutely terrible so I re-named them, but maybe we can modify these to be more what you're thinking of?

llarp/crypto/types.hpp
```diff
@@ -161,58 +143,24 @@ namespace llarp
...
   constexpr inline bool IsToStringFormattable<SecretKey> = true;
   template <>
   constexpr inline bool IsToStringFormattable<PrivateKey> = true;
-  template <>
-  constexpr inline bool IsToStringFormattable<IdentitySecret> = true;
 
   using ShortHash = AlignedBuffer<SHORTHASHSIZE>;
   using LongHash = AlignedBuffer<HASHSIZE>;
 
   struct Signature final : public AlignedBuffer<SIGSIZE>
```
tewinget, line 156: `using`?

llarp/crypto/types.cpp
```diff
@@ -32,14 +33,38 @@ namespace llarp
...
+  operator==(const PubKey& lhs, const PubKey& rhs)
+  {
+    return lhs.as_array() == rhs.as_array();
+  }
+
+  bool
+  operator==(const PubKey& lhs, const RouterID& rhs)
+  {
+    return lhs.as_array() == rhs.as_array();
+  }
```
jagerman, line 58: This one can go away too.

llarp/crypto/types.cpp
```diff
@@ -32,14 +33,38 @@ namespace llarp
     return oxenc::to_hex(begin(), end());
   }
 
+  PubKey::operator RouterID() const
+  {
+    return {as_array()};
+  }
```
jagerman, line 39: I wonder if this is still needed?

PR #2221, opened 2023-11-03
Config Refactor

dr7ana:
finally, some good food

### DIFF
--- a/llarp/config/config.cpp
+++ b/llarp/config/config.cpp
@@ -41,21 +41,21 @@ namespace llarp
   }  // namespace
 
   void
-  RouterConfig::defineConfigOptions(ConfigDefinition& conf, const ConfigGenParameters& params)
+  RouterConfig::define_config_options(ConfigDefinition& conf, const ConfigGenParameters& params)
   {
     constexpr Default DefaultJobQueueSize{1024 * 8};
     constexpr Default DefaultWorkerThreads{0};
     constexpr Default DefaultBlockBogons{true};
 
-    conf.defineOption<int>(
+    conf.define_option<int>(
         "router", "job-queue-size", DefaultJobQueueSize, Hidden, [this](int arg) {
           if (arg < 1024)
             throw std::invalid_argument("job-queue-size must be 1024 or greater");
 
-          m_JobQueueSize = arg;
+          job_que_size = arg;
         });
 
-    conf.defineOption<std::string>(
+    conf.define_option<std::string>(
         "router",
         "netid",
         Default{llarp::LOKINET_DEFAULT_NETID},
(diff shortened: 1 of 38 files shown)


llarp/config/config.cpp
```diff
@@ -932,50 +931,97 @@ namespace llarp
...
+      {
+        auto h = arg_v.substr(0, pos);
+        uint16_t p = std::atoi(arg_v.substr(pos + 1).data());
 
-      if (addr)
+        maybe = oxen::quic::Address{std::string{h}, p};
+
+        // TODO: unfuck llarp/net
+        // if (net_ptr->IsLoopbackAddress(addr->port()))
+        //   throw std::invalid_argument{fmt::format("{} is a loopback address", arg)};
```
jagerman: We should still do this, but via: https://github.com/oxen-io/oxen-libquic/pull/65

dr7ana: fixed with your previous commits

llarp/config/config.cpp
```diff
@@ -936,48 +933,95 @@ namespace llarp
...
+        arg_v = std::string_view{arg};
       }
-      if (not addr)
+
+      if (arg_v[0] == ':')
       {
-        // infer public address
-        if (auto maybe_ifname = net_ptr->GetBestNetIF())
-          addr = net_ptr->GetInterfaceAddr(*maybe_ifname);
+        uint16_t res = std::atoi(arg_v.substr(1).data());
```
jagerman: We have a nicer/safer/more modern `llarp::parse_int<uint16_t>` for integer parsing.  (There's another place for it below)

llarp/config/config.cpp
```diff
@@ -1035,57 +1087,60 @@ namespace llarp
...
+          if (auto a = parse_addr_for_link(arg); a and a->is_addressable())
+            addr = a;
           else
-            throw std::invalid_argument{fmt::format("invalid port value: '{}'", val)};
-
-          inbound_listen_addrs.emplace_back(std::move(*addr));
+            addr = oxen::quic::Address{""s, DEFAULT_LISTEN_PORT};
         });
+
+    // DISCUSS: drop this shit?
```
jagerman: I think we have to keep this, unfortunately.  I think there is still quite a bit of use of it in the wild (it wasn't all *that* long ago that we deprecated it).  The `INTERFACE=` option is probably less important to worry about keeping, but `1.2.3.4=5678` and `*=1234` are probably still noticeably used.

llarp/config/config.cpp
```diff
@@ -936,48 +933,95 @@ namespace llarp
...
+          using_new_api = true;
+        });
+
     conf.define_option<std::string>(
         "bind",
         "inbound",
         RelayOnly,
         MultiValue,
         Comment{
+            "********** DEPRECATED **********",
```
jagerman, line 1024: Instead of a comment like this, we should just add the ~~`Deprecated`~~ `Hidden` flag to the option definition, so that in generated configs (`lokinet -c -g dir/lokinet.ini` or `lokinet -r -g dir/lokinet.ini` which is what generates a new default config) the section won't show up at all.

jagerman: Whoops, I'm wrong: we want `Hidden`, not `Deprecated` to hide it but still process it.  (`Deprecated` is for a no-longer-does-anything-at-all option that we want to just ignore but not error on if we see it).

llarp/config/config.cpp
dr7ana: ugh

llarp/config/config.cpp
dr7ana: I didn't add a deprecated flag to the function parameters? I just put deprecated as the first comment line. I added "Hidden" after "RelayOnly" and "MultiValue," is that what you meant?

jagerman (approved):

PR #2223, opened 2023-11-14
onion encrypt path build frames

tewinget:
Depends #2216

path build frames should be onioned at each hop to avoid a bad actor controlling two nodes in a path being able to know with certainty, temporal correlation is hard to avoid) that they're hops on the same path.  This is desirable as in the worst case someone could be your edge hop and terminal hop on a path, and now the terminal hop knows your IP, making the path basically pointless.

### DIFF
--- a/llarp/link/link_manager.cpp
+++ b/llarp/link/link_manager.cpp
@@ -1120,12 +1120,7 @@ namespace llarp
 
     try
     {
-      std::string frame_payload;
-      std::string frame, hash, hop_payload, commkey, rx_id, tx_id, upstream;
-      ustring other_pubkey, outer_nonce, inner_nonce;
-      uint64_t lifetime;
-
-      auto payload_list = oxenc::bt_deserialize<std::deque<std::string>>(m.body());
+      auto payload_list = oxenc::bt_deserialize<std::deque<ustring>>(m.body());
       if (payload_list.size() != path::MAX_LEN)
       {
         log::info(link_cat, "Path build message with wrong number of frames");
@@ -1134,18 +1129,18 @@ namespace llarp
       }
 
       oxenc::bt_dict_consumer frame_info{payload_list.front()};
-      hash = frame_info.require<std::string>("HASH");
-      frame = frame_info.require<std::string>("FRAME");
+      auto hash = frame_info.require<ustring>("HASH");
+      auto frame = frame_info.require<ustring>("FRAME");
 
       oxenc::bt_dict_consumer hop_dict{frame};
-      hop_payload = frame_info.require<std::string>("ENCRYPTED");
-      outer_nonce = frame_info.require<ustring>("NONCE");
-      other_pubkey = frame_info.require<ustring>("PUBKEY");
+      auto hop_payload = hop_dict.require<ustring>("ENCRYPTED");
+      auto outer_nonce = hop_dict.require<ustring>("NONCE");
+      auto other_pubkey = hop_dict.require<ustring>("PUBKEY");
 
       SharedSecret shared;
       // derive shared secret using ephemeral pubkey and our secret key (and nonce)
       if (!crypto::dh_server(
-              shared.data(), other_pubkey.data(), _router.pubkey(), inner_nonce.data()))
+              shared.data(), other_pubkey.data(), _router.pubkey(), outer_nonce.data()))
       {
         log::info(link_cat, "DH server initialization failed during path build");
         m.respond(serialize_response({{messages::STATUS_KEY, PathBuildMessage::BAD_CRYPTO}}), true);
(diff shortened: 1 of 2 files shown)


llarp/link/link_manager.cpp
```diff
@@ -1118,114 +1118,97 @@ namespace llarp
...
-        commkey = hop_info.require<std::string>("COMMKEY");
-        lifetime = hop_info.require<uint64_t>("LIFETIME");
-        inner_nonce = hop_info.require<ustring>("NONCE");
-        rx_id = hop_info.require<std::string>("RX");
-        tx_id = hop_info.require<std::string>("TX");
-        upstream = hop_info.require<std::string>("UPSTREAM");
+      // hash data and check against given hash
+      ShortHash digest;
+      if (!crypto::hmac(
+              digest.data(), reinterpret_cast<unsigned char*>(frame.data()), frame.size(), shared))
```
dr7ana: Suggestion:
Putting this comment here once but it applies below as well. If you wanted to avoid reinterpret casting to unsigned char so many times, you could just do:
```C++
ustring whatever;

oxenc::bt_dict_consumer btdc{m.data()};

whatever = btdc.require<ustring>("WHATEVER");
```
I know some of these need to be `char` for specific uses, but maybe a few of these cases could be streamlined?

llarp/link/link_manager.cpp
```diff
@@ -1118,114 +1118,97 @@ namespace llarp
...
+      commkey = hop_info.require<std::string>("COMMKEY");
+      lifetime = hop_info.require<uint64_t>("LIFETIME");
+      inner_nonce = hop_info.require<ustring>("NONCE");
+      rx_id = hop_info.require<std::string>("RX");
+      tx_id = hop_info.require<std::string>("TX");
+      upstream = hop_info.require<std::string>("UPSTREAM");
+
       // populate transit hop object with hop info
       // TODO: IP / path build limiting clients
       auto hop = std::make_shared<path::TransitHop>();
```
dr7ana, line 1184: Clarifying question:
Why do we make a shared pointer here? I had to pull up the whole file (stupid github UI), and it seems like we don't put it anywhere at the end of this function?

llarp/path/pathbuilder.cpp
```diff
@@ -429,40 +429,72 @@ namespace llarp
           path_cat, "{} building path -> {} : {}", Name(), path->ShortName(), path->HopsString());
 
       oxenc::bt_list_producer frames;
-
+      std::vector<std::string> frame_str(path::MAX_LEN);
```
dr7ana, line 432: Medium-ish?:

So this constructor jumped out at me, mostly because I've had to reference vector constructors to double check the 1 and 2 parameter versions and what they do. So [here](https://en.cppreference.com/w/cpp/container/vector/vector) we see that
```C++
std::vector<int> d(3); // creates a 3-element vector holding {0, 0, 0}
```
Which got me thinking, is this vector of strings making a vector of empty strings `path::MAX_LEN` long? Or is it making a vector that could hold strings reserved to size `path::MAX_LEN`? I had a feeling it was the first one, whereas using [`vector::reserve`](https://en.cppreference.com/w/cpp/container/vector/reserve) seems like it wouldn't create new objects in the memory you're reserving.

So I put that shit into godbolt:
```C++
// Method 1:
std::vector<std::string> temp(5);

// Method 2:
std::vector<std::string> temp;
temp.reserve(5);
```
The second method produced non-trivially less instructions than the first; also `.size()` returns 5 for method 1 and 0 for method 2, but `.capacity()` returns 5 for both.

I know this is a weirdly deep dive into one line, but I was curious because this is something I've looked into before in regards to vector constructors

dr7ana (commented):
Looks solid, you cranked this out Tom. Will probably ask you later to walk me through it just to ensure I'm tracking everything

llarp/link/link_manager.cpp
tewinget: Not only are you correct, but also the above is incorrect in a different way entirely as well.  Well spotted, will fix.

llarp/link/link_manager.cpp
tewinget: These numbers subject to change, but lines 1248 and 1286 will insert the hop into the "our transit hops" container (if terminus) or (upon success response from next hop), respectively.

llarp/path/pathbuilder.cpp
tewinget: > also .size() returns 5 for method 1 and 0 for method 2, but .capacity() returns 5 for both.

That was the intention, as I populate the strings in the vector in reverse order.  Can change to declaring the vector and calling resize on separate lines if it's more readable, but reserve wouldn't be sufficient in this case.

llarp/link/link_manager.cpp
tewinget: fixed the other issues I spotted and almost forgot to change the thing you mentioned before pushing...that would've been hilarious.  Either way, fixed and pushed.

PR #2224, opened 2023-11-15
RC gossip

tewinget:
Depends #2216

Adds a "gossip_rc" command.  As part of the ongoing RC refactor, the notion of find/lookup a single RC and "network exploration" are going away.  These are removed in this PR, but can be separated if desired.


llarp/link/link_manager.cpp
```diff
@@ -398,6 +383,27 @@ namespace llarp
     });
   }
 
+  void
+  LinkManager::gossip_rc(const RemoteRC& rc)
+  {
+    for (auto& [rid, conn] : ep.conns)
+      send_control_message(rid, "gossip_rc", std::string{rc.view()}, nullptr);
```
dr7ana: Minor:
Doesn't the third parameter default to `nullptr`?

llarp/nodedb.cpp
```diff
@@ -99,6 +99,73 @@ namespace llarp
...
+    auto itr = router_whitelist.begin();
+    if (sz > 1)
+      std::advance(itr, randint() % sz);
+    return *itr;
+  }
+
+  bool
+  NodeDB::is_connection_allowed(const RouterID& remote) const
+  {
+    if (pinned_edges.size() && pinned_edges.count(remote) == 0 && !bootstraps.count(remote))
```
dr7ana: Hilariously minor:
So minor that I'm laughing writing it. You alternate between `container.count(target) == 0` and `!container.count(target)` and periodically a `not containter.count(target)` (maybe that third one is a residual formalism from my terrible practice, since I usually put `not` instead of `!`)

llarp/router/router.cpp
```diff
@@ -925,15 +913,17 @@ namespace llarp
       // the whitelist enabled and we got the whitelist
       // check against the whitelist and remove if it's not
       // in the whitelist OR if there is no whitelist don't remove
-      if (has_whitelist and not _rc_lookup_handler.is_session_allowed(rc.router_id()))
+      if (not node_db()->is_connection_allowed(rc.router_id()))
```
dr7ana, line 878: Bringing the `rc_lookup_handler` functions into `NodeDB` is pretty clean

llarp/router/router.cpp
```diff
@@ -1033,13 +1018,20 @@ namespace llarp
     return _link_manager.get_random_connected(result);
   }
 
+  std::unordered_set<RouterID>
+  Router::router_whitelist() const
+  {
+    return _node_db->whitelist();
```
dr7ana: Two things:
1) Maybe less minor: `NodeDB::whitelist()` returns a const reference to the unordered set, so should the invoking `Router::router_whitelist()` method do the same?
2) Super minor: I feel like calling it `Router::router_whitelist()` is redundant, maybe we just call the getters after the object? Like `Router::get_whitelist()` and so on. I know `get_{whatever}` is a little on the nose, but a few lines down you call `NodeDB::set_router_whitelist()`, which is very descriptive and good

llarp/service/outbound_context.cpp
```diff
@@ -407,6 +405,18 @@ namespace llarp::service
 
     std::vector<Introduction> intros = current_intro.intros;
 
+    // don't consider intros for which we don't have the RC for the pivot
+    auto itr = intros.begin();
+    while (itr != intros.end())
+    {
+      if (not ep.router()->node_db()->has_router(itr->router))
```
dr7ana: Minor:
Maybe we can bring the call to `NodeDB::has_router` into a method in `Router` that describes what you're checking, like `Router::has_peer_router`? (or something less dumb lol)

llarp/link/link_manager.cpp
```diff
@@ -384,10 +389,19 @@ namespace llarp
...
-      send_control_message(rid, "gossip_rc", std::string{rc.view()}, nullptr);
+    {
+      // don't send back to the owner...
+      if (rid == rc_rid)
+        continue;
+      // don't gossip RCs to clients
+      if (not conn->remote_is_relay)
+        continue;
+
+      send_control_message(rid, "gossip_rc", serialized_rc, nullptr);
```
dr7ana: Minor:
Same thing about the default callback parameter

dr7ana (commented):
Did a first pass and left a few thoughts, looks solid thanks for cranking this out so quickly Tom

llarp/link/link_manager.cpp
tewinget: It does, yeah.  I'm contemplating whether we want it to default, actually.  The reason I explicitly pass `nullptr` here is to signal to future-me (or others reading it) that I intentionally don't have a response handler, rather than accidentally forgot.

llarp/nodedb.cpp
tewinget: Sometimes inconsistencies like that stick out to me like a sore thumb and other times they're just background noise...

I agree this should be consistent, but I always waffle back and forth on how to express "`count()` is or is not zero" (bool or integer compare).  Will address in future commit in PR (or yell at me if I forget).

llarp/router/router.cpp
tewinget: They had to live somewhere, and rc_lookup_handler is thankfully obsolete.  It was spaghetti from the moment I wrote it. :'(

llarp/router/router.cpp
tewinget: 1) yes.  I moved this from the header (need to fix a dependency loop between NodeDB and Router so NodeDB is forward-declared in the header...that's a future fix though), and I didn't notice it was making a copy...for who knows what reason

2) they should really all be like "relay_whitelist" anyway, I think.  Will keep in mind to do a pass over variable/class/method names involving the word "router" and fixing them when relevant to PRs.

llarp/service/outbound_context.cpp
tewinget: The method name here is dumb, agreed.  As for having a method that simply passes through the request to a sub-object: it's something I've done quite a bit in the past and find it's almost never worth the abstraction if it's safe/sane for the caller to just use the sub-object directly.  Either way, either this should change to something like you suggest here, or `router_whitelist` (rename pending) above (and probably others) should go away and callers should do `router->node_db()->whitelist()` if they need it.  I prefer the latter; thoughts?

llarp/link/link_manager.cpp
tewinget: Yup, just being explicit to convey intent.

llarp/service/outbound_context.cpp
dr7ana: I don't feel super strongly about it, except that if we're doing `ep.router().whatever()->whatever_method()->whatever2->yadda_yadda()` more than once in a file maybe we should method it?

tewinget: I can see that on a case-by-case basis.  Bit subjective, and I'm by no means picky about it.

PR #2225, opened 2023-11-25
RC and RouterID fetching

tewinget:
Implements the mechanisms by which a client will fetch the current list of RouterIDs and RCs from the network.  Relays will use the latter on bootstrap as well, confirming registration against the RouterIDs they get from oxend.

TODO: RouterIDs are fetched from 12 different relays (if possible); the logic to wait for all the responses is in place, but the logic to reconcile them is not.

TODO: RCs are fetched from a connected edge.  The logic to distrust this edge if its RC list deviates too far from our known RouterID list is not yet implemented.

Depends #2224

### DIFF
--- a/llarp/link/link_manager.cpp
+++ b/llarp/link/link_manager.cpp
@@ -7,6 +7,7 @@
 #include <llarp/messages/exit.hpp>
 #include <llarp/messages/path.hpp>
 #include <llarp/messages/rc.hpp>
+#include <llarp/messages/router_id.hpp>
 #include <llarp/nodedb.hpp>
 #include <llarp/path/path.hpp>
 #include <llarp/router/router.hpp>
@@ -14,6 +15,7 @@
 #include <oxenc/bt_producer.h>
 
 #include <algorithm>
+#include <exception>
 #include <set>
 
 namespace llarp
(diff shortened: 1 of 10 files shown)


llarp/messages/dht.hpp
```diff
@@ -6,7 +6,6 @@ namespace llarp
 {
   namespace FindRouterMessage
   {
```
jagerman: Should go away, I think?

llarp/link/link_manager.cpp
```diff
@@ -382,6 +393,271 @@ namespace llarp
...
+          {
+            log::info(link_cat, "RC Fetch to {} returned error.", source);
+            return;
+          }
+
+          try
+          {
+            oxenc::bt_dict_consumer btdc{m.body()};
+            btdc.required("rcs");
+            auto btlc = btdc.consume_list_consumer();
```
jagerman:
```suggestion
            auto btlc = btdc.require<oxenc::bt_list_consumer>("rcs");
```

llarp/link/link_manager.cpp
```diff
@@ -382,6 +393,271 @@ namespace llarp
...
+              //       and catch it below so we know what about parsing failed?
+              rcs.emplace_back(btlc.consume_dict_consumer());
+            }
+
+            node_db->ingest_rcs(source, std::move(rcs), timestamp);
+          }
+          catch (const std::exception& e)
+          {
+            // TODO: Inform NodeDB of failure (perhaps just a call to rotate_rc_source())
+            log::info(link_cat, "Failed to parse RC Fetch response from {}", source);
```
jagerman:
```suggestion
            log::info(link_cat, "Failed to parse RC Fetch response from {}: {}", source, e.what());
```

llarp/link/link_manager.cpp
```diff
@@ -382,6 +393,271 @@ namespace llarp
...
+            return;
+          }
+        });
+  }
+
+  void
+  LinkManager::handle_fetch_rcs(oxen::quic::message m)
+  {
+    // this handler should not be registered for clients
+    assert(not _router.is_service_node());
```
jagerman: Perhaps stronger than an assert?

tewinget: could do, but not sure what would be appropriate.  I inverted the condition though, so good you gave me pause.

jagerman: This shouldn't be a registered endpoint for a client, so assert is fine.

llarp/link/link_manager.cpp
```diff
@@ -427,6 +430,120 @@ namespace llarp
...
+          {
+            // TODO: keep track of this failure for relay quality metrics?
+            log::info(link_cat, "RC Fetch to {} timed out", source);
+            return;
+          }
+          if (not m)
+          {
+            log::info(link_cat, "RC Fetch to {} returned error.", source);
+            return;
+          }
```
dr7ana: You could go:
```C++
if (not m)
{
    log::info(link_cat, "RC fetch to {} {}", source, m.timed_out ? "timed out" : "returned error");
    return;
}
```

llarp/link/link_manager.cpp
```diff
@@ -427,6 +430,120 @@ namespace llarp
...
+
+    const auto& rcs = node_db->get_rcs();
+    const auto now =
+        std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now());
+    try
+    {
+      oxenc::bt_dict_consumer btdc{m.body()};
+
+      btdc.required("explicit_ids");
+      auto explicit_ids = btdc.consume_list<std::vector<std::string>>();
```
dr7ana, line 496:
```C++
auto explicit_ids = btdc.require<std::vector<std::string>>("explicit_ids");
```

And all other instances of `::required(...)` followed by `::consume`

llarp/link/link_manager.cpp
```diff
@@ -427,6 +430,120 @@ namespace llarp
...
+              {{messages::STATUS_KEY, RCFetchMessage::INVALID_REQUEST}}));
+          return;
+        }
+        explicit_relays.emplace(reinterpret_cast<const byte_t*>(sv.data()));
+      }
+
+      oxenc::bt_dict_producer resp;
+
+      {
+        auto btlp = resp.append_list("rcs");
```
dr7ana: Suggestion RE Naming for clarity's sake:

Call the main dict producer "btdp" and then call the subproducers "subdict" and "sublist"

llarp/messages/rc.hpp
```diff
@@ -0,0 +1,28 @@
...
+  serialize(std::chrono::system_clock::time_point since, const std::vector<RouterID>& explicit_ids)
+  {
+    oxenc::bt_dict_producer btdp;
+
+    try
+    {
+      btdp.append("since", since.time_since_epoch() / 1s);
+      auto id_list = btdp.append_list("explicit_ids");
+      for (const auto& rid : explicit_ids)
+        id_list.append(rid.ToView());
```
dr7ana: For uniformity's sake, maybe wrap the sublists in their own braces like you did above

llarp/link/link_manager.cpp
```diff
@@ -544,6 +546,120 @@ namespace llarp
...
+                source,
+                edge);
+            node_db->ingest_router_ids(edge, {});  // empty response == failure
+            return;
+          }
+          try
+          {
+            oxenc::bt_dict_consumer btdc{m.body()};
+            btdc.required("routers");
+            auto router_id_strings = btdc.consume_list<std::vector<ustring>>();
```
dr7ana, line 577: These can be combined too

llarp/nodedb.cpp
```diff
@@ -114,7 +114,27 @@ namespace llarp
...
-  {}
+  {
+    auto conn_count = router.link_manager().get_num_connected();
+    if (conn_count == 0)
+    {
+      // not connected to any nodes yet, so no sensible source
+      return;
+    }
+    RemoteRC new_source{};
+    router.link_manager().get_random_connected(new_source);
```
dr7ana, line 134: Question about our off band talk today: In our "From the Top" logic today, won't we need to rotate the RC node selecting from the list of stale RC's, rather than the connected nodes?

llarp/nodedb.hpp
```diff
@@ -124,6 +126,12 @@ namespace llarp
     void
     update_rcs();
 
+    void
+    fetch_router_ids();
+
+    void
+    select_router_id_sources(std::unordered_set<RouterID> excluded);
```
dr7ana: Suggestion: Default this parameter to an empty set, so you don't have to specify it above when you invoke this function

llarp/link/link_manager.cpp
tewinget: good call, but it should probably be a bit different still, i.e. extract and print the response status string in the non-timeout case.

llarp/link/link_manager.cpp
tewinget: ty, I was probably doing it wrong as I tried (something resembling) that and the compiler complained.  This works.

tewinget: noting for posterity: `oxenc::bt_dict_consumer::require` doesn't currently support consuming lists/dicts; we can change this if/when we update oxenc to support that.

llarp/link/link_manager.cpp
tewinget: went with `rc_bt_list`, can do `sublist` if preferred.

dr7ana: nah all good just being a nit pick

llarp/link/link_manager.cpp
```diff
@@ -427,6 +432,232 @@ namespace llarp
...
+                return;
+              }
+              router_ids.emplace_back(s.data());
+            }
+            node_db->ingest_router_ids(edge, std::move(router_ids));
+            return;
+          }
+          catch (const std::exception& e)
+          {
+            log::info(link_cat, "Error handling fetch RouterIDs response: {}", e.what());
```
dr7ana, line 600: What do you want to do in this error case? Roll on directly into `node_db->ingest_router_ids`? Or re-throw the exception

tewinget: The throw could come from handling bad data (and almost always will barring memory issues, but we can have multiple catch blocks and different exceptions if we need to at some point).  In that case we don't really want the exception to bubble up, but we do still need to signal to node_db somehow that the request to that relay failed.  Using `ingest_router_ids` with an empty result to signal that now, but we can have an explicit failure method if that seems more correct.

PR #2226, opened 2023-11-28
Bootstrap from scratch rework

tewinget:
depends #2225

### DIFF
--- a/llarp/link/link_manager.cpp
+++ b/llarp/link/link_manager.cpp
@@ -160,16 +160,12 @@ namespace llarp
           [this, msg = std::move(m)]() mutable { handle_gossip_rc(std::move(msg)); });
     });
 
-    for (auto& method : direct_requests)
+    for (auto& [method, callback] : direct_requests)
     {
       s->register_command(
-          std::string{method.first}, [this, func = method.second](oxen::quic::message m) {
+          std::string{method}, [this, func = callback](oxen::quic::message m) {
             _router.loop()->call([this, msg = std::move(m), func = std::move(func)]() mutable {
-              auto body = msg.body_str();
-              auto respond = [m = std::move(msg)](std::string response) mutable {
-                m.respond(std::move(response), not m);
-              };
-              std::invoke(func, this, body, std::move(respond));
+              std::invoke(func, this, std::move(msg));
             });
           });
     }
--- a/llarp/link/link_manager.cpp
+++ b/llarp/link/link_manager.cpp
@@ -434,48 +434,9 @@ namespace llarp
 
   void
   LinkManager::fetch_rcs(
-      const RouterID& source, rc_time since, const std::vector<RouterID>& explicit_ids)
+      const RouterID& source, std::string payload, std::function<void(oxen::quic::message m)> func)
   {
-    send_control_message(
-        source,
-        "fetch_rcs",
-        RCFetchMessage::serialize(since, explicit_ids),
-        [this, source = source](oxen::quic::message m) {
-          if (m.timed_out)
-          {
-            // TODO: keep track of this failure for relay quality metrics?
-            log::info(link_cat, "RC Fetch to {} timed out", source);
-            return;
-          }
-          try
-          {
-            oxenc::bt_dict_consumer btdc{m.body()};
-            if (not m)
-            {
-              auto reason = btdc.require<std::string_view>(messages::STATUS_KEY);
-              log::info(link_cat, "RC Fetch to {} returned error: {}", source, reason);
-              return;
-            }
-
-            auto btlc = btdc.require<oxenc::bt_list_consumer>("rcs"sv);
-            auto timestamp = rc_time{std::chrono::seconds{btdc.require<int64_t>("time"sv)}};
-
-            std::vector<RemoteRC> rcs;
-
-            while (not btlc.is_finished())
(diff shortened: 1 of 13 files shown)


llarp/nodedb.cpp
```diff
@@ -109,7 +109,176 @@ namespace llarp
...
+      // TODO: reconcile all the responses, for now just insert all
+      for (const auto& [rid, responses] : router_id_fetch_responses)
+      {
+        // TODO: empty == failure, handle that case
+        for (const auto& response : responses)
+        {
+          client_known_routers.insert(std::move(response));
+        }
+      }
+      router_id_fetch_in_progress = false;
```
dr7ana: Since this boolean is set to true inside `::fetch_router_ids()` and then set to false inside `::ingest_router_ids()` inside a response lambda for `send_control_message`, should we make it atomic?

llarp/link/link_manager.cpp
```diff
@@ -427,6 +432,232 @@ namespace llarp
...
+
+      btdc.required("explicit_ids");
+      auto explicit_ids = btdc.consume_list<std::vector<std::string>>();
+      auto since_time = rc_time{std::chrono::seconds{btdc.require<int64_t>("since")}};
+
+      if (explicit_ids.size() > (rcs.size() / 4))
+      {
+        log::info(
+            link_cat, "Remote requested too many relay IDs (greater than 1/4 of what we have).");
+        m.respond(serialize_response({{messages::STATUS_KEY, RCFetchMessage::INVALID_REQUEST}}));
```
dr7ana: I fixed these in my branch extending yours, but all these errors need to have `true` as the second parameter to `m.respond`. I made
 
```C++
inline const auto INVALID_REQUEST = messages::serialize_response({{messages::STATUS_KEY, "Invalid relay ID requested"}});
```

to use for the invocations

llarp/link/link_manager.cpp
```diff
@@ -427,6 +432,232 @@ namespace llarp
...
+      }
+
+      resp.append("time", now.time_since_epoch().count());
+
+      m.respond(std::move(resp).str(), false);
+    }
+    catch (const std::exception& e)
+    {
+      log::info(link_cat, "Exception handling RC Fetch request: {}", e.what());
+      m.respond(messages::ERROR_RESPONSE);
```
dr7ana: same here RE: second parameter

llarp/link/link_manager.cpp
```diff
@@ -427,6 +432,232 @@ namespace llarp
...
+        for (const auto& [_, rc] : rcs)
+        {
+          if (last_time.at(rc.router_id()) > since_time or explicit_relays.count(rc.router_id()))
+            rc_bt_list.append_encoded(rc.view());
+        }
+      }
+
+      resp.append("time", now.time_since_epoch().count());
+
+      m.respond(std::move(resp).str(), false);
```
dr7ana: This defaults to false for second parameter

llarp/link/link_manager.hpp
```diff
@@ -345,13 +345,12 @@ namespace llarp
...
-    // TODO: new RC fetch endpoint (which will be both client->relay and relay->relay)
     std::unordered_map<
         std::string_view,
-        void (LinkManager::*)(std::string_view body, std::function<void(std::string)> respond)>
+        void (LinkManager::*)(oxen::quic::message m)>
         direct_requests = {
-            {"publish_intro"sv, &LinkManager::handle_publish_intro},
-            {"find_intro"sv, &LinkManager::handle_find_intro}};
+            {"fetch_rcs"sv, &LinkManager::handle_fetch_rcs},
+            {"fetch_router_ids"sv, &LinkManager::handle_fetch_router_ids}};
```
dr7ana, line 353: Even as I'm typing this i realize how minor of an optimization it could be, but it leads to an idea of something you and i have chatted about before:

Why don't we encode all the endpoint names? These are definitely more descriptive written out, but it would save some space going from:
```
"publish_intro" --> "PI"
"find_intro" --> "FI"
"fetch_rcs" --> "FRC"
"fetch_router_ids" --> "FRID"
```
It's the secondary reason why we use low-value ASCII characters for bt keys (primary being that they always appear first); by extension what do you think about changing all the bt keys to single letter values?

The part we've talked about:
Sometime next week you and I could sit down and revise the old docs that had the incorrect single-letter mappings to the message types into something sensical and cohesive...?

llarp/nodedb.cpp
tewinget: All lokinet logic is functionally single-threaded, but if/when we spin jobs off to other threads this may become a consideration.

llarp/link/link_manager.hpp
tewinget: I'm all for smaller messages, eventually.  For the time being until we've done a good amount of testing, I would prefer the more descriptive keys if only because we'll be poring over log files and dumping bt-encoded containers to figure out where we went wrong and having it be human readable takes precedence for that purpose, imo.  Definitely not gonna die on that hill though.

dr7ana: agreed. but lets make sure to sit down and knock out that conversation sometime, i agree that after testing is a great time for it

jagerman:
Obsoleted by changes in #2227.

PR #2227, opened 2023-11-28
Trust model

dr7ana:


llarp/config/config.cpp
```diff
@@ -1250,6 +1250,15 @@ namespace llarp
         [this](std::string arg) { rpc_addr = oxenmq::address(arg); });
 
     // Deprecated options:
+    conf.define_option<std::string>("lokid", "jsonrpc", RelayOnly, Deprecated, [](std::string arg) {
```
jagerman: This should not be marked `Deprecated`: that tag means the option is ignored (with a warning about being ignored) if used, but this option needs to be a fatal error if specified and not ignored.  It should be `Hidden`, however, so that it doesn't go into generate configs.

llarp/config/config.cpp
```diff
@@ -61,8 +61,8 @@ namespace llarp
         "netid",
         Default{llarp::LOKINET_DEFAULT_NETID},
         Comment{
-            "Network ID; this is '"s + llarp::LOKINET_DEFAULT_NETID
-                + "' for mainnet, 'gamma' for testnet.",
+            "Network ID; this is '"s + llarp::LOKINET_DEFAULT_NETID + "' for mainnet, "s
+                + llarp::LOKINET_TESTNET_NETID + "for testnet."s,
```
jagerman: Needs some reformatting; this is currently becoming:
```ini
# Network ID; this is 'lokinet' for mainnet, testnetfor testnet.
```
in a generated config (`./daemon/lokinet -r -g foo/lokinet.ini`)

PR #2241, opened 2024-10-03
Detect and recover from a bad service node state

tewinget:
If a node is deregistered while running (or is started before being registered), it can fail to properly connect to the network once registered (again).  This does not *fix* that, but these symptoms should be sufficient to detect such a state, and the node should recover upon restart.


jagerman:
This has been carried as a patch in the debs for a couple months now, and I recently had a registration that reproduced the bug causing lokinet to get stuck with no peers, which successfully triggered this workaround and a restart that unstuck it without intervention.

PR #2252, opened 2025-01-28
Fix invalid oxend rpc call

jagerman:
This call is wrong (and has always been wrong): the field should be set in the top-level parameters, not inside `fields`.

In current dev lokinet, however, the invalid field becomes an error, because dev tries to parse all `fields` keys with `bool` values, and this ends up returning an error to Lokinet instead of a response, which breaks Lokinet's service node updating code when used with oxend 11.


jagerman:
Also noting in passing: this has been fixed for a while in the WIP code, probably because it encountered the same issue with testnet nodes at some point.

PR #2266, opened 2025-09-09
Added missing header in router.cpp

mpretty-cyro:


jagerman:
Cherry-picked in 37fc67c14

PR #2272, opened 2025-09-25
Session and path message refactor

tewinget:
This adds authentication to path and session messages, and encryption to session control messages and path control responses.  All traffic should now be correctly encrypted and authenticated.


llarp/crypto/crypto.hpp
```diff
@@ -20,6 +22,19 @@ namespace llarp::crypto
     /// xchacha symmetric cipher
     void xchacha20(std::span<std::byte> buf, const SharedSecret&, const SymmNonce&);
 
+    static constexpr size_t MAC_SIZE = crypto_aead_xchacha20poly1305_ietf_ABYTES;
+
+    /// encrypts a buffer in-place, putting a MAC in the final bytes
+    void xchacha20_poly1305_encrypt(std::span<std::byte> buf, const SharedSecret& secret, const SymmNonce& nonce);
+    void xchacha20_poly1305_encrypt(std::vector<std::byte>& buf, const SharedSecret& secret, const SymmNonce& nonce);
```
jagerman: This second overload is unnecessary: a `std::vector<std::byte>` is already implicitly convertible to a mutable span.

llarp/crypto/crypto.hpp
```diff
@@ -20,6 +22,19 @@ namespace llarp::crypto
     /// xchacha symmetric cipher
     void xchacha20(std::span<std::byte> buf, const SharedSecret&, const SymmNonce&);
 
+    static constexpr size_t MAC_SIZE = crypto_aead_xchacha20poly1305_ietf_ABYTES;
+
+    /// encrypts a buffer in-place, putting a MAC in the final bytes
```
jagerman:
```suggestion
    /// encrypts a buffer in-place, putting a MAC in the final bytes (which must be allocated
    /// in the span but are not part of the data that is actually encrypted).
```

Issue #2288, opened 2026-08-08
Client-mode AAAA lookups return synthetic 'NS localhost.loki.' instead of real answer/SERVFAIL (v0.9.14)


cpradmin:
### Summary

On lokinet 0.9.14, client-mode AAAA queries against arbitrary `.loki` service addresses — both a target service and a separately-tested, independently-documented "known good" control address — consistently return a synthetic authoritative `NS localhost.loki.` record instead of either a real resolved address or a clean failure (SERVFAIL/NXDOMAIN). The router's own logs show genuine network-level lookup attempts against multiple peer relays failing ("failed to lookup ... from ...snode"), while the client's own service simultaneously connects to the network successfully (introset published, paths built). The bug reproduces identically across two independent, unrelated environments.

### Environment 1 — bare metal, static binary

- OS: Nobara Linux 44 (Fedora-based), kernel 7.1.3
- lokinet: v0.9.14, official static release binary (`lokinet-linux-amd64-v0.9.14.tar.xz`)
- Install: manual, systemd service, client-mode config (`lokinet -g -f`, no `-r`)
- DNS wiring tested two ways: systemd-resolved split-DNS (`resolvectl dns/domain` → `127.3.2.1`), and a dedicated dnsmasq forwarder (`server=/loki/127.3.2.1`) — both produced the identical result below
- Also tested via a separate SOCKS5 setup (lokinet + Dante in a container, [soren-work/lokinet-proxy](https://github.com/soren-work/lokinet-proxy)) — same underlying failure, different symptom shape (see "SOCKS5" below)

### Environment 2 — clean LXC container, official package

- OS: fresh Ubuntu 24.04 LXC container (Proxmox), no prior configuration
- lokinet: v0.9.14, installed via the **official apt repo** (`deb.oxen.io`), not the static binary
- DNS: lokinet's own bundled `lokinet-resolvconf` integration (ran automatically on install)
- Router status at test time: `known/connected: 25/3`, `paths/endpoints 8/0`, multiple paths built in under 1.5s

### Reproduction

```
dig @127.3.2.1 AAAA <any-.loki-service-address>
```

Result (both environments, both target and control addresses):

```
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: ...
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; ANSWER SECTION:
[...]

PR #2289, opened 2026-08-08
Fix client-mode AAAA lookups returning synthetic NS localhost.loki. placeholder (#2288)

cpradmin:
Fixes the placeholder-record symptom in #2288.

## Root cause

`m_UseV6` was hardcoded `false` in `llarp/handlers/tun.cpp`, which forced every
client-mode AAAA lookup for a `.loki` address into the `isV6 && !SupportsV6()`
branch of `HandleHookedDNSMessage`, unconditionally returning the synthetic
`NS localhost.loki.` record described in #2288 instead of a real answer or a
clean failure. `.loki` service addresses are IPv6-only, so client mode needs
`m_UseV6 = true` for AAAA resolution to work at all.

## Verification

Built this patch and tested live against mainnet (client-mode config, no
`-r`):

- A real, currently-published introset now resolves to a real `AAAA` answer
  instead of the placeholder.
- A nonexistent `.loki` address now returns `NXDOMAIN` instead of the
  placeholder.
- A control `A` query is unaffected (still clean `NXDOMAIN`, unchanged from
  before this patch).

## Scope — what this does NOT fix

Per the maintainer feedback already on #2288: this patch only addresses which
lookups get the bogus `NS` placeholder record. It does **not** address the
separate point raised there that Lokinet's `NXDOMAIN` usage is itself
semantically wrong for a "record doesn't exist" case (should be `NOERROR` +
0 answers / no-data, per how `session-router` handles it) — that's a
different, deeper issue in the DNS response path this patch leaves
untouched. Flagging that clearly so this isn't read as a full fix of #2288,
just the placeholder-record part of it.


jagerman:
This change is not correct: this would enable Lokinet's broken IPv6 support, which breaks things considerably more than having an unwanted NS DNS response.

